PR #35, opened 2021-02-09
Initramfs generation

### LINKED ISSUE #1: Initramfs generation
Need to handle initramfs generation - both for LiveCDs and for system installs. We currently are using dev packages to achieve that to ease out and discover what we really need, but it should be enhanced to be extensible.

This card is also about generating an initramfs that could be used directly for #29 

Alternatives:
- https://github.com/u-root/u-root
- dracut

mudler:
Those changeset introduce [u-root](https://github.com/u-root/u-root) to generate the initramfs, and provides an init written in Golang. It also leverages `go-udev` to automatically detect the required kernel modules, by also probing the `modules.alias` file.
The Go init also setups the network connection, allowing to load cloud-config init over the net (not tested yet, might miss the ca-certificates.crt bundle file in initramfs).

It also drops kernel-minimal, tumbleweed-minimal and the init package, as now it is all generated by one package.

Fixes #1


mudler:
CI so green

YAY

PR #41, opened 2021-02-15
Configure /etc/ssl and default dns server

### LINKED ISSUE #37: Permanent custom cloud-config during install
Currently the installation process copies over the squashfs content, we should provide a way to persist also a custom cloud-config during install.

- [x] Modify the installer to accept a custom cloud-config
- [x] Copy the cloud-config over the persistent partition
- [x] QA

Needs #5

mudler:
Those were required in order to make curl work.
Note: `update-ca-certificates` didn't the trick - somehow when being run at early stages seems to fail. Didn't investigated(yet)

Related to #37

### DIFF
--- a/packages/cloud-config/definition.yaml
+++ b/packages/cloud-config/definition.yaml
@@ -1,3 +1,3 @@
 name: cloud-config
 category: system
-version: "0.4.8"
+version: "0.4.9"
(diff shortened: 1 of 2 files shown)


davidcassany (approved):
Looks good :+1:

Issue #8, opened 2021-01-08
A/B seamless Dual copy updates


mudler:
Upgrades shouldn't run on the real system, either be transactional with rollbacks or just apply the upgrade on a transient rootfs (a/b) that finally gets promoted once final checks succeeds.

The rootfs partition will be "image based", that means:
 - single version for whole OS, that gets upgraded from 1 package only, or either directly from a Docker Image with a custom component (doesn't have to be luet)
- update entire OS and reboot
- rollback on old partition if boot into updated system is not successful
- ships a configuration for the upgrade channel, that can be override with a cloud-init approach

## Action items

- [x] Create release-channel file, which is a luet config pointing to the DockerHub repository we use to release final-images ( https://github.com/mudler/cOS/blob/master/Makefile#L17 ). The configuration will be either constructed in the OEM cloud-init config, or in the `system/cos` package
- [x] Such file has to be editable also via cloud-init, if not supplied by it
- [x] Create an upgrade script, which sets up the new partition and flag it properly where the system will boot next - and install the docker image with luet or a different tool. Once the installation is over, boot into the new partition
- [x] If the system fails to boot the new partition, fallback to the old one
- [x] Make sure the above steps are idempotent, and can be executed if things either goes good or bad. Ensure that the partition 'flip' works as expected
- [x] Make sure the above steps works for other release channels too

~~Needed by #21~~
Depends on ~~#29~~, Related to #5

mudler:
This is fixed now, there will be follow-ups like #64  #23 for fallbacks and recovery scenarios

PR #43, opened 2021-02-15
Fixups to cos-upgrade

mudler:
Now the state partitions defaults to 2GB.
It now also detect booting partition in cos-upgrade, doing the partition switch based on the one we are booting into ( so it is idempotent)

Also, setup TMPDIR and XDG_RUNTIME_DIR until we have a persistent state partition for the update process.
TMPDIR for luet can be annotated in the config file, but temporary do it here until the process is completely fixed

Also wipe passive partition, until we get a persistent db for luet.

Add gawk to the base image, which is used in cos-upgrade

I avoided to touch the cloud-init file, to avoid any conflict, it's something that can be tweaked afterwards. Until #37 is fixed, can't really split into multiple files.

Related to #8

### DIFF
--- a/packages/installer/upgrade.sh
+++ b/packages/installer/upgrade.sh
@@ -27,6 +27,35 @@ find_partitions() {
         echo "OEM partition cannot be found"
         exit 1
     fi
+
+    CURRENT=$(df $0 | tail -1 | gawk '{print $1}')
+    if [ -z "$CURRENT" ]; then
+        echo "Could not determine current partition"
+        exit 1
+    fi
+    if [ -z "$ACTIVE" ]; then
+        echo "Could not determine active partition"
+        exit 1
+    fi
+    if [ -z "$PASSIVE" ]; then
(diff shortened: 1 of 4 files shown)


mudler:
cc @davidcassany can you review ?

davidcassany (approved):
Looks good to me

packages/installer/upgrade.sh
```diff
@@ -27,6 +27,35 @@ find_partitions() {
         echo "OEM partition cannot be found"
         exit 1
     fi
+
+    CURRENT=$(df $0 | tail -1 | gawk '{print $1}')
```
davidcassany, line 31: I would probably base it on the current mount points rather than having it based on `df` output command order, but those are details we can later polish.

mudler: yup, for now I tried to make it simple, so I'm just checking where the current script is coming from

PR #46, opened 2021-02-17
Split cloud-init files

### LINKED ISSUE #42: Multiple cloud-init files breaks boot
See https://github.com/mudler/cOS/issues/37#issuecomment-779294008

While testing custom oem config files, seems that having multiple yamls loaded at the initramfs stage breaks boot up.

Splitting into a separate card as it seems completely unrelated to the oem mechanism itself

mudler:
- Drop setup of /etc/fstab, and mkdir command which is not required
anymore.
- Also enables getty explictly on tty1.
- Reorders the big default cloud init file into multiple ones
- Sets /etc in initramfs, drop from cloud-init
- Bumps cos with the new kernel

cc @davidcassany hope doesn't conflict with your work, it seems split files are working as expected now and that should ease out touching this package

Fixes #42


davidcassany (approved):
Looks good to me. This is an easy way to stack stages and the files defining those stages

PR #47, opened 2021-02-19
Minor fixups to cos-upgrade (2)

mudler:
This PR has mostly fixups to the installer and the upgrade script.

Mainly, for the upgrade, the interesting part is that now we update grub. As we don't have a partition holding the system state, we don't retain the grub generated files anywhere - and during upgrade it means we wipe them off too. As we don't have decided yet what  to do with those, simply recall grub update after pulling the new image into the partition for now

This also fixes some glitches noticed on first boot after the system upgrade - now seems to go rather smoothly. 

The only thing to fixup with grub now is the fallback mechanism which doesn't seems to work correctly: at least the good side of updating grub in this stage is propagating that sorta of changes too.

Relates  to  #8


davidcassany (approved):
Looks good :+1:

Issue #5, opened 2021-01-08
File system layout


mudler:
By default, only the relevant dir for the container runtime should be writable.

We should have then:
- [x] read only `/usr` - `/` if feasable
- [x] ephemeral `/etc` that gets wiped out each time (tmpfs mount?)
- [x] persistency only on `/usr/local` and `/oem` folder

Mind also to switch back https://github.com/mudler/cOS/blob/31749456b54f3eb60ad203cd3f212c6717b29320/packages/initramfs/loader.sh#L225

Persistency partitions and `/etc/fstab` configuration can be driven by cloud-init config files, currently in https://github.com/mudler/cOS/blob/31749456b54f3eb60ad203cd3f212c6717b29320/packages/cloud-config/yipfiles/01_defaults.yaml#L12 and from the installation phase
Every user-side or oem configuration, will be handled by #4 and #24

## Action items:

- [x] Investigate layout from other projects on how they feased problematics regarding this topic
- [x] Configure the layout in the init, and cloud-config files
- [x] Modify installer to create a persistent partition (that occupy all the space left)
- [x] Tweak init to mount such partition
- [x] QA

~~Depends on #4~~ Merged

davidcassany:
I did a search on MicroOS and the layout there looks like:

* `/etc` is an overlayfs on tmpfs
* `/` is a readonly mount
*  `root`, `/opt`, `/home`, `/usr/local` and `/srv` read/write btrfs subvolume
* `/var` read/write partition
*  Then some `boot` artifacts are read write too.

mudler:
I'll close this as we merged #48 , if needed will file new issues for follow-ups

Issue #33, opened 2021-01-28
Kernel should create unique symlink


mudler:
https://github.com/mudler/cOS/blob/85e764da2ab6698d9c3d216de0e55834a8ee25d7/packages/kernels/build.yaml#L9

Instead of symlinking with the explicit minimal kernel name, there should be a unique one. So the `installer` doesn't need to guess the name when creating the GRUB configuration, like in: https://github.com/mudler/cOS/blob/85e764da2ab6698d9c3d216de0e55834a8ee25d7/packages/installer/installer.sh#L137

Action items:
- Drop explicit symlink, and pick a unique one based from the package name
- Make other kernels, if presents, to `conflict` between each other (optional, not strictly required at the moment)
- Change the installer to use that one
- Verify installer still works

mudler:
Fixed in #48

PR #48, opened 2021-02-19
Set immutable rootfs and use dracut based standard initrd

### LINKED ISSUE #40: Dracut based initramfs
Now that the initramfs is split into a separate file, it's more easy to provide the logic of the `initramfs` package with Dracut too.

davidcassany:
This commit is two fold.

From one side it sets the use of dracut based
initrds and sets the cos-setup config to run after switching the root.
On the other hand this commit also sets the ephemeral /etc, /var and
/root mountpoints on tmpfs.

All cos-setup configurations are applied on
tmpfs after switching the root, right after local-fs.target (so all
the local filesystems are already there)

Fixes #40 related #5


davidcassany:
This PR is not rebased on latest master, so this is still work in progress.

packages/base-conf/definition.yaml
```diff
@@ -0,0 +1,3 @@
+name: "base-conf"
+category: "system"
+version: "0.0.1+24"
```
davidcassany, line 3: For convenience while developing I created this config package, it basically holds the layout setup.

packages/cloud-config/yipfiles/01_defaults.yaml
```diff
@@ -1,25 +1,8 @@
...
-        - path: /etc/fstab
-          content: |
-                    tmpfs   /etc         tmpfs   rw,nodev,nosuid,size=20MB          0  0
-#                    LABEL=COS_OEM        /oem       ext4  defaults                    0  1
-#                    LABEL=COS_PERSISTENT /usr/local ext4  data=ordered                0  2
-#                    LABEL=COS_GRUB       /boot/efi  vfat  defaults                    0  0
-          permissions: 0644
-          owner: 0
-          group: 0
+    fs.after:
```
davidcassany, line 3: I changed to apply all on the fly settings over the ephemeral /etc. Then, since /etc is the default one from Tumbleweed untouched I just run few tweaks and the whole yaml can be reduced.

packages/cloud-config/yipfiles/01_defaults.yaml
```diff
@@ -50,124 +33,13 @@ stages:
...
-       - /usr/bin/systemctl enable cos-setup-reconcile.timer
-       - /usr/bin/systemctl enable cos-setup-fs-after.service
-       - /usr/bin/systemctl enable cos-setup-boot.service
-       - /usr/bin/systemctl enable cos-setup-network.service
-       - /usr/bin/systemctl mask purge-kernels
-       - mkdir /etc/ssl
-       - ln -s /var/lib/ca-certificates/pem /etc/ssl/certs
-       - ln -s /var/lib/ca-certificates/ca-bundle.pem /etc/ssl/ca-bundle.pem
+# commands always tries to expand vars (aka $..) so we can't pass a hashed passwd here
+#       - /bin/bash -c "/usr/sbin/usermod -p '$1$Oncxs0L7$U9/3Kuc2nP4oJowCNEquJ0' -g root root"
```
davidcassany, line 38: I really struggled hours because of that... I'd say this might be a bug of the `yip` tool, I could not manage a way to run `/usr/sbin/usermod -p '$1$Oncxs0L7$U9/3Kuc2nP4oJowCNEquJ0' -g root root` and prevent variable expansion. Is there anything I missed about yip usage?

packages/cos/build.yaml
```diff
@@ -11,43 +11,31 @@ requires:
...
 # In opensuse, /bin/sh is a symlink to /etc/alternatives/sh, which points to bash.
 # We don't supply etc/alternatives, so handle the symlink by ourselves
 - rm -rf /bin/sh && ln -s /usr/bin/bash /bin/sh
+- systemctl enable multi-user.target
+- systemctl set-default multi-user.target
+- systemctl enable cos-setup-reconcile.timer
+- systemctl enable cos-setup-fs-after.service
+- systemctl enable cos-setup-boot.service
+- systemctl enable cos-setup-network.service
+- systemctl mask purge-kernels.service
```
davidcassany, line 34: Since `/etc/` is not wiped at this stage we can include here some fundamentals.

packages/initrd/cos-config-launcher.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
davidcassany, line 1: This hook is damn ugly, probably would be nicer to just install cos-setup inside the initrd and create an early service that loads COS_OEM volume if any and run initramfs.after from the initramfs itself. But this might be a more sophisticated approach to do `ignition` or `combustion` like tweaks.

packages/livecd/live-boot/package/boot/uefi/loader/entries/cos-x86_64.conf
```diff
@@ -1,4 +1,4 @@
 title cOS Linux Live
 version x86_64
 efi /minimal/x86_64/kernel.xz
-options initrd=/minimal/x86_64/rootfs.xz cdroot rootdelay=7
+options initrd=/minimal/x86_64/rootfs.xz root=live:/dev/cdrom rd.live.dir=/ rd.live.squashimg=rootfs.squashfs
```
davidcassany, line 4: This `/dev/cdrom` parameter can be avoided if we set a LABEL to the ISO, I think it is the `-volid` parameter of `xorriso`.

davidcassany (commented):
Few comments about the implementation

packages/cloud-config/yipfiles/01_defaults.yaml
mudler: Can be caused by `yip` templating, but either on how we run commands in `yip` - in both cases we rely on external libraries, so we might have to look at that closer

packages/initrd/cos-config-launcher.sh
mudler: I'd say it's fine anyway, it opens up for further tweaks that we might need really soon

iso/cOS-local.yaml
```diff
@@ -1,6 +1,8 @@
 packages:
   rootfs:
   - system/cos
+  - system/initrd
```
mudler, line 4: Just a reminder, maybe that was committed accidentally: we need to make `system/cos` depend on those two instead before merging, so we have one image only for the upgrade.

mudler:
Looks good to me! :+1: Just the small nitpick about the `system/cos` image dependency

iso/cOS-local.yaml
davidcassany: Yes I know, it was just to speed up the testing. This is clearly still work in progress, but I wanted to start showing it just in case there was stuff to discuss. And, at least, I wanted to make a PR that eventually can be build locally and produces something close to the intended result of this PR.

davidcassany:
@mudler all right I did a quick rebase and some of the grub stuff I left it untouched, meaning this undoes some some the latest master changes. Just need to figure out how to adapt them rather then roll it back, in any case, at least for now, the current state does not conflict and in my local env the resulting iso was capable to boot, install and reboot with the expected results.

Still work in progress to ensure it does not break and undo latest upgrade changes.

For reference I did the test with:
```
$ qemu-img create -f raw ~/imagesToTest/test.img 16G
$ qemu-kvm -m 2048 -hda ~/imagesToTest/test.img -bios /usr/share/qemu/ovmf-x86_64.bin -cdrom cOS-0.20210219.iso -boot d --serial stdio
```
Note that this PR expects the `/dev/cdrom` to be available and this might not match running in other hipervisors other than the above test.

davidcassany:
@mudler I added a golang version bump just to verify the nature of the failure https://github.com/mudler/cOS/pull/48/commits/50c0d4ac90bc5699353a89fcfa6f0c729dd25f8c. I am afraid we are seeing the same issue I am seeing in my Leap host since early last week... something as simple as:

`docker run --rm -ti opensuse/tumbleweed zypper install -y git` fails unless it is executed as `--privileged`. I could not figure out why neither understand the nature of the problem, but apparently some regular bash file checks and not properly working causing issues here and there.

I debugged dracut generation and it fails because of this

```
david@localhost:~> docker run --rm -ti opensuse/tumbleweed
22ec195c62dc:/ # touch /myNewFile
22ec195c62dc:/ # ls /
.dockerenv  bin  boot  dev  etc  home  lib  lib64  mnt  myNewFile  opt  proc  root  run  sbin  srv  sys  tmp  usr  var
22ec195c62dc:/ # if [[ -w / ]]; then echo "Write permissions for /"; else echo "No permissions for /"; fi
No permissions for /
```

So I can effectively write on `/` (or anywhere else), but if I run a `[[ -w <path> ]]` test on bash it tells me I have no permission. This is solved running the container as privileged. I am wondering if this could be related to latest [docker releases](https://docs.docker.com/engine/release-notes/19.03/#190315), however what puzzles me is that I am seeing the same behavior in podman too. I started having this issue a week ago.

mudler:
I tried to build this branch locally and I get: 

```
(15/48) Installing: libmagic1-5.39-1.4.x86_64 [........done]                                                                                                                                                                          [5/969]
(16/48) Installing: libgdbm6-1.19-1.2.x86_64 [......done]  
(17/48) Installing: libexpat1-2.2.10-1.2.x86_64 [.........done]                                                       
(18/48) Installing: libedit0-3.1.snap20180525-4.3.x86_64 [..........done]                                             
(19/48) Installing: libeconf0-0.3.8+git20200710.5126fff-2.2.x86_64 [....done]                                         
(20/48) Installing: libdb-4_8-4.8.30-38.24.x86_64 [............done]                                                  
(21/48) Installing: libcrypt1-4.4.17-1.2.x86_64 [...........done]                                                     
(22/48) Installing: libcbor0_8-0.8.0-1.2.x86_64 [......done]                                                          
(23/48) Installing: libcares2-1.17.1-1.2.x86_64 [......done]                                                          
(24/48) Installing: libaudit1-2.8.5-5.2.x86_64 [.......done]                                                          
(25/48) Installing: gzip-1.10-6.2.x86_64 [.........done]   
(26/48) Installing: chkstat-1550_20210125-27.2.x86_64 [...........done]                                               
(27/48) Installing: libnsl2-1.3.0-1.4.x86_64 [.......done]                                                            
(28/48) Installing: rsync-3.2.3-2.2.x86_64 [..........done]                                                           
(29/48) Installing: file-5.39-1.4.x86_64 [.....done]       
(30/48) Installing: libgdbm_compat4-1.19-1.2.x86_64 [.....done]                                                       
(31/48) Installing: libmetalink3-0.1.3-3.9.x86_64 [......done]                                                        
(32/48) Installing: perl-base-5.32.0-2.2.x86_64 [............done]                                                    
(33/48) Installing: libfido2-1-1.6.0-1.2.x86_64 [........done]                                                        
(34/48) Installing: libsemanage1-3.1-1.2.x86_64 [...........done]                                                     
(35/48) Installing: permissions-config-1550_20210125-27.2.x86_64 [.......done]                                        
Additional rpm output:                                     
ERROR: fillup not found. This should not happen. Please compare                                                       
/etc/sysconfig/security and /sysconfig.security and        
update by hand.                                            
error opening /etc/sysconfig/security: No such file or directory                                                      


(36/48) Installing: less-563-1.3.x86_64 [.........done]    
(37/48) Installing: wget-1.21.1-1.2.x86_64 [...........done]                                                          
(38/48) Installing: perl-5.32.0-2.2.x86_64 [............done]                                                         
(39/48) Installing: openssh-clients-8.4p1-4.1.x86_64 [............done]                                               
(40/48) Installing: permissions-20210125.1550-27.2.x86_64 [...done]                                                   
(41/48) Installing: perl-Error-0.17029-1.4.noarch [......done]                                                        
(42/48) Installing: pam-1.5.1-2.2.x86_64 [............done]                                                           
(43/48) Installing: shadow-4.8.1-5.2.x86_64 [............done]                                                        
(44/48) Installing: sysuser-shadow-3.0-10.1.noarch [....done]                                                         
(45/48) Installing: openssh-server-8.4p1-4.1.x86_64 [..error]                                                         
Installation of openssh-server-8.4p1-4.1.x86_64 failed:    
Error: Subprocess failed. Error: RPM failed: find: '/etc/init.d/rc[35].d': No such file or directory                                                                                                                                         
ERROR: neither useradd nor busybox found!                  
error: %prein(openssh-server-8.4p1-4.1.x86_64) scriptlet failed, exit status 1                                        
error: openssh-server-8.4p1-4.1.x86_64: install failed     

Abort, retry, ignore? [a/r/i] (a): a                       
Warning: %posttrans scripts skipped while aborting:        
    pam-1.5.1-2.2.x86_64.rpm                               
    shadow-4.8.1-5.2.x86_64.rpm                            
    openssh-server-8.4p1-4.1.x86_64.rpm                    

Problem occurred during or after installation or removal of packages:                                                 
Installation has been aborted as directed.                 
Please see the above error message for a hint.             
The command '/bin/sh -c zypper in -y wget tar gzip git && zypper install -y -t pattern devel_basis' returned a non-zero code: 8
```

I think we somehow get logs lost in CI, but seems something new to me :eyes:

mudler:
> > I think we somehow get logs lost in CI, but seems something new to me eyes
> 
> This is the same I asked you about last Wednesday on rocket chat, I started seeing it last Tuesday. At first I thought it was something wrong with my new package as the failure was happening on my dracut package. Then at some point a wanted to do a full rebuild of everything an cleaned all caches and binaries and then started to fail on the golang package, at that point I thought it could have been some transitory issue for tumbleweed and started to using the published repository for the ISO creation. But then other failures started to appear... until I did not start running the `opensuse/tumbleweed` image shell manually I did not realize all issues were connected to some wrong permissions detection. I still have no clue where the root issue is coming from. I see this on my Leap host and I expected to be something specific to my host as on my Tumbleweed laptop I do not have this issue, I only see it on my Leap workstation.

About the log loss, I think it's due to https://github.com/mudler/luet/pull/183 . But the error looks somehow genuine - do we have `adduser`? I think worth trying updating the base image, as things may have changed there

> 
> Since this is also happening in the github actions I start to think this might not be something specific to my host, do we know which docker version is running for the github actions? Under which distro is all this running?

You should be able to see from the github action logs:

```
2021-02-22T06:18:04.2778808Z Environment: ubuntu-18.04
2021-02-22T06:18:04.2779277Z Version: 20210211.1
2021-02-22T06:18:04.2780193Z Included Software: https://github.com/actions/virtual-environments/blob/ubuntu18/20210211.1/images/linux/Ubuntu1804-README.md
2021-02-22T06:18:04.2781026Z ##[endgroup]
....
2021-02-22T06:18:12.9457216Z ##[group]Run docker-practice/actions-setup-docker@0.0.1
2021-02-22T06:18:12.9457945Z with:
2021-02-22T06:18:12.9458415Z   docker_version: 19.03
2021-02-22T06:18:12.9458797Z   docker_channel: stable
2021-02-22T06:18:12.9459152Z ##[endgroup]
2021-02-22T06:18:12.9841312Z [command]docker version
2021-02-22T06:18:13.7510476Z Client:
2021-02-22T06:18:13.7514770Z  Version:           19.03.15+azure
2021-02-22T06:18:13.7515876Z  API version:       1.41
2021-02-22T06:18:13.7516530Z  Go version:        go1.13.15
2021-02-22T06:18:13.7517278Z  Git commit:        48d30b5b32e99c932b4ea3edca74353feddd83ff
2021-02-22T06:18:13.7517802Z  Built:             Thu Jan 28 22:27:59 2021
2021-02-22T06:18:13.7518168Z  OS/Arch:           linux/amd64
2021-02-22T06:18:13.7518699Z  Context:           default
2021-02-22T06:18:13.7519036Z  Experimental:      true
2021-02-22T06:18:13.7519271Z 
2021-02-22T06:18:13.7519526Z Server:
2021-02-22T06:18:13.7519970Z  Engine:
2021-02-22T06:18:13.7520334Z   Version:          20.10.2+azure
2021-02-22T06:18:13.7520729Z   API version:      1.41 (minimum version 1.12)
2021-02-22T06:18:13.7521116Z   Go version:       go1.13.15
2021-02-22T06:18:13.7521596Z   Git commit:       8891c58a433a823ce0a65b57efff45f32ee9cb45
2021-02-22T06:18:13.7522097Z   Built:            Fri Dec 25 15:07:49 2020
2021-02-22T06:18:13.7522420Z   OS/Arch:          linux/amd64
2021-02-22T06:18:13.7522767Z   Experimental:     false
2021-02-22T06:18:13.7523171Z  containerd:
2021-02-22T06:18:13.7523486Z   Version:          1.3.9+azure
2021-02-22T06:18:13.7523999Z   GitCommit:        ea765aba0d05254012b0b9e595e995c09186427f
2021-02-22T06:18:13.7524796Z  runc:
2021-02-22T06:18:13.7525643Z   Version:          1.0.0-rc92
2021-02-22T06:18:13.7526144Z   GitCommit:        ff819c7e9184c13b7c2607fe6c30ae19403a7aff
2021-02-22T06:18:13.7526774Z  docker-init:
2021-02-22T06:18:13.7527083Z   Version:          0.19.0
2021-02-22T06:18:13.7527381Z   GitCommit:

**Comment by davidcassany** [comment]
>  But the error looks somehow genuine - do we have `useradd`? I think worth trying updating the base image, as things may have changed there
> 
> >

Yes, we have `useradd`, in fact it is just installed right before `openssh-server`, is provided by `shadow` package.

> (43/48) Installing: shadow-4.8.1-5.2.x86_64 [............done] 

I bet the problem is that the `openssh-server` scriptlet tries to detect whether there is `busybox` or `adduser` by running some shell file check that returns unexpected results. Adding busybox to this transaction just causes other sort of failures in some other parts of the installation, I went mad trying to sort this out on Wednesday.

**Comment by mudler** [comment]
> > But the error looks somehow genuine - do we have `useradd`? I think worth trying updating the base image, as things may have changed there
> > >
> 
> Yes, we have `useradd`, in fact it is just installed right before `openssh-server`, is provided by `shadow` package.
> 
> > (43/48) Installing: shadow-4.8.1-5.2.x86_64 [............done]
> 
> I bet the problem is that the `openssh-server` scriptlet tries to detect whether there is `busybox` or `adduser` by running some shell file check that returns unexpected results. Adding busybox to this transaction just causes other sort of failures in some other parts of the installation, I went mad trying to sort this out on Wednesday.

Looks we are not the only ones: https://forums.opensuse.org/showthread.php/550195-Install-git-in-docker 

maybe  we can work around it with `--no-recommends` for the time being

**Comment by davidcassany** [comment]
> maybe we can work around it with `--no-recommends` for the time being to not block this PR

This is already part of this PR... So it will not fix it. Anyway, the conclusion of this forum is not really satisfying, as this is not only causing issues on `openssh-server` installation, this is causing issues in bash scripts executions, so it leads to hard to identify issues as it can potentially break package installations and who knows what else. However the  `--security-opt seccomp:unconfined` to make it work reveals some further details about the root cause, so probably we can workaround it by applying some configuration in the host.

**Comment by davidcassany** [comment]
@mudler I am about to revert the last commit, as I just added it to verify this was the issue. So I guess it will fail again on the initrd package again as there won't be any changes for the golang package.

**Comment by mudler** [comment]
> > maybe we can work around it with `--no-recommends` for the time being to not block this PR
> 
> This is already part of this PR... So it will not fix it. Anyway, the conclusion of this forum is not really satisfying, as this is not only causing issues on `openssh-server` installation, this is causing issues in bash scripts executions, so it leads to hard to identify issues as it can potentially break package installations and who knows what else. However the `--security-opt seccomp:unconfined` to make it work reveals some further details about the root cause, so probably we can workaround it by applying some configuration in the host.

yes indeed, I'm having a look at https://github.com/mudler/luet/issues/146 so we can pass-by backends to the arg

File: packages/livecd/live-boot/package/boot/uefi/loader/entries/cos-x86_64.conf (hunk 6 above)
Comment (mudler) [reply packages/livecd/live-boot/package/boot/uefi/loader/entries/cos-x86_64.conf]: I've added the support for it over weekend, should be enough to set `label: "foo"` in the `iso/cOS-local.yaml` file

Comment (davidcassany) [reply packages/livecd/live-boot/package/boot/uefi/loader/entries/cos-x86_64.conf]: Awesome, gonna test it :)

**Comment by mudler** [comment]
I've been experimenting in the `debug` branch, and after switching to leap everything seems back to normal: https://github.com/mudler/cOS/commit/9dce3ae906f4c542586e8917a764205eac9d755c

Seems to me some issue is in the TW packages, but since it's out of purpose for this PR and there is no reason to block it, I'd suggest to rather go for leap for now and document the issue with tumbleweed in a separate one.

As we will eventually move to a SLE based, there is no real reason for us to stick to tw (and we need more stability right now too).

File: iso/cOS-local.yaml [hunk 8]
```diff
@@ -9,12 +11,12 @@ packages:
...
-  kernel_file: "vmlinuz-vanilla"
-  rootfs_file: "initramfs"
+  kernel_file: "vmlinuz"
+  rootfs_file: "initrd"
 
 overlay: "true"
 image_prefix: "cOS-0."
 image_date: "true"
-
+label: "COS_LIVE"
```
mudler, line 20: This is now setting the new ISO label

iso/cOS-local.yaml
```diff
@@ -9,12 +11,12 @@ packages:
   - live/boot
 
 initramfs:
-  kernel_file: "vmlinuz-vanilla"
-  rootfs_file: "initramfs"
+  kernel_file: "vmlinuz"
```
mudler, line 14: We should be good with #33  too

packages/livecd/live-boot/package/boot/uefi/loader/entries/cos-x86_64.conf
```diff
@@ -1,4 +1,4 @@
 title cOS Linux Live
 version x86_64
 efi /minimal/x86_64/kernel.xz
-options initrd=/minimal/x86_64/rootfs.xz cdroot rootdelay=7
+options initrd=/minimal/x86_64/rootfs.xz root=live:/dev/cdrom rd.live.dir=/ rd.live.squashimg=rootfs.squashfs console=tty1 console=ttyS0
```
mudler, line 4: small reminder: same is needed for syslinux

mudler: I've tried to boot with `root=live:LABEL=COS_LIVE` without much luck

mudler:
I guess we can also drop `u-root` and `initramfs`, which are no longer used

mudler:
VirtualBox_cOS_24_02_2021_09_06_11

Just tried but doesn't seem to mount properly the `/sysroot` here (vbox), having a :eyes:

davidcassany:
> Update: non-EFI boots fine, but seems cloud-config didn't triggered and can't login with default credentials

@mudler Oh just seen the message now, I hope I did not step over you with my commit. I can EFI boot, install and reboot on the installed systemd with this last setup. Regarding the cloud-init I fixed few details, there is no need to call `*.after` and `*.before` stages explicitly as `cos-setup` already does it and with the current setup only `fs` stage and forward make sense.

mudler:
with https://github.com/mudler/cOS/pull/48/commits/0bb561b100f1e6ae90d9e3cc8f17ba96f3c7e386 I can finally boot on EFI too :+1: . To note, with an installed system, now ISO seems to mount persistence and oem also in live

davidcassany:
> with [0bb561b](https://github.com/mudler/cOS/commit/0bb561b100f1e6ae90d9e3cc8f17ba96f3c7e386) I can finally boot on EFI too +1 . To note, with an installed system, now ISO seems to mount persistence and oem also in live

This happens if the devices are already there. This could be considered a feature or not, depends on the expectations. Basically if udev detects COS_OEM and COS_PERSISTENT labels they get mounted, this could be interesting to provide custom configuration from first boot even on live boots, however I understand that within the current installer implementation this is just likely to produce wrong results. 

I am also thinking about a more sophisticated setup within initramfs so we can better instruct custom setup before switch root. I noticed that with the current approach adding custom udev rules is not so easy. But this can be done in a second round.

packages/cloud-config/yipfiles/01_defaults.yaml
mudler: this should be solved now, the new `yip` version has a `users` field that can be used to set passwords for users:

```yaml
stages:
 ...
     - name: "Setup users"
       users: 
          root: "$1$Oncxs0L7$U9/3Kuc2nP4oJowCNEquJ0"
```

mudler (approved):

Issue #32, opened 2021-01-28
Simplify makefile


mudler:
Besides some general refactoring which is needed, we should have one command that :

- generates all packages (e.g. already present `make rebuild-all`)
- the ISO (e.g. already present `make local-iso`)
- Finally a target to run qemu with it (to write `make run-qemu`)

Consider also adapting current GHA pipelines to test the new process too (without `run-qemu`, if that's hard to test)

kkaempf:
This should also merge `.envrc` into `Makefile`

mudler:
All the items on the issue were covered, closing it

PR #51, opened 2021-02-25
Semplify Makefile

mudler:
Drop rebuild-full and rebuild-all. Keeps only `make build `.

It also drops `SUDO`, we now warn about it and let the user decide. The exception is `make deps` which requires root to install the binaries locally.

By defaults takes the packages to build from the given ISO_SPEC file, defaulting
to cOS-local.yaml which we use currently

Related to #32


davidcassany (approved):
Good, I like these changes

Issue #27, opened 2021-01-20
Generate ISOs with no host dependencies


mudler:
Currently the build iso process is unfriendly - it requires deps from the host system, and moreover the dependency on luet-extensions make it rather complex to follow.

Keep in mind that the building process might require "external" artifacts as the EFI image, and those could be provided separately, or as part of the system that we are trying to boot from the ISO.

We should explore if it's feasable to replace the process with a single golang binary  which uses https://github.com/diskfs/go-diskfs/

https://github.com/diskfs/go-diskfs/blob/master/examples/bootable_iso.go

Update: a small PoC it's here: https://github.com/mudler/luet-makeiso

mudler:
I think we can close this. https://github.com/rancher-sandbox/cOS-toolkit/pull/56 was merged and it's into this direction. We had to keep only `xorriso` and `squashfs` due to technical limitations of the underlying library, but the code to build the iso natively is there ( it just fails when booting on EFI when building natively). 

This already simplified the build process and we got rid of the dependencies which made the UX a pitfall from the derivative perspective

PR #56, opened 2021-02-26
Use makeiso to generate ISO instead of geniso-isospec

mudler:
[luet-makeiso](https://github.com/mudler/luet-makeiso) is a golang binary having almost all needed to build the ISO (that I'll be happy to move to our own org once things are settled).
The only host requirement is mksquashfs and xorriso - until then that's the only
requirement.

`makeiso` is way more extensible with the new Golang structure, and
easier to embed into pipelines, definetly much more than the current extension. 

We now just use `jq` and `yq` to build the package list to build, and to generate the repository when running a build against the remote repositories, so derivatives don't need to install that anymore

Related to #27

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -9,11 +9,7 @@ cos-build() {
 
 docker run \
  -e FLAVOR=$flavor \
- -ti --privileged=true \
- --device=/dev/loop-control:/dev/loop-control \
- --device=/dev/loop0:/dev/loop0 \
- --cap-add SYS_ADMIN \
- --rm \
+ -ti --rm \
  -v /var/run/docker.sock:/var/run/docker.sock \
  -v $PWD:/cOS \
  cos-builder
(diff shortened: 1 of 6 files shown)


davidcassany (approved):

mudler:
Seems there are some issue with EFI and QEMU, will move to wip until figure out

mudler:
After a retake on it, the issue seems to be when constructing the iso itself. For the moment I've replaced it with `xorriso`, but that still gives us much more improvement as we generate the required images without any mount in the go binary, with a consequent drop of permissions when we run inside docker (and we can even generate the iso as user, but we loose permission bits).  So now with EFI works.

I've also embedded the repositories in the iso spec file, so we don't need to specify 2 yamls anymore. Now it's sufficient to add 
```yaml
repositories:
- ..
- ..
```

in the iso spec to specify where the packages are coming from. I've adapted the Makefile accordingly to drop the generation of the repository file. Some logic is still there (with `yq`) when we run `make iso` which instead generates it from the remote repositories

PR #58, opened 2021-03-01
Handle install and upgrades from single images

mudler:
Instead of having an ACTIVE and PASSIVE partition, we have ACTIVE and
PASSIVE .img file that we swap during upgrade. The flipping happens between `/cOS/active.img` and `/cOS/passive.img`, and a transition image which is being created during the upgrade process.

This is far more robust and less failure prone and doesn't involve
anymore grub installation during upgrades.

### DIFF
--- a/packages/installer/upgrade.sh
+++ b/packages/installer/upgrade.sh
@@ -82,18 +82,17 @@ upgrade() {
     mount_persistent
     ensure_dir_structure
 
-    mkdir -p /run/tmp
+    mkdir -p /usr/local/tmp/upgrade
     # FIXME: XDG_RUNTIME_DIR is for containerd, by default that points to /run/user/<uid>
     # which might not be sufficient to unpack images. Use /usr/local/tmp until we get a separate partition
     # for the state
     # FIXME: Define default /var/tmp as tmpdir_base in default luet config file
-    XDG_RUNTIME_DIR=/run/tmp TMPDIR=/run/tmp luet install -y $UPGRADE_IMAGE
+    XDG_RUNTIME_DIR=/usr/local/tmp/upgrade TMPDIR=/usr/local/tmp/upgrade luet install -y $UPGRADE_IMAGE
     luet cleanup
-    rm -rf /tmp/upgrade/var/tmp/*
-    rm -rf /run/tmp
-    umount $TARGET/oem || true
-    umount $TARGET/usr/local || true
-    umount $TARGET || true
+    rm -rf /usr/local/tmp/upgrade
+    umount $TARGET/oem
+    umount $TARGET/usr/local
+    umount $TARGET
 }
 
 switch_active() {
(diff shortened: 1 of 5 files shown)


packages/installer/upgrade.sh
```diff
@@ -84,50 +56,55 @@ find_upgrade_channel() {
 }
 
 mount_image() {
+    STATEDIR=/run/initramfs/isoscan
```
mudler, line 59: I'm not satisfied with this, but it gets automatically mounted there after boot. Should be ok for the time being

packages/installer/upgrade.sh
```diff
@@ -84,50 +56,55 @@ find_upgrade_channel() {
...
     mount_persistent
     ensure_dir_structure
+
+    mkdir -p /run/tmp
     # FIXME: XDG_RUNTIME_DIR is for containerd, by default that points to /run/user/<uid>
     # which might not be sufficient to unpack images. Use /usr/local/tmp until we get a separate partition
     # for the state
     # FIXME: Define default /var/tmp as tmpdir_base in default luet config file
-    XDG_RUNTIME_DIR=/var/tmp TMPDIR=/var/tmp luet install -y $UPGRADE_IMAGE
+    XDG_RUNTIME_DIR=/run/tmp TMPDIR=/run/tmp luet install -y $UPGRADE_IMAGE
```
mudler, line 91: This is tricky. Probably the best would be to use a temporary `/usr/local/tmp`

mudler: moved do `/usr/local/tmp` in https://github.com/mudler/cOS/pull/58/commits/f2048686547e6e23f6e0a87671236cfd1c20215a

packages/installer/upgrade.sh
davidcassany: The isoscan script is really simple, we should be capable to provide our own. We could even avoid the "scan" functionality which looks for the iso image at any available device, there is no need for that, we know the device that holds the image

mudler:
@davidcassany I gave another try, and reduced the HDD size to 10G just to play it safe. There are also https://github.com/mudler/cOS/pull/58/commits/75c40b38c094b0c9bcd6539086a2defdb9d7077a which is not related by I noticed the folder while iterating.

packages/installer/upgrade.sh
mudler: Shall we look over a replacement for isoscan? maybe we can do in a separate PR I guess the changes to the upgrade scripts would be really minimal

davidcassany: Yes we can do it in a separate PR, no hurries for now.

davidcassany (approved):
Looks good to me :+1:

mudler:
Opened #63 for the recovery partition follow-up

PR #59, opened 2021-03-01
Use new cloud-init sugar syntax

mudler:


packages/cos/os-release
```diff
@@ -0,0 +1,6 @@
+NAME="cOS"
+ID="cOS"
+ANSI_COLOR="0;32"
+BUG_REPORT_URL="https://github.com/mudler/cOS/issues"
```
mudler, line 4: We will change this, probably by auto-generating it from the package data. just using it as placeholder mostly

davidcassany (approved):
Looks good :+1:

PR #62, opened 2021-03-02
Require /oem mount point for cos-setup-fs.service

davidcassany:
This commit adds a dependency of cos-setup-fs.service with the /oem mount
point. In addition the timeout of the /oem and /usr/local is raised a
little so it is really unlikely they are missed on the first attempt to
mount them and their dependency to /etc is dropped. In fact we do not
care about what is mounted first /etc or /oem, we care that both are
mounted when cos-setup-fs.service is executed.


mudler (approved):

packages/base-conf/fstab
```diff
@@ -1,2 +1,2 @@
-LABEL=COS_PERSISTENT    /usr/local  auto    nofail,x-systemd.device-timeout=1ms,x-systemd.requires-mounts-for=/etc,x-systemd.wanted-by=local-fs.target    0  2
-LABEL=COS_OEM           /oem        auto    nofail,x-systemd.device-timeout=1ms,x-systemd.requires-mounts-for=/etc,x-systemd.wanted-by=local-fs.target    0  2
+LABEL=COS_PERSISTENT    /usr/local  auto    nofail,x-systemd.device-timeout=5s,x-systemd.wanted-by=local-fs.target    0  2
```
mudler, line 1: Just wondering if we will need the same for `COS_PERSISTENT`? As we let the user store `cloud-config` files also in `/usr/local/cloud-config`

davidcassany: Oh I did not consider this scenario, but yes, if we let the user place init files in `/usr/local/cloud-config` we should do the same there.

davidcassany: let me adapt it.

davidcassany: done

mudler:
Looks good here! :+1:

Issue #61, opened 2021-03-02
cOS variants


mudler:
It is possible already to specify a different base image in https://github.com/mudler/cOS/blob/master/packages/base/definition.yaml#L6 but it's not sufficient to build `cOS` on top of a `foo` distribution.

We have already a mechanism for templating globally, that can be exploited to provides the values in the specific templates ( e.g. https://github.com/mudler/cOS/tree/master/values ) .

The Issue is about creating a new `cOS` variant (ubuntu, debian, etc) , and attaching to the `CI`and build it aside the standard `cOS`. 

The values files in `values` corresponds to a specific variant which is being targeted while running `make build`. For now, we are sticking with a generic "amd64" which is empty only because the mechanism can be used also to provide multi-arch  builds, but we can reuse the same mechanism to provide different `cOS` flavors. 

See also: https://luet-lab.github.io/docs/docs/concepts/packages/templates/#interpolating-globally

davidcassany:
I am starting to have a look

PR #68, opened 2021-03-03
Reorder CI and generate QEMU images on GHA

mudler:
This PR introduces few changes:
- it introduces a very simple packer template, and a make target `make packer` to run packer and generate vbox and qemu images. It also generates now QEMU images on GHA for PRs and branches.
- it reorders the GHA pipelines. Now there is only one for PR and branches, and that one is ordered in separated jobs. Each one (build, iso and qemu) are uploading artifacts for each run. Only the pushes to `master` are published.
- it introduces a new make target `make iso` to generate iso from the remote repository state. this is used for a new nightly workflow that builds fresh images (introduced in this pr).
- renames `cOS-local.yaml` in `cOS.yaml` inside `iso/` . We might even consider to move this file top-level instead at some point (if we discover that is not useful to keep a folder hierarchy after #61 )


davidcassany (approved):
Looks good to me :+1:

PR #70, opened 2021-03-04
cOS reset

### LINKED ISSUE #63: Recovery partition to reset state
We should have a recovery partition which is not `COS_STATE` that can be used to recovery and reset `COS_STATE` and grub. It should also be possible to reset entirely by pruning `/oem` and `/usr/local` and start from scratch

Related to #23

### LINKED ISSUE #64: cos-reset
We should have a command, `cos-reset` that resets entirely `COS_STATE` (recreating images, installing grub) and optionally `COS_OEM` and `COS_PERSISTENT`. 

Relates to #23

mudler:
This PR introduces a recovery partition which is prepared during install (`COS_RECOVERY`). It can be used later to restore the system as it was at install time, by running `cos-reset` ( when booting from recovery ). 

An initial implementation allowed also to reset a system from the version available in the repositories, but after thinking about it I realized it's not really needed, as the user should be able to do that by running `cos-upgrade` (and use the recovery to get back to a _known_ working state).

It is also possible to update the recovery partition running `cos-upgrade` on the running system (not the recovery) by setting `UPGRADE_RECOVERY=true`. We require 8 gigs for the recovery partition to be sure to have enough
space to update it.

Fixes #64 and fixes #63

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Run `cos-installer <device>` to start the installation process. Remove the ISO a
 
 _Note_: `cos-installer` supports other options as well. Run `cos-installer --help` to see a complete help.
 
-## Upgrades / reset:
+## Upgrades:
 
 To upgrade the system, just run `cos-upgrade` and reboot.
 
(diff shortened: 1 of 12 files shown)


davidcassany (approved):
Looks good

PR #71, opened 2021-03-04
Parse boot cmdline in dotfile notation for cloud-init parameters

mudler:
Only available after #72 is merged.

It allows to pass by cloud-init options via boot cmdline, e.g.:

```
stages.fs[0].authorized_keys.root[0]=github:mudler
```

by piping /proc/cmdline to yip, enabling the dotnotation syntax

### DIFF
--- a/packages/cos-setup/cos-setup
+++ b/packages/cos-setup/cos-setup
@@ -4,7 +4,8 @@
 # it emit also "stage.before" and "stage.after" to able to hook
 # into different stages. E.g. if one depends on another for network setup
 STAGE="${1:-boot}"
-set -- $(cat /proc/cmdline)
+CMDLINE="$(cat /proc/cmdline)"
+set -- $CMDLINE
 for x in "$@"; do
     case "$x" in
         cos.setup=*)
@@ -27,4 +28,9 @@ for x in "$@"; do
         yip -s "$STAGE".after "${x#cos.setup=}"
         ;;
     esac
+done
+
+# Read cmdline from dotnotation and execute yip file:
+for s in "$STAGE".before "$STAGE" "$STAGE".after; do
+    echo $CMDLINE | yip -d -s "$s"
 done


packages/cos-setup/cos-setup
```diff
@@ -27,4 +28,9 @@ for x in "$@"; do
         yip -s "$STAGE".after "${x#cos.setup=}"
         ;;
     esac
+done
+
+# Read cmdline from dotnotation and execute yip file:
+for s in "$STAGE".before "$STAGE" "$STAGE".after; do
+    echo $CMDLINE | yip -d -s "$s"
```
davidcassany, line 35: I don't really understand it, what's the expected content of CMDLINE at this point? how yip behaves with this sort of pipe setup? I mean, is yip expecting a path on stdin? I'd expect something like `cos.setup=/oem` to set the path on the command line, but thus was already covered, isn't it?

davidcassany (commented):
Not really understanding it, sorry :/

packages/cos-setup/cos-setup
mudler: yes, indeed this is another use case which is not very well clear from the PR, my bad!

This is the scenario:

From grub, you can edit the boot cmdline and add e.g. `stages.network[0].authorized_keys.root[0]=github:mudler` to add that github key to the root user, or set whatever we expose in the cloud-init config


The implementation side is a bit tricky, `yip` composes a yaml file from a sequence of dotnotations, e.g.: `stages.network[0].authorized_keys.root[0]=github:mudler`, is equivalent to: 
```yaml
stages:
  boot:
  - authorized_keys:
     root:
     - github:name
```
and then it parses it as a cloud-init. In this case, we read the whole cmdline, and we pipe it to `yip`, enabling the dotnotation syntax ( with `-d` ). The cmdline arguments become all a yaml then which is parsed in search for cloud-init keywords.

davidcassany (approved):
Looks good

PR #74, opened 2021-03-05
Add a Fedora cOS variant

### LINKED ISSUE #61: cOS variants

davidcassany:
This commit adds a variant for cOS based on Fedora. Small refactor in
Makefile has been done to allow different flavors. Each flavor maps to a
set to an specific yaml file including the required additional values.
The default flavor is set to opensuse.

In addition the kernel package has been dropped and included as part of
the base distro package, this is done to facilitate templating across
different distros.

Finally there is an unrelated update for the mount point of the ephemeral
overlay volume that gets mounted in /run/overlay now instead of
/data.overlay.

Fixes #61


Makefile
```diff
@@ -43,23 +43,24 @@ endif
 endif
 
 clean:
-	 rm -rf $(ROOT_DIR)/build/ $(ROOT_DIR)/.qemu
+	 rm -rf $(DESTIBNATION) $(ROOT_DIR)/.qemu
```
mudler, line 46:
```suggestion
	 rm -rf $(DESTINATION) $(ROOT_DIR)/.qemu
```

Makefile
```diff
@@ -13,9 +12,10 @@ export TREE?=$(ROOT_DIR)/packages
 BUILD_ARGS?=--pull --no-spinner --only-target-package
 
 VALIDATE_OPTIONS?=-s
-ARCH?=amd64
-REPO_CACHE?=raccos/$(ARCH)
-FINAL_REPO?=raccos/releases-$(ARCH)
+FLAVOR?=opensuse
```
mudler, line 15: :+1:

Makefile
```diff
@@ -80,30 +81,30 @@ publish-repo:
     --type docker
 
 serve-repo:
-	LUET_NOLOCK=true $(LUET) serve-repo --port 8000 --dir $(ROOT_DIR)/build
+	LUET_NOLOCK=true $(LUET) serve-repo --port 8000 --dir $(DESTIONATION)
```
mudler, line 84:
```suggestion
	LUET_NOLOCK=true $(LUET) serve-repo --port 8000 --dir $(DESTINATION)
```

mudler (approved):
Looks good! it seems although that the ci fails for the missing `values` files

I'll try to hook later the CI to build ISO/images for the fedora spin

mudler:
Just a later thought: we have also to configure the upgrade channel somehow, at the moment is configured in the cloud-config package, but would make sense to let it configure in the `values`

Makefile
davidcassany: ouch :man_facepalming:  fixed now

Makefile
davidcassany: Oh man, I solved the conflicts on Makefile too fast :sweat:

davidcassany:
> Looks good! it seems although that the ci fails for the missing `values` files
> 
> I'll try to hook later the CI to build ISO/images for the fedora spin

Damn I forgot to commit the values files :sweat:

Makefile
```diff
@@ -13,9 +12,10 @@ export TREE?=$(ROOT_DIR)/packages
 BUILD_ARGS?=--pull --no-spinner --only-target-package
 
 VALIDATE_OPTIONS?=-s
-ARCH?=amd64
-REPO_CACHE?=raccos/$(ARCH)
-FINAL_REPO?=raccos/releases-$(ARCH)
+FLAVOR?=opensuse
+DESTINATION?=$(ROOT_DIR)/build
```
davidcassany, line 16: JFYI I'd would have made the this line like
```
DESTINATION?=$(ROOT_DIR)/build-$(FLAVOR)
```
So I could easily build the iso from opensuse or fedora without having to rebuild the repository. However the geniso extension is dependent on the  `$(ROOT_DIR)/build` path. No big deal, but something to keep in mind for #56

mudler: Is geniso using it explictly or it's caused by https://github.com/mudler/cOS/blob/master/iso/cOS.yaml#L20 ?

davidcassany: Oh, I guess you are right, I did not look into that. I saw it was failing in one of the geniso scripts but I did not check where the path was defined.

Issue #24, opened 2021-01-15
OEM customizations


mudler:
A mechanism to extend cloud-init configuration mechanism ( see #4 ) for OEM customizations. This includes also executing cloud-init yaml on boot from a persistent folder, like `/oem` or `/usr/local/oem` for example.

The installation process should allow optional cloud-init config to be configured as persistent as well.

mudler:
Closing for now, as we have an oem mechanism. We will create follow-ups when necessary

PR #77, opened 2021-03-08
Move system OEM files to /system/oem

mudler:
Those are defaults that should be never override directly from the user,
but rather be immutable. Previously they were replaced during upgrades,
but this has the disadvantage that fallback and standard boot share the
same OEM. In this way, each image ships their own oem which are
immutable , but will still leave /oem as persistent for oem changes.

Related to #24


mudler:
Superseded by #79

PR #78, opened 2021-03-08
Add audit package for Fedora variant

davidcassany:
This commit adds `audit` package for Fedora based variant in order to
avoid audit logs to be redirected to stderr by default.


davidcassany:
@mudler PR is to avoid very annoying messages on the console after login on a Fedora based cOS. `audit` package includes the audit configuration, without it looks like all messages are simply redirected to stderr and it is really annoying having constantly spurious asynchronous messages on the console.

mudler (approved):
Looks good!

PR #79, opened 2021-03-09
Allow to upgrade from single docker images

mudler:
supersedes #77 as includes it

This allows also to upgrade directly from docker images specified by the user, instead of the regular release channel.

The mechanism is simple, instead of calling `luet install`, we just unpack the image specified by CLI or config file. The config file is an env file which is populated by the cloud-init OEM (`02_upgrades.yaml`) so in this way can be easily override from runtime. I would leave initentionally the commit that first adds up the golang utility just for history, in case we are going back to that implementation.

The PR is many fold,  to sum up:
- It fixes the default `02_upgrades.yaml` propagating the flavor to the release channel configuration file
- It simplifies the `Makefile`: now `ISO` for the various subcommand is implied to the iso found in the current folder, and append `--live-output` to luet to follow the build
- `cos-upgrade` accepts now a Docker image optionally as argument (`--docker-image <image>`)
- `cos-upgrade` accepts now `--recovery` to upgrade the recovery partition
- CI: adds the fedora pipeline


packages/cloud-config/build.yaml
```diff
@@ -3,6 +3,11 @@ requires:
...
-- mkdir -p /oem
-- cp -rfv yipfiles/*.yaml /oem
-- chmod -R 600 /oem
\ No newline at end of file
+- mkdir -p /system/oem
+- cp -rfv yipfiles/*.yaml /system/oem
+- chmod -R 600 /system/oem
+{{if .Values.is_opensuse}}
+- sed -i 's/:FLAVOR:/opensuse/g' /system/oem/02_upgrades.yaml
+{{else}}
```
davidcassany, line 11: I would use `{{else if .Values.is_fedora}}` here to be consistent. Other packages have no default, you must either select fedora or opensuse. So if you do not provide a `--values` with the corresponding flavor the it will just fail.

davidcassany (approved):
I just have a comment regarding the opensuse vs fedora templating.

packages/cloud-config/build.yaml
mudler: right! fixed it :+1:

PR #85, opened 2021-03-17
Signed images

mudler:
@davidcassany I had to push already a cos version with the signed images in order to test them thru the CI.

This PR mostly is changing CI bits. The notable change is just `verify: true` set in `/etc/luet/luet.yaml` and the addition of tests to cover the signed images.

From the CI side, it moves the publish steps after tests are passed , and adds the release steps (untested, still)

Fixes: #25

PR #86, opened 2021-03-17
Immutable-rootfs dracut module

davidcassany:
This commit adds a dracut module to control the immutable rootfs layout.

The dracut module can be controlled by the `rootfs` stage in cloud init
files or by kernel command line parameters.

It also runs the `initramfs` cloud-init stage before switching root but
after having the immutable rootfs set.


packages/cloud-config/oem/00_rootfs.yaml
```diff
@@ -0,0 +1,15 @@
...
+#
+# This file is part of cOS and will get reset during upgrades.
+#
+# Before you change this file manually,
+# consider copying this file to /usr/local/cloud-config or
+# copy the file with a prefix starting by 90, e.g. /oem/91_custom.yaml
+name: "Rootfs Layout Settings"
+stages:
+  rootfs:
+    - environment_file: /tmp/cos-layout
```
mudler, line 11: It's not really important at the moment, but for the long run maybe we should keep all of those under `/etc`, even under our own folder (we already have `/etc/cos-upgrade-image` which lets configure the upgrade image, but feels like we need a proper place for those)

mudler:
Looks good here :+1: I'm not approving just because this PR needs a cos version bump and a CI test pass.

#85 have already pushed (and published) a new version of `cOS` which is ahead of master in order to test the CI and image signing process (and tests, too). So we have to wait until that gets merged first

packages/cloud-config/oem/00_rootfs.yaml
davidcassany: This is a temporary file that is only used at boot time to read `rootfs` stage contents, that's why I thought about `/tmp`. Probably even somewhere under `/run` could make sense, so after booting it could help for debugging or verifying what was actually applied. In any case note that this gets created inside in initramfs and only used in that context, so it won't ever be part of the image itself.

I agree we should think up for a better place. Probably `/run/cos/cos-layout.env`? So this way path applies for both, before and after the switch-root.

davidcassany:
> Looks good here +1 I'm not approving just because this PR needs a cos version bump and a CI test pass.
> 
> #85 have already pushed (and published) a new version of `cOS` which is ahead of master in order to test the CI and image signing process (and tests, too). So we have to wait until that gets merged first

Yes I will rebase on top of 85, gonna be easier :+1:

packages/cloud-config/oem/00_rootfs.yaml
mudler: It's indeed fine for the moment being, was just thinking out loud, because might not be the only one that needs a proper placement.  `/run/cos/*` sounds even better :+1:

mudler:
I think the vagrant failures are due to the fact that the provisioner expected `/oem` to be mounted after a `partprobe`, which previously was happening to work - one way to fix it is to mount that partition manually and copy over the `vagrant.yaml` file

davidcassany:
> I think the vagrant failures are due to the fact that the provisioner expected `/oem` to be mounted after a `partprobe` right after install, which previously was happening to work - one way to fix it is to mount that partition manually and copy over the `vagrant.yaml` file

I think the problem is that the `/run/cos/cos-layout.env` path included a new directory structure `/run/cos` that no one created it in advance, meaning that the cos-layout.env file was not created hence the layout was not configured and `/oem` was not present. 

All this falls in the area of how to detect boot issues...

mudler (dismissed):

mudler:
That's a good point indeed, we could also change cloud-init to create the entire folder parent if not present, but I was  refering to: 

``` ==> virtualbox-iso: Warning: Unable to open /dev/sr0 read-write (Read-only file system).  /dev/sr0 has been opened read-only.
==> virtualbox-iso: cp: cannot create regular file '/oem/vagrant.yaml': No such file or directory
==> virtualbox-iso: Provisioning step had errors: Running the cleanup provisioner, if present...
==> virtualbox-iso: Cleaning up floppy disk...
```

In the CI logs, It looks https://github.com/rancher-sandbox/cOS/blob/master/packer/images.json#L77 doesnt work anymore (which is fine) but might require a bit more logic on the packer provisioner

davidcassany:
> In the CI logs, It looks https://github.com/rancher-sandbox/cOS/blob/master/packer/images.json#L77 doesnt work anymore (which is fine) but might require a bit more logic on the packer provisioner

Right just added ta manual mount to the packer. I believe previously it was working because in live boots we used the have mount unit in systemd for `/oem` and `/usr/local` which were failing, so on `partprobe` systemd was detecting new devices and then mounting them.

mudler (approved):

mudler (approved):

PR #87, opened 2021-03-18
tests: use ginkgo from CLI, add FlakeAttempts and reset on each upgrade tests

mudler:
This should stabilize single tests by starting and finishing with a clean state each `It` block using`Reset()` before each one of them.

It also runs the suite tests with `ginkgo` instead of `go test`, so we can pass by few arguments overall that will help out:

- we make now the tests fail fast
- if there are failures, they are attempted 3 times
- more verbosity and progress printed to the terminal

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -289,8 +289,10 @@ jobs:
 
       - name: Run tests 🔧
         run: |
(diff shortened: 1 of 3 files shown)


davidcassany (approved):
Looks good

Issue #31, opened 2021-01-28
Grub install without chroot in the installer


mudler:
https://github.com/mudler/cOS/blob/85e764da2ab6698d9c3d216de0e55834a8ee25d7/packages/installer/installer.sh#L173

We are chrooting, while it shouldn't be necessary. Let's have a closer look and get rid of the chroot call

PR #90, opened 2021-03-19
Unflake installation

mudler:
The active img file which contains the partition where we did the setup was previously unmounted cleanly, which causes the subsequents copy of the image flaky as they attempt to copy from a live mounted image.

### DIFF
--- a/packages/installer/installer.sh
+++ b/packages/installer/installer.sh
@@ -18,11 +18,11 @@ umount_target() {
         umount ${TARGET}/oem || true
         umount ${TARGET}/usr/local || true
         umount ${TARGET}/proc || true
-        umount ${TARGET}/dev || true
-        umount ${TARGET}/sys || true
+        umount -R ${TARGET}/dev || true
+        umount -R ${TARGET}/sys || true
         umount ${TARGET}/boot/efi || true
         umount ${TARGET}/boot/grub2 || true
-        umount ${TARGET} || true
+        umount ${TARGET}
     fi
 }
 
@@ -31,7 +31,7 @@ cleanup2()
     sync
     umount_target
     umount ${STATEDIR} || true
-    umount ${RECOVERY} || true
+    umount ${RECOVERYDIR} || true
 }
 
 cleanup()
@@ -116,7 +116,7 @@ do_format()
         parted -s ${DEVICE} mkpart primary ext4 18050MB 100% # persistent
         parted -s ${DEVICE} set 2 ${BOOTFLAG} on
     fi
-   
+
     partprobe ${DEVICE} 2>/dev/null || true
     sleep 2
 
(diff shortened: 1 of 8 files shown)


tests/Vagrantfile
```diff
@@ -0,0 +1,15 @@
...
+# vi: set ft=ruby :
+
+Vagrant.configure("2") do |config|
+  config.vm.box = "cos"
+
+  config.vm.provider :virtualbox do |vb|
+    vb.customize ["modifyvm", :id, "--memory", "4098", "--cpus", "2"]
+  end
+
+  config.vm.provider :libvirt do |libvirt|
```
mudler, line 11: I've also added the accell setup for libvirt with vbox here, but didn't tested

tests/sut_test.go
```diff
@@ -106,14 +106,17 @@ func (s *SUT) BootFrom() int {
 }
 
 func (s *SUT) EventuallyConnects(t ...int) {
-	dur := 500
+	dur := 120
```
mudler, line 109: 120 should be enough, so at maximum in case of failure we wait 6 mins

packages/installer/installer.sh
```diff
@@ -265,8 +265,12 @@ install_grub()
     mkdir ${TARGET}/sys || true
     mkdir ${TARGET}/tmp || true
     mount -t proc proc ${TARGET}/proc
+
     mount --rbind /dev ${TARGET}/dev
+    mount --make-rslave ${TARGET}/dev
+
     mount --rbind /sys ${TARGET}/sys
```
davidcassany, line 272: I usually do a `mount -t sysfs /sys ${TARGET}/sys` probably this avoind then having to `umount -R ${TARGET}/sys`.

packages/installer/installer.sh
```diff
@@ -18,11 +18,11 @@ umount_target() {
...
         umount ${TARGET}/usr/local || true
         umount ${TARGET}/proc || true
-        umount ${TARGET}/dev || true
-        umount ${TARGET}/sys || true
+        umount -R ${TARGET}/dev || true
+        umount -R ${TARGET}/sys || true
         umount ${TARGET}/boot/efi || true
         umount ${TARGET}/boot/grub2 || true
-        umount ${TARGET} || true
+        umount ${TARGET}
```
davidcassany, line 25: Probably we could consider doing only a single `umount --recursive ${TARGET}` to unmount everything. In the past I also noted that in some cases the lazy umount helps so probably:

```
umount -l --recursive ${TARGET}
```

does the job in a single command and there is not need to umount manually all of them.

davidcassany (approved):
Looks good to me :+1: I just added a couple of comments about the mount/umount commands as additional ideas in case there are still some failures.

packages/installer/installer.sh
mudler: I did avoided to use `-l` here because when we umount $TARGET we want to be sure  it's not there, because its the image file that later on gets copied (hence the fix here). If something else umounts it later then it would be too late, because we copy it over in the STATE partition

packages/installer/installer.sh
mudler: yup that works too indeed, but generally I'd like to get rid of this block entirely, it is mostly there because of #31

PR #98, opened 2021-03-31
Add major packages needed to run longhorn and k3s

### LINKED ISSUE #30: Packages to add to the rootfs image
For the ones coming from zypper, should be as simple as adding them here: https://github.com/mudler/cOS/blob/85e764da2ab6698d9c3d216de0e55834a8ee25d7/packages/tumbleweed/definition.yaml#L5

- [ ]        conntrack-tools
- [x]        curl
- [x]        dosfstools
- [x]        e2fsprogs
- [ ]        findutils
- [ ]        haveged
- [x]        iproute2
- [ ]        iptables
- [ ]        lsscsi
- [ ]        lvm2
- [ ]        mdadm
- [ ]        multipath-tools
- [ ]        nfs-utils
- [ ]        open-iscsi
- [ ]        open-vm-tools
- [ ]        python-azure-agent*
- [ ]        qemu-guest-agent
- [ ]        rng-tools
- [ ]        rancher-agent

mudler:
This PR addresses few points:

- Adds required packages to run k3s/longhorn (Fixes #30 ). The ISO size it's not even that much bigger, now it's around 400MB (~100MB)
- Adds a very simple cli (`cos-feature`) to enable/disable pre-defined cloud configs shipped with cOS. It can be used to quickly enable k3s/longhorn/harvester locally for testing/development. I've moved the vagrant user setup there, so can be also enabled easier from CI
- Splits the tests to separate test suites. This makes the flow more easier to follow and easier to catch errors on the CI. It also adds a simple test suite to test k3s after install and the recovery test suite
- Moves the image format to ext2 in place of ext4. During this change subset, I've found out that the tests flakiness were caused by the installation which was not working correctly all the times. A journal process was left behind and prevents the loopmount to be detached, causing an unclean journaling when the machine gets reboot after install. For now until we address #31 I wouldn't mind to switch to ext2 as we don't use the images for rw access anyway.

### DIFF
--- a/packages/cos-features/features/harvester.yaml
+++ b/packages/cos-features/features/harvester.yaml
@@ -1,0 +1,43 @@
+stages:
+   network:
+     - name: "Setup k3s"
+       directories:
+       - path: "/usr/local/bin"
+         permissions: 0755
+         owner: 0
+         group: 0
+       - path: "/run/harvester"
+         permissions: 0755
+         owner: 0
+         group: 0
+       commands:
+       - |
+            curl -sfL https://get.k3s.io | \
+            INSTALL_K3S_VERSION="v1.20.4+k3s1" \
+            INSTALL_K3S_EXEC="--data-dir /usr/local/rancher/k3s" \
+            sh -
+       systemctl:
+        start:
+        - iscsid
+     - name: "Setup Longhorn"
+       commands:
+       - |
+          curl https://raw.githubusercontent.com/longhorn/longhorn/v1.1.0/deploy/longhorn.yaml | \
+          sed -e 's/#- name: KUBELET_ROOT_DIR/- name: KUBELET_ROOT_DIR/g' -e 's$#  value: /var/lib/rancher/k3s/agent/kubelet$  value: /var/lib/kubelet$g'  | \
+          sed -e 's|default-data-path:|default-data-path: /usr/local/rancher/longhorn|g' | \
+          k3s --data-dir /usr/local/rancher/k3s kubectl apply -f -
+       commands:
+       - |
+          if [ ! -e "/usr/local/bin/helm" ]; then
+            curl https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 | bash
+          fi
+     - name: "Harvester"
+       commands:
+       - |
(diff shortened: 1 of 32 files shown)


packages/cos-features/features/harvester.yaml
```diff
@@ -0,0 +1,43 @@
+stages:
+   network:
+     - name: "Setup k3s"
+       directories:
+       - path: "/usr/local/bin"
+         permissions: 0755
+         owner: 0
+         group: 0
+       - path: "/run/harvester"
```
davidcassany, line 9: Shouldn't it be part of the harvester step? I think it is good that `Setup k3s` is equivalent for all. So in the future if we want to add some sort of yaml dependency structure to pull _stages_ from another one it will be straight forward to adapt these files :)

packages/immutable-rootfs/30cos-immutable-rootfs/module-setup.sh
```diff
@@ -21,7 +21,7 @@ install() {
     declare moddir=${moddir}
     declare systemdutildir=${systemdutildir}
     inst_multiple \
-        mount mountpoint yip cos-setup sort
+        mount mountpoint yip cos-setup sort rmdir
```
davidcassany, line 24: I think this is some sort of leftover, as I don't see rmdir being used anywhere.

davidcassany (dismissed):
Looks good to me, the feature concept is cool :+1: just a couple of little comments

packages/immutable-rootfs/30cos-immutable-rootfs/module-setup.sh
mudler: true, but it actually failed otherwise while finding the images after install during the isoscan step. It complained about missing rmdir and prevented to boot

packages/cos-features/features/harvester.yaml
mudler: Good catch! fixed in https://github.com/rancher-sandbox/cOS-toolkit/pull/98/commits/37d5e07984bf4aa995ef28260d8f55f9e5dcaa86 :+1:

PR #100, opened 2021-04-02
Enable GRUB fallback

### LINKED ISSUE #69: Automatically fallback in case of boot failure
This is a follow-up card from #8 . Seems the mechanism isn't correctly working and in place for GRUB. 

We should by default fallback to the COS_PASSIVE grub entry, and if that fails, enter to the recovery partition #63 .

Also, we should configure https://github.com/systemd/systemd/blob/a2689fa5fcbfa31eca51db1b2b4a808835429332/docs/AUTOMATIC_BOOT_ASSESSMENT.md automatic boot assesment too if possible.

## Action items

- [ ] Setup grub for fallback in case of boot failure to the `COS_PASSIVE` boot entry. If it possible, in case of failure, switch automatically to recovery.
- [ ] Setup if possible systemd Automatic Boot assessment features

mudler:
Also adds a test suite for fallback tests and re-organizes Makefile
accordingly.

I'll create a separate follow-up for the systemd story

Fixes #69

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -150,6 +150,19 @@ test-clean:
 	cd $(ROOT_DIR)/tests && vagrant destroy || true
 	vagrant box remove cos || true
 
-test: test-clean tests/Vagrantfile prepare-test
-	cd $(ROOT_DIR)/tests && ginkgo $(GINKGO_ARGS) ./smoke ./upgrades ./features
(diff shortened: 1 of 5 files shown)


davidcassany (approved):
Looks good!

PR #101, opened 2021-04-06
Add ubuntu flavor

mudler:
This PR adds an ubuntu flavor

ubuntu

There are few changes @davidcassany to the layout , that's why I would merge as it is if tests are passing. It also fixes the #31 bug as now doesn't require chrooting anymore

### DIFF
--- a/packages/base/build.yaml
+++ b/packages/base/build.yaml
@@ -2,11 +2,11 @@ image: "{{.Values.image}}"
 
 steps:
 {{if eq .Values.distribution "opensuse" }}
-- zypper in -y --no-recommends {{.Values.leap_packages}} {{.Values.common_packages}} {{.Values.kernel_package}}
+- zypper in -y --no-recommends {{.Values.packages}} {{.Values.kernel_package}}
 - zypper cc
 {{else if eq .Values.distribution "fedora" }}
 - echo "install_weak_deps=False" >> /etc/dnf/dnf.conf
-- dnf install -y {{.Values.fedora_packages}} {{.Values.common_packages}} {{.Values.kernel_package}}
+- dnf install -y {{.Values.packages}} {{.Values.kernel_package}}
 - dnf clean all
 - |
     kernel=$(ls /boot/vmlinuz-* | head -n1) && \
--- a/packages/base/build.yaml
+++ b/packages/base/build.yaml
@@ -1,10 +1,10 @@
 image: "{{.Values.image}}"
 
 steps:
-{{if .Values.is_opensuse}}
+{{if eq .Values.distribution "opensuse" }}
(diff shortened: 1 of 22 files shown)


packages/base/build.yaml
```diff
@@ -1,15 +1,31 @@
...
 - echo "install_weak_deps=False" >> /etc/dnf/dnf.conf
-- dnf install -y {{.Values.fedora_packages}} {{.Values.common_packages}} {{.Values.kernel_package}}
+- dnf install -y {{.Values.packages}} {{.Values.kernel_package}}
 - dnf clean all
 - |
     kernel=$(ls /boot/vmlinuz-* | head -n1) && \
     ln -sf "${kernel#/boot/}" /boot/vmlinuz
+{{else if eq .Values.distribution "ubuntu" }}
+# Fun: https://bugs.launchpad.net/ubuntu/+source/grub2/+bug/1879558
+- apt-get update && apt-get install -y {{.Values.packages}} {{.Values.kernel_package}} || true
```
mudler, line 23: I'm trying to see if I can get rid of the `|| true` here now. At that time I was hitting https://bugs.launchpad.net/ubuntu/+source/grub2/+bug/1879558 , and post-install hooks were failing due to both grub-386 and amd64 missing

davidcassany: I believe I already went through this issue. Grub debian packages are so broken and it's deb sepc is just insane. At time I realized one option was to split grub packages installation into two transactions (bios boot first and then efi boot on a second round). Alternatively, for Focal version I found that installing this exact list of grub packages was not failing in a single transaction and was pulling the appropriate dependencies.

```
grub-pc-bin
grub2-common
grub-efi-amd64-bin
```

So from your case I would drop `grub2` package from the list (which is already duplicated by the way)

values/ubuntu.yaml
```diff
@@ -0,0 +1,28 @@
+image: ubuntu:20.10
+distribution: "ubuntu"
+
+packages: >-
+    systemd
+    grub-pc-bin
+    grub-efi-amd64-bin
+    grub2
```
davidcassany, line 8: I would remove it

values/ubuntu.yaml
```diff
@@ -0,0 +1,28 @@
...
+    iproute2
+    squashfs-tools
+    parted
+    dracut
+    dracut-live
+    e2fsprogs
+    dosfstools
+    coreutils
+    multipath-tools
+    grub2
```
davidcassany, line 19: duplicate

davidcassany (commented):
Looks good to me, just a comment regarding the gurb packages for ubuntu.

Itxaka (approved):
Looks good to me, but I dont know about the grub changes that much so....

PR #103, opened 2021-04-12
[WIP] Drop isoscan use and limit dracut hooks usage

### LINKED ISSUE #65: Replace isoscan with a custom dracut script
The isoscan script is really simple, we should be capable to provide our own. We could even avoid the "scan" functionality which looks for the iso image at any available device, there is no need for that, we know the device that holds the image

_Originally posted by @davidcassany in https://github.com/mudler/cOS/pull/58#discussion_r585442021_

while isoscan works, it is better to provide a script that doesn't scan at all, as we already know were the img files are (`COS_STATE`).
Also, isoscan forces `cos-upgrade` to remount the `COS_STATE` partition `rw` as it is already mounted in `/run/initramfs/isoscan` after boot.

davidcassany:
This commit it two fold:

* Drop isoscan use in favor of a self provided script that is not
  devoted to ISO images and offers some further flexibility. Fixes #65

* Drop most dracut hooks in favor of early boot systemd services. This
  is relevant in case we want or need to recode the layout tools so
  we are not tied to shell scripts.


davidcassany:
This PR changes the boot command on grub to not use the isoscan script (part of dmsquash module for live ISOs) in favor of a similar version provided by us. This offers us flexibility to decide how we want to loop back the rootfs images so we are not limited to the behavior of isoscan, which was thought having a pretty different use case in mind.

So the upgrade tests will not work on this PR, this is an upgrade breaking PR. Also a good chance to think if we want to consider this sort of use cases in the future, I mean, what if an upgrade requires changes on the grub configuration? Should we provide some strategy for those cases?

packages/immutable-rootfs/30cos-immutable-rootfs/cos-setup-rootfs.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=cOS system early rootfs setup
+DefaultDependencies=no
+After=initrd-root-fs.target dev-disk-by\x2dlabel-COS_OEM.device
```
davidcassany, line 4: I'd like to avoid having this dependency to `initrd-root-fs.target`. This is required because after the installation the only place where we have cloud-init files is in `/system/oem` which is part of the rootfs. I consider this a limitation, because it prevents us to run some cloud-configs before the rootfs is mounted. I think it would be nice to copy the `rootfs` stage of cloud init files to `COS_OEM` volume at installation time, this way we could have the `rootfs` stage that runs before mounting the actual root. I do not have a clear use case now for that, but I think this could be relevant for the semipersistent /etc  for example, it could be the place to hook before mounting the disk (COS_OEM could be some USB stick or remote location).

packages/immutable-rootfs/30cos-immutable-rootfs/module-setup.sh
```diff
@@ -20,13 +20,24 @@ installkernel() {
...
+    inst_script "${moddir}/cos-mount-layout.sh" "/sbin/cos-mount-layout"
+    inst_script "${moddir}/cos-loop-img.sh" "/sbin/cos-loop-img"
+    inst_simple "${moddir}/cos-immutable-rootfs.service" \
+        "${systemdsystemunitdir}/cos-immutable-rootfs.service"
+    inst_simple "${moddir}/cos-setup-rootfs.service" \
+        "${systemdsystemunitdir}/cos-setup-rootfs.service"
+    mkdir -p "${initdir}/${systemdsystemunitdir}/initrd-fs.target.requires"
+    ln_r "../cos-immutable-rootfs.service" \
+        "${systemdsystemunitdir}/initrd-fs.target.requires/cos-immutable-rootfs.service"
+    ln_r "../cos-setup-rootfs.service" \
```
davidcassany, line 40: I added the work in progress mostly because I think this service should be part of `cos-setup` package, I just did not have time enough to do the change when I worked on that two weeks ago.

PR #120, opened 2021-04-26
[WIP]ci: Separate CI build by flavor

Itxaka:
To do this a shared go template is used and a
values yaml file to contain values when executing
the template

A small golang utility is provided so it can be used to
generate the different gihub actions yaml files and
expanded in the future in case we need extra actions
or values for flavours

The generated yamls for both opensuse and fedora are added
as well, while the old one with both in the matrix key is
removed


Itxaka:
This blocked until we got the other patches for ci merged, makes no sense to rebase it until then.

Itxaka:
I dont think this is necessary now, we can reopen it later if needed

PR #122, opened 2021-04-26
ci: upload images in the same job

Itxaka:
There is no need to run a separate job for generating the vagrant
images once we have generated a provider images. The post processors
work based off the output from the builders so vagrant creates a box
file based of the qemu/vbox image and compress just compressed the
qemu/vbox image

They are both generated at the same time so we can use that to our
advantage and upload the artifacts at the same time

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -171,45 +171,17 @@ jobs:
 
       - name: Build QEMU Image 🔧
         run: |
-          PACKER_ARGS="-var='accellerator=none' -var='sleep=5m' -only qemu" make packer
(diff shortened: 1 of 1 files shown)


Itxaka:
you can test this by manually running 

```
packer build -timestamp-ui -var='accellerator=none' -var='iso=/home/itxaka/projects/cOS-toolkit/cOS-0.20210426.iso' -var='feature=vagrant' -only qemu images.json
```

and then checking the outputs. The tar.gz file contains a qcow image and the .box image (once un-gziped) contains a box.img and vagrant config file

mudler (approved):
I think we had those because originally we thought to also give away images without the vagrant user set, but right now indeed we don't need them anymore :+1:  Looks good!

PR #123, opened 2021-04-26
ci: Separate tests to run them in parallel

Itxaka:
Instead of running the test all at the same time, we can
separate them using a matrix so they run in parallel

This patch also separates the upgrade tests into using
signed and unsigned images so we can also run those in
parallel, as they are the ones taking most of the time

### DIFF
--- a/tests/upgrades-images-signed/tests_suite_test.go
+++ b/tests/upgrades-images-signed/tests_suite_test.go
@@ -1,0 +1,13 @@
+package cos_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestTests(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "cOS Upgrate test Suite - Images signed")
+}
--- a/tests/upgrades-images-signed/upgrade_test.go
+++ b/tests/upgrades-images-signed/upgrade_test.go
@@ -1,0 +1,56 @@
+package cos_test
+
+import (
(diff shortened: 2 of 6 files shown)


mudler (approved):
Looks good here :+1:

tests/upgrades-images-signed/tests_suite_test.go
```diff
@@ -9,5 +9,5 @@ import (
 
 func TestTests(t *testing.T) {
 	RegisterFailHandler(Fail)
-	RunSpecs(t, "cOS Upgrate test Suite")
+	RunSpecs(t, "cOS Upgrate test Suite - Images signed")
```
mudler, line 12:
```suggestion
	RunSpecs(t, "cOS Upgrate test Suite - Signed images")
```

tests/upgrades-images-signed/upgrade_test.go
```diff
@@ -0,0 +1,56 @@
+package cos_test
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/rancher-sandbox/cOS/tests/sut"
+)
+
+var _ = Describe("cOS Upgrade tests - Images signed", func() {
```
mudler, line 9:
```suggestion
var _ = Describe("cOS Upgrade tests - Signed images", func() {
```

Itxaka:
I'll change the names in a post commit as to not lose the approval :)

mudler:
No worries :) I think we don't dismiss approvals at the moment :+1:

PR #152, opened 2021-05-04
Fix testing for command dependencies

kkaempf:
The old way (using the output of 'which' as a target) was essentially
a no-op.

Signed-off-by: Klaus Kämpf <kkaempf@suse.de>

### DIFF
--- a/make/Makefile.iso
+++ b/make/Makefile.iso
@@ -14,6 +14,11 @@ ifneq ($(strip $(REPO_CACHE)),)
 	BUILD_ARGS+=--image-repository $(REPO_CACHE)
 endif
 
+MKSQUASHFS?=$(shell which mksquashfs 2> /dev/null)
+ifeq ("$(MKSQUASHFS)","")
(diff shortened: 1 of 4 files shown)


make/Makefile.iso
```diff
@@ -14,6 +14,11 @@ ifneq ($(strip $(REPO_CACHE)),)
 	BUILD_ARGS+=--image-repository $(REPO_CACHE)
 endif
 
+MKSQUASHFS?=$(shell which mksquashfs 2> /dev/null)
+ifeq ("$(MKSQUASHFS)","")
+MKSQUASHFS="/usr/bin/mksquashfs"
```
mudler, line 19: similar thing should be done for `xorriso` too :+1:

mudler (approved):

Itxaka (approved):
nice!

make/Makefile.iso
kkaempf: Let me merge this one first and continue testing. I'm sure I'll find even more :wink:

PR #161, opened 2021-05-05
Improve iso targets in makefile

kkaempf:
iso target improvements
    
- use full path to `yq` binary
- don't overwrite existing .iso file
- make sure ISO file exists before running packer


kkaempf:
Build checks pass, (usual :-/) timeout during test.

make/Makefile.iso
```diff
@@ -38,7 +38,9 @@ $(MKSQUASHFS):
 	@echo "'mksquashfs' not found, install 'squashfs' package."
 	@exit 1
 
-create-repo: $(LUET) $(DESTINATION)
+create-repo: $(LUET) $(DESTINATION) $(DESTINATION)/tree.tar.zst
+
+$(DESTINATION)/tree.tar.zst:
```
mudler, line 43: thing is, if we iterate locally and run multiple create-repos this won't work. It's not a one-off, but you could run create-repo more than one, maybe a `.PHONY` here?

kkaempf: Hmm, I think a PHONY would do the same like reverting this diff.
If 'create-repo' is supposed to be run multiple times, let me change the dependencies of `local-iso` instead.

mudler (approved):

PR #166, opened 2021-05-06
[WIP][DONOTMERGE] Add mtree generation and check to build/upgrade

Itxaka:
Adds the luet-mtree package to cos in order to verify the upgrade images


davidcassany (commented):
Looks good to me, nice!

davidcassany (approved):

.github/workflows/build.yaml
```diff
@@ -26,162 +26,169 @@ jobs:
...
-      with:
-        name: sign-${{ matrix.flavor }}
-        path: /tmp/sign_images.log
-        if-no-files-found: warn
+      - name: Create repo
+        run: |
+          sudo -E make create-repo
+      - name: Generate manifests
+        run: |
+          for f in build/*tar*; do
```
mudler, line 131: Can't we make this part of the same extension? I guess it should be trivial to make it also work as an internal plugin. In this way if someone else has to embed this in other kinda of CIs it's enough to pass by something like `--plugin luet-mtree` during publish/create

mudler (approved):
Looks good!

.github/workflows/build.yaml
Itxaka: I havent checked how luet works with plugins, but I'll have a look!

> Can't we make this part of the same extension? 

You mean, iterate over a dir/glob and generate manifests for all the files found?
Should be possible, but then we either autogenerate the output file based on the input name, or lose the possibility of the stdout output, unless I rework it to be under a flag.

I think it would be a good addition, yes. But I think I rather concentrate on the manifest cehck and rework this at a later stage :D

PR #179, opened 2021-05-12
Append the mtree checksums to metadata

Itxaka:
During build time after generating the mtree checksums we can easily
integrate those in the metadata so they are automatically pushed and
downloaded on upgrade

### DIFF
--- a/.github/append_manifests.py
+++ b/.github/append_manifests.py
@@ -1,0 +1,25 @@
+#!/usr/bin/env python
+
+import yaml
+import sys
+
+
+def main():
+    if len(sys.argv) != 4:
+        print("This utility gets a yaml file, a content file and a key and inserts the content from the content"
+              "file into the yaml file at the given key")
+        print("Missing arguments: [YAML FILE] [CONTENT FILE] [KEY TO ADD THE CONTENT AT]")
+        exit(1)
+
+    with open(sys.argv[1], "r") as stream:
+        yaml_file = yaml.safe_load(stream)
+
+    with open(sys.argv[2], "r") as mtree:
+        yaml_file[sys.argv[3]] = mtree.readlines()
+
+    with open(sys.argv[1], "w") as outfile:
+        outfile.write(yaml.dump(yaml_file))
+
+
+if __name__ == "__main__":
+    main()
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -122,15 +122,20 @@ jobs:
         run: |
           sudo -E make build
           ls -liah $PWD/build
-
-      - name: Create repo
-        run: |
-          sudo -E make create-repo
       - name: Generate manifests
         run: |
           for f in build/*tar*; do
             sudo -E luet mtree -- generate $f -o "$f.mtree"
           done
+      - name: Append manifests to metadata
+        run: |
+          for f in build/*mtree; do
+            BASE_NAME=`basename -s .package.tar.zst.mtree $f`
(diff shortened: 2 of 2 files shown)


mudler (approved):

.github/append_manifests.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python
```
mudler, line 1: can maybe replaced by `yq` here? I'm a bit.... :D  reluctant to add python as a dependency

Itxaka: see my comment below :D

mudler:
> I had to create an extra python script to do due to limitations on passing teh content to `yq` in the command line. Unfortunately, using `yq` resulted in errors because the shell could not bear passing the full checksum for the os-system, as its way too big, hence the necessity of having to do it with python.
> 
> I chose python because requires no compilation, its straightforward to do, can live along the code in the repo and all the needed reqs are already installed in the gihub runners.

ok I see :+1:  at the end if we will squash this into a golang plugin we won't need it around for long so I can live with it :D

Itxaka:
> ok I see +1 at the end if we will squash this into a golang plugin we won't need it around for long so I can live with it !

>Update: a later thought, sometimes due to yq limitations I found myself to pipe and use jq with it. e.g. YAML -> JSON -> manipulation with jq -> yq -> YAML and jq allows to pass files to slurp data.
I know it sounds weird, but it has practical usage and workarounds lots of limitations, e.g. Luet-lab/extensions@4c59609/extensions/autobump-github/luet-autobump-github#L85. On the topic, I know that we eventually we will drop both of them in favor of a single CLI, but I lately checked out dasel and seems to perform very nice too.

This is mainly a shell limitation as the os-system checksum is way too big to use as a shell variable, and unless yq/jq support inserting into a key X from file Y (which is what my script does) we wont be able to fix this in the command line.

I tested several things but was unable to do it in any way directly with the installed tools :/

mudler:
> > ok I see +1 at the end if we will squash this into a golang plugin we won't need it around for long so I can live with it !
> 
> > Update: a later thought, sometimes due to yq limitations I found myself to pipe and use jq with it. e.g. YAML -> JSON -> manipulation with jq -> yq -> YAML and jq allows to pass files to slurp data.
> > I know it sounds weird, but it has practical usage and workarounds lots of limitations, e.g. [Luet-lab/extensions@4c59609](https://github.com/Luet-lab/extensions/commit/4c59609)/extensions/autobump-github/luet-autobump-github#L85. On the topic, I know that we eventually we will drop both of them in favor of a single CLI, but I lately checked out dasel and seems to perform very nice too.
> 
> This is mainly a shell limitation as the os-system checksum is way too big to use as a shell variable, and unless yq/jq support inserting into a key X from file Y (which is what my script does) we wont be able to fix this in the command line.
> 
> I tested several things but was unable to do it in any way directly with the installed tools :/

With jq you can https://stackoverflow.com/questions/36273218/jq-read-txt-file-and-write-the-values-to-json-file

But I'm fine with the python script as well :+1:

davidcassany:
> ok I see +1 at the end if we will squash this into a golang plugin we won't need it around for long so I can live with it !

I fully agree on that, we know this is a mid step to start rolling. But at the end we need some better way to do all that, it all should all just be part of the `system/cos` build procedure IMHO, so it is transparent to the CI.

davidcassany (approved):

Itxaka:
> With jq you can stackoverflow.com/questions/36273218/jq-read-txt-file-and-write-the-values-to-json-file

Uh... interesting...

PR #184, opened 2021-05-14
Embed grub configuration into the image

### LINKED ISSUE #109: Embed grub configuration into published images
Would be really good if the kernel command line is somehow embedded into the images, this way each image could include it's own parameters and would allow us to upgrade kernel parameters as needed without having to worry about being backward compatible.

Probably we could make that each menu item in the grub menu causes a reload from the appropriate partition to the specific grub.cfg file devoted to that option in the menu

Probably something like 

```
menuentry "cOS" --id cos {
  search.fs_label COS_STATE root
  set img=/cOS/active.img
  loopback loop0 /$img
  set root=($root)
  source (loop0)/boot/grub.cfg
}
```
where the `/boot/grub.cfg` file inside the `active.img` includes something like

```
  linux (loop0)/boot/vmlinuz console=tty1 ro root=LABEL=COS_ACTIVE iso-scan/filename=/cOS/active.img panic=5
  initrd (loop0)/boot/initrd
```

This would smooth solving #65 and also facilitate the introduction of kernel parameters through upgrades which can be handy for in cases like #135.

davidcassany:
This commit embedds the grub configuration into the image, in fact, the
embedded part affects only the boot arguments which are provided within
the image.

The main grub.cfg is still static but provided as an static file within
the image too.

This commit allows us to change boot parameters on upgrades and also
easily allows the user to include additional parameters at run time if
needed.

Fixes #109


Itxaka (approved):
Looks good!

packages/grub-config/build.yaml
```diff
@@ -0,0 +1,8 @@
+requires:
+- name: "base"
+  category: "distro"
+  version: ">=0"
+
+steps:
+- mkdir /etc/cos
+- cp config/grub.cfg config/bootargs.cfg /etc/cos
```
Itxaka, line 8: do we need to also set the permissions/owner/group here?

davidcassany: Good question, I do not have any strong opinion. I don't think it hurts having it readable and this is part of the immutable region, thus there aren't many chances to modify it...

davidcassany:
Note this PR modifies the grub configuration in a non backward compatible way. Meaning that it will break upgrade tests and from this point in time it won't be possible to rollback further than this point, which I don't think it is a problem for this early stage.

packages/cos/setup.yaml
```diff
@@ -25,49 +25,3 @@ stages:
...
         VERSION: ":VERSION:"
         PRETTY_NAME: ":PRETTY_NAME:"
-    - name: "Grub"
-      directories:
-      - path: /etc/cos
-        permissions: 0755
-        owner: 0
-        group: 0
-      files:
-      - path: /etc/cos/grub.cfg
```
mudler, line 35: Note: this is being used here currently when refreshing grub during `cos-reset`. https://github.com/rancher-sandbox/cOS-toolkit/blob/master/packages/installer/reset.sh#L71

mudler: do we have to carry the same for bootargs?

mudler: I think I've misread it, it should be ok as the grub package ships in /etc/cos/grub.cfg at the end

mudler (approved):

packages/cos/setup.yaml
davidcassany: No bootargs are not _installed_ they are just directly read from the loop device at runtime from `/etc/cos/bootargs.cfg`. See https://github.com/rancher-sandbox/cOS-toolkit/blob/481d882cf9f964dcd7113695026e14ee25cdb505/packages/grub-config/config/grub.cfg#L13-L17

davidcassany: So each image is free to provide its own bootargs config and obligated to provide a valid ones, otherwise it won't boot.

tests/recovery/recovery_test.go
```diff
@@ -24,6 +24,9 @@ var _ = Describe("cOS Recovery upgrade tests", func() {
 
 				version := out
 				By("upgrading to raccos/releases-opensuse:cos-system-0.5.0")
+				By("turning off verification, and pointing to an old release branch")
+				out, err = s.Command("sed -i 's/verify: true/verify: false/g' /etc/luet/luet.yaml")
```
Itxaka, line 28: can we make this a helper function so its clearer and reusable?

davidcassany: I'd say the approach should be to get rid of all this as soon as we stop using DCT and start using other validation strategies.

davidcassany: While I see the point on doing it as helper function I'd say first finalize the DCT and validation story and then adapt tests to make them cleaner. They will need some adaptations at that point too.

Itxaka (approved):

PR #185, opened 2021-05-14
Add plugin to check the mtree signatures on upgrade

Itxaka:


packages/installer/image-mtree-check.sh
```diff
@@ -0,0 +1,31 @@
...
+
+  image="$repo":"$name"-"$category"-"$download_version".metadata.yaml
+  tmpdir=/tmp/"$name"-"$category"-"$download_version"-metadata
+  mtree_output=/tmp/"$name"-"$category"-"$download_version".mtree
+
+  echo "Getting $image metadata"
+
+  luet util unpack "$image" "$tmpdir"
+  # This is terrible, if this works we need to move to a golang binary FAST
+  curl https://github.com/mikefarah/yq/releases/download/v3.4.1/yq_linux_amd64 -L --silent --output /tmp/yq && chmod +x /tmp/yq
```
mudler, line 24: Maybe we can add yq in the toolchain dir for the moment as we do for the other components, wdyt?

Itxaka: Im struggling with that, like Its the way to go, but adding a package to the "base" to then remove it...I dont really like it. But I think its the way to go until we move to the single binary/plugin/whatever

mudler: yep.. at the moment we have no much choice, but it's fine for the moment :) we have issues to track this :t-rex:

Itxaka:
bloacked by https://github.com/rancher-sandbox/cOS-toolkit/pull/190

Itxaka:
Should be ready to review!

packages/cos/build.yaml
```diff
@@ -22,6 +22,9 @@ requires:
 - name: "selinux-policies"
   category: "system"
   version: ">=0"
+- name: "yq"
```
mudler, line 25: This should go into the `copy` part, otherwise it will bring `golang` inside `system/cos` (the installable artifact)

Itxaka: good point! probably an artifact of my rebase!

Itxaka: ah no, this was added by myself

packages/installer/upgrade.sh
```diff
@@ -124,7 +124,7 @@ upgrade() {
     export HOME=/tmp # Docker Content Trust data is stored in $HOME/.docker. We don't need those to persist
 
     if [ -n "$CHANNEL_UPGRADES" ] && [ "$CHANNEL_UPGRADES" == true ]; then
-        luet install --system-target /tmp/upgrade --system-engine memory -y $UPGRADE_IMAGE
+        luet install --plugin image-mtree-check --system-target /tmp/upgrade --system-engine memory -y $UPGRADE_IMAGE
```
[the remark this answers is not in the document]
Itxaka: if the user provides a single image, what would change here? I havent explored that yet so Im not sure of that workflow but as long as the image is a tar or extracted somewhere, the workflow should be the same, get metadata, extract mtree values from metadata, run luet-mtree binary with the contents and the signatures

Even if the image is not a tar, we can always expand on that luet-mtree/unified CLI binary to open whatever we want I guess?

Itxaka:
upgrade tests passed, so that means the check passed correctly :)

Still, gonna add a test suite for this, I want to be sure not to brake any updates workflow

packages/installer/upgrade.sh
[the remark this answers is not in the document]
mudler: yup sounds right, in that case It would hit later on L134, which would require I manual download of the metadata (if present)

packages/installer/upgrade.sh
```diff
@@ -124,7 +124,7 @@ upgrade() {
     export HOME=/tmp # Docker Content Trust data is stored in $HOME/.docker. We don't need those to persist
 
     if [ -n "$CHANNEL_UPGRADES" ] && [ "$CHANNEL_UPGRADES" == true ]; then
-        luet install --system-target /tmp/upgrade --system-engine memory -y $UPGRADE_IMAGE
```
mudler, line 127: There is no way to opt-out from verification right? I think we should respect `--verify` here and disable/enable the plugin loading on request

Itxaka: yep, good pointer! I was wondering how to add tests to verify this and just stumble upon this, there is currently no switch to enable/disable this and that is bad. Let me update this with the verify switch on place for this!

Itxaka: done!

packages/installer/upgrade.sh
```diff
@@ -218,8 +221,8 @@ while [ "$#" -gt 0 ]; do
         --recovery)
             UPGRADE_RECOVERY=true
             ;;
-        --verify)
-            VERIFY=true
+        --no-verify)
```
mudler, line 224: We should also update docs accordingly as we are changing the default here.. more importantly for derivatives that are not signing images, should set a default `VERIFY=false` in the upgrade channel configuration file, e.g. here https://github.com/rancher-sandbox/cos-toolkit-sample-repo/blob/master/packages/sampleOS/02_upgrades.yaml#L48

Itxaka: oh good point, its the second time I change this around so docs need updating.

Will send also a patch for the derivatives

Itxaka: done!

Tests still need rework thought, that can come afterwards in a different PR to rework the whole upgrade testing part.

mudler (approved):

PR #186, opened 2021-05-14
Drop docker content trust

Itxaka:
Is not necessary now that we check the sha of the image contents and it
ties us to dockerhub.


mudler (changes requested):
It looks good, but there are pieces missing in order to disable it by default:

- https://github.com/rancher-sandbox/cOS-toolkit/blob/7c228bb2bb28f1c2682e9ccddeb6a3d6c03eb6b0/packages/cloud-config/oem/02_upgrades.yaml#L35 -> needs to be dropped
- This logic https://github.com/rancher-sandbox/cOS-toolkit/blob/7c228bb2bb28f1c2682e9ccddeb6a3d6c03eb6b0/packages/installer/upgrade.sh#L130 needs to be inverted
- Docs updates :)

Itxaka:
woops, sorry, this was mean to have a WIP tag as I sent it just before finishing on monday :)

mudler (approved):

Itxaka:
Need to update to reflect changes made on 0.5.0 as the grub changes are not backwards compatible :/

tests/upgrades-images-signed/upgrade_test.go
```diff
@@ -19,24 +19,48 @@ var _ = Describe("cOS Upgrade tests - Images signed", func() {
...
 				out, err := s.Command("source /etc/os-release && echo $VERSION")
 				Expect(err).ToNot(HaveOccurred())
 				Expect(out).ToNot(Equal(""))
 
 				version := out
-				By("upgrading to an old signed image")
-				out, err = s.Command("cos-upgrade --docker-image raccos/releases-opensuse:cos-system-0.5.0")
+
+				By("running cos-upgrade with --verify and an signed image")
+				out, err = s.Command("cos-upgrade --verify --docker-image raccos/releases-opensuse:cos-system-0.4.9-9")
```
davidcassany, line 63: This test might not work, however this is not executed currently on the CI. So if tests pass, I would not mind about this, since there is #197 which is meant to solve these issues. It will popup there and conflicts can be easily solved there.

Itxaka: argh, we gonna have a conflict here because as the verify args are changing around I moved some tests from signed <-> unsigned as they make more sense there

tests/upgrades-images-unsigned/upgrade_test.go
```diff
@@ -19,22 +20,22 @@ var _ = Describe("cOS Upgrade tests - Images unsigned", func() {
 	})
 	Context("After install", func() {
 		When("images are not signed", func() {
-			It("fails to upgrade to a version which is not signed", func() {
-				out, err := s.Command("cos-upgrade --docker-image raccos/releases-opensuse:cos-system-0.5.0")
-				Expect(err).To(HaveOccurred())
-				Expect(out).Should(ContainSubstring("No valid trust data"))
+			It("upgrades to latest available (master) and reset", func() {
```
davidcassany, line 23: I guess this should try to upgrade to 0.5.0, but if it tries to pull an earlier one it will fail...

Itxaka: as far as I understand, master should be on 0.5.0 already and it will try to update to whatever is latest, which should always be >= 0.5.0 as you pushed that image no?

Im more worried about the other test below as I'll have to remove it, there is no point into changing into a unsigned release branch anymore as the main branch (release-opensuse) is already unsigned....

PR #190, opened 2021-05-17
condensed output for mtree checksums

Itxaka:
Read all into one line so its easier to manipulate afterwards and also
saves space when storing it as a big blob instead of one item per line

Required for #185 


davidcassany (approved):
LGTM

mudler (approved):

PR #194, opened 2021-05-18
Bump luet-mtree to 0.0.3

Itxaka:
Now it ignores empty files, uses less keywords by default allowing more
control on what keywords to use

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -111,7 +111,7 @@ jobs:
       - name: Install deps
         run: |
           sudo -E make deps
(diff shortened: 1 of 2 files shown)


mudler (approved):

.github/workflows/build.yaml
```diff
@@ -111,7 +111,7 @@ jobs:
       - name: Install deps
         run: |
           sudo -E make deps
```
mudler, line 113: thinking out loud,  :thinking: ... we could actually add the `cos` repository itself in `make deps` and consume the same binary in the tree

PR #197, opened 2021-05-20
Re-enable tests using published signed images

davidcassany:
This commit re-adds to the CI tests that make use of published signed
images. It also adds few additional checks on tests to ensure reboots
are booting to the expected system and no fallback boot happened.

Upgrade tests have been slightly simplify to avoid some reboots and
resets while keeping the fundamental functionality tests.

Finally, it also bumps a new system/cos release to it can be used in
tests.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -254,9 +254,7 @@ jobs:
     strategy:
       matrix:
         flavor: ["opensuse"]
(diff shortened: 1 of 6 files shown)


davidcassany:
This PR requires https://github.com/rancher-sandbox/cOS-toolkit/actions/runs/859713141 to be green or any other more recent successful `master` build. Otherwise the images used in tests won't be there.

mudler (approved):

PR #198, opened 2021-05-20
Fix output for sign images

Itxaka:
Fix output when dealing with a different event than image.post.build
Append to the output log
Add start/end lines to the log


davidcassany (approved):
Looks good :+1:

mudler (approved):

PR #199, opened 2021-05-20
system/cos cleanups

mudler:
Since luet `0.15.0` is released it's available the [copy](https://luet-lab.github.io/docs/docs/concepts/packages/specfile/#copy) keyword, in this way it's easy to disjoin packages that doesn't need to share a common base, and moreover we can drop some of the workaround that we had in the specfiles

### DIFF
--- a/packages/cos/build.yaml
+++ b/packages/cos/build.yaml
@@ -1,20 +1,11 @@
 # Refer to https://luet-lab.github.io/docs/docs/concepts/packages/specfile/#build-specs
 # for the syntax format
 requires:
-- name: "luet"
-  category: "toolchain"
-  version: ">=0"
-- name: "luet-mtree"
-  category: "toolchain"
-  version: ">=0"
 - name: "cos-setup"
   category: "system"
   version: ">=0"
 - name: "installer"
   category: "utils"
-  version: ">=0"
-- category: "toolchain"
-  name: "yip"
   version: ">=0"
 - name: "cloud-config"
   category: "system"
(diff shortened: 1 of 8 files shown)


packages/cos/build.yaml
```diff
@@ -32,6 +23,27 @@ requires:
...
   version: ">=0"
 
+# https://luet-lab.github.io/docs/docs/concepts/packages/specfile/#copy
+copy:
+- package: 
+    category: "toolchain"
+    name: "yip"
+    version: ">=0"
+  source: "/usr/bin/yip"
+  destination: "/usr/bin/yip"
```
davidcassany, line 33: Out of curiosity,  are `source` and `destination` mandatory? I would expect that without them the `copy` procedure just untars the package contents in `/`. Is that the case? And if so, why aren't we using it here? `yip` package only includes `/usr/bin/yip` after all.

davidcassany (approved):
Looks good to me. I am just surprised by the  `source`and `destination` elements of the copied package in build.yaml file. I would not have expected those. In any case this kind of unrelated to this PR, as this is tied to the luet copy feature.

packages/cos/build.yaml
mudler: `copy` is meant to be a low-level function, it indeed just transliterates to `COPY ...` internally, but it makes sure there is a build image for the underlying package. 

For package contents we have a separate issue https://github.com/mudler/luet/issues/173, but the problem there is about mix and match: that is to say, at that point, or you use `requires` or the final images

mudler: ok - turns out that since `copy` is there already it's really straightforward.. so I've tried it and works, I'll push it forward in the next release

mudler: `0.16.0` should include support for `join`, https://github.com/mudler/luet/releases/tag/0.16.0

Issue #144, opened 2021-05-04
ci: Add a test for the sample repo


mudler:
At the moment, we are not testing if our changes are breaking the workflow in https://github.com/rancher-sandbox/cos-toolkit-sample-repo which is our quickstart for newcomers.

**Describe the solution you would like**
A step in the CI workflow which builds the sample repo as advertized in the README against the changes in the patchset

Sub-Issues/PR:
- https://github.com/rancher-sandbox/cOS-toolkit/pull/205
- #274

Itxaka:
I had a quick look at this, and it seems that we can get this to be automated by using workflow_dispatch and a personal access token.

Basically making a build workflow on the sample repo that runs `on: workflow_dispatch` and then at the end of a cos-toolkit master test run we fire that event using `https://github.com/marketplace/actions/workflow-dispatch` like this:

```
      - name: Invoke CI in cos-toolkit-sample-repo
        uses: benc-uk/workflow-dispatch@v1
        with:
          workflow: NAME OF THE REMOTE WORKFLOW
          repo: rancher-sandbox/cos-toolkit-sample-repo 
          token: ${{ secrets.PERSONAL_TOKEN }}
```

It should trigger the mentioned workflow there. This way we could have not just one for the sample repo, but for other test appliances that we have.

Of course...we would need to check the CI of those sample repos manually to check if we broke them...but we could always automate that in the sample repos to post to slack or whatever I guess

mudler:
This keeps going to be unnoticed. I've moved to Untriaged.

 See https://github.com/rancher-sandbox/cOS-toolkit/issues/274 and how scheduled jobs in our sample repo were failing too https://github.com/rancher-sandbox/cos-toolkit-sample-repo/runs/2823450942?check_suite_focus=true

mudler:
Closing, sample repos were archived. We have docs for creating derivatives which does not involve the direct usage of luet, and that is our supported way of creating derivatives.

PR #204, opened 2021-05-24
Switch to Github Container Registry

### LINKED ISSUE #118: Move away from Dockerhub in CI to prevent rate limit failures
Too often the CI fails due to rate limit issues from Dockerhub.

We should move to a different registry without such problems.

mudler:
This PR switch from DockerHub to GHCR.

It also drops `yq` from our dependencies in favor of adding to the build command ```--all```. That is also required for derivative that wish to build from the repositories, the values data are interpolated in the metadata.yaml of each package, that means that if we don't build it downstream repositories can't access to it too.

Fixes #118


mudler:
@Itxaka @davidcassany this is high prio and If pass tests I have to merge it as is, otherwise the sample repos (and all our docs) are invalidated :/

Context:
see the sample repo failures in https://github.com/rancher-sandbox/cos-toolkit-sample-repo/actions

We should really tackle #144, it gets really unnoticed..

Itxaka:
I thought that gh registry was much more conservative in terms of artifacts bandwidth and so on? Or because we are on a business plan that is not the case?

Itxaka (approved):

mudler:
Ok, changed the PR to switch to Quay instead, @Itxaka can you have a look again please?

README.md
```diff
@@ -23,22 +23,14 @@ It is designed to reduce the maintenance surface, with a flexible approach to pr
 
 cOS derivatives are built from containers, and completely hosted on image registries. The build process results in a single container image used to deliver regular upgrades in OTA approach. Each derivative built with `cos-toolkit` inherits by default the [following featuresets](/docs/derivatives_featureset.md).
 
-cOS supports different release channels, all the final and cache images used are tagged and pushed regularly [to DockerHub](https://hub.docker.com/r/raccos/releases-amd64/) and can be pulled for inspection from the registry as well.
+cOS supports different release channels, all the final and cache images used are tagged and pushed regularly [to Github Container Registry](https://quay.io/repository/costoolkit/releases-opensuse) and can be pulled for inspection from the registry as well.
```
Itxaka, line 26:
```suggestion
cOS supports different release channels, all the final and cache images used are tagged and pushed regularly [to Quay Container Registry](https://quay.io/repository/costoolkit/releases-opensuse) and can be pulled for inspection from the registry as well.
```

README.md
```diff
@@ -23,22 +23,14 @@ It is designed to reduce the maintenance surface, with a flexible approach to pr
...
-$ docker run -ti --rm raccos/releases-opensuse:cos-system-0.4.30 /bin/bash
-```
-
-cOS Images are signed, and during upgrades Docker Content Trust is enabled.
-
-You can inspect the images signatures for each version:
-
-```bash
-$ docker trust inspect raccos/releases-opensuse:cos-system-0.4.32
+$ docker run -ti --rm quay.io/costoolkit/releases-opensuse:cos-system-... /bin/bash
```
Itxaka, line 33: More explicit that you need to set a version here?

```suggestion
$ docker run -ti --rm quay.io/costoolkit/releases-opensuse:cos-system-$VERSION /bin/bash
```

Itxaka (changes requested):
readme small issue, sorry :/

README.md
mudler: gah! :facepalm:  I thought I catched up all the occurences.. thanks!

Itxaka (approved):

Issue #209, opened 2021-05-25
ci: docker-build test fails for unavailable space


mudler:
Seems we run out of space in GH workers when building from the docker image

**cos-toolkit version:**
N/A

**CPU architecture, OS, and Version:**
N/A

**Describe the bug**
``` 📦  build/golang-1.16.4+3 🐋  Generating 'package' image from raccos/fedora:builder-b3dec7ea9a4bb0531b15ad057fa45532 as raccos/fedora:41e31bc8233e5547b495f66497770089de58af8579f65cc58917883eaba8a3dc with build steps
 🐋  Downloaded image: raccos/fedora:41e31bc8233e5547b495f66497770089de58af8579f65cc58917883eaba8a3dc
 📦  build/golang-1.16.4+3 🔨  Generating delta
 Error: while resolving multi-stage images: failed building multi-stage image: Failed compiling build/golang-1.16.4+3: Error met while generating delta: Could not generate changes from layers: Error met while unpacking dst image raccos/fedora:41e31bc8233e5547b495f66497770089de58af8579f65cc58917883eaba8a3dc: failed while extracting rootfs for raccos/fedora:41e31bc8233e5547b495f66497770089de58af8579f65cc58917883eaba8a3dc: Failed exporting image: write /var/tmp/luet/extraction907140095/dst930592905/tmprootfs313814694/.docker_temp_510053849: no space left on device
```

**To Reproduce**

**Expected behavior**
Successful build

**Logs**
https://github.com/rancher-sandbox/cOS-toolkit/runs/2659492726

**Additional context**

mudler:
That's not anymore sporadically. It happens all the time now

PR #216, opened 2021-05-26
make: set back --only-target-package as BUILD_ARGS default

mudler:
Now that CI has been split into different BUILD_ARGS there is no need
for keeping this default. With --only-target-package we save a lot of
disk space if we don't need metadata for all of the packages built.

Fixes: #209


Itxaka (approved):
nice!

Issue #113, opened 2021-04-22
Add checksum verification for assets downloaded externally


mudler:
We need to add checksum verification for the assets downloaded during packagebuild, coming from outside.

Those includes:
- livecd/systemd-boot
- livecd/syslinux
- build/golang

~~Our automated [CI bump-bot](https://github.com/Luet-lab/extensions/tree/master/extensions/autobump-github) should update the checksum accordingly when opening new PR.~~ This has been done separately, see https://github.com/rancher-sandbox/cOS-toolkit/pull/229/files

The only two other components are `yip` and `luet` but with #108 won't be needed anymore.

## Requirements

 - [x] External assets have a checksum 
   - [x] https://github.com/rancher-sandbox/cOS-toolkit/pull/229 
   - [x] https://github.com/rancher-sandbox/cOS-toolkit/pull/230
 - [x] That checksum is checked during package build

mudler:
This is part of #95 which is required for MVP

mudler:
I'll split the CI autobump in a separate story

Itxaka:
All PRs merged!

PR #229, opened 2021-05-31
Add golang checksum hook

mudler:
The support has been added with: https://github.com/Luet-lab/extensions/commit/39b72f0a2b3fc188458e10483e4d9d6b7f46a529 (devkit 0.10.2). 
Specs with a `autobump.checksum_hook` field will be executed as a script to fetch the checksum which is then annotated in `labels.package.checksum`. It supports the templating syntax as usual, allowing to dynamically fetch checksums by package versions. 

The checksum should be accessible with: `{{ ( index .Values.labels "package.checksum" ) }}` from the `build.yaml` file

To test it locally, run `luet autobump-github` (with the devkit @ version 0.10.2)

### DIFF
--- a/packages/golang/definition.yaml
+++ b/packages/golang/definition.yaml
@@ -12,4 +12,6 @@ labels:
   autobump.version_hook: |
     curl -s -L 'https://golang.org/VERSION?m=text' | sed 's/go//g'
   package.version: "1.16.4"
+  autobump.checksum_hook: "curl -q -L https://storage.googleapis.com/golang/go{{.Values.labels.package.version}}.linux-{{.Values.arch}}.tar.gz.sha256"
+  package.checksum: "7154e88f5a8047aad4b80ebace58a059e36e7e2e4eb3b383127a28c711b4ff59"
 arch: "amd64"


Itxaka (approved):

Itxaka:
Although I find it quite difficult to understand how are we gonna check this later. We have no type of checksum so we cannot infer what type is it. sha256? 512? md5?

mudler:
I think that's fine to be static in this... If the `build.yaml` and `definition.yaml` of a package agree that they are checking on sha256, that can even be a separate key in the `definition.yaml` file

mudler:
For completeness, then from the build.yaml file, the checksum should be accessible with: `{{ ( index .Values.labels "package.checksum" ) }}`

Itxaka:
> I think that's fine to be static in this... If the build.yaml and definition.yaml of a package agree that they are checking on sha256, that can even be a separate key in the definition.yaml file

Yeah I was thinking on `package.checksum_type` so different stuff can be run in a more generic way based of that value. Does luet templates supports defining helper templates? That way a shared function could work across all packages based of that checksum_type.

In any case, not needed now.

mudler:
> > I think that's fine to be static in this... If the build.yaml and definition.yaml of a package agree that they are checking on sha256, that can even be a separate key in the definition.yaml file
> 
> Yeah I was thinking on `package.checksum_type` so different stuff can be run in a more generic way based of that value. Does luet templates supports defining helper templates?

Not yet, that would be a nice addition :)

> That way a shared function could work across all packages based of that checksum_type.

Yup, that would be great indeed! I'll open an issue to track the templating functions :+1:

davidcassany (approved):
Looks good to me, even I am not sure I understand the motivation of this change. Is it just to add some extra verification that the source binaries are not changeable without raising an alarm?

mudler:
> Looks good to me, even I am not sure I understand the motivation of this change. Is it just to add some extra verification that the source binaries are not changeable without raising an alarm?

yup it is related to #113 it would make sure that wherever we are building the package the source should always be the same.  At the moment, for e.g. let's say that you are building the `golang` spec from scratch could happen that if you get a 50x back, we will just try out to uncompress a 500 HTML page with tar. 

The mechanism above is to make it dynamic so our CI bot can also update the checksum automatically during a version bump ( new golang is out -> updates both version and checksum, so `build.yaml` can just consume those)

PR #232, opened 2021-06-01
Check external assets checksums

Itxaka:
Checks the external assets checksums during build

Partial: https://github.com/rancher-sandbox/cOS-toolkit/issues/113

### DIFF
--- a/packages/livecd/syslinux/definition.yaml
+++ b/packages/livecd/syslinux/definition.yaml
@@ -2,4 +2,4 @@ category: "live"
 name: "syslinux"
 version: "6.03"
 labels:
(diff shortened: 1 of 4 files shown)


packages/livecd/syslinux/definition.yaml
```diff
@@ -2,4 +2,4 @@ category: "live"
 name: "syslinux"
 version: "6.03"
 labels:
-  package.checksum: "250b9bd90945d361596a7a69943d0bdc5fc0c0917aa562609f8d3058a2c36b3a"
+  package.checksum: "26d3986d2bea109d5dc0e4f8c4822a459276cf021125e8c9f23c3cca5d8c850e"
```
Itxaka, line 5: checksum changed because the old one is for the `tar.gz` package and we download the `tar.xz` which is half the size

mudler (approved):

PR #234, opened 2021-06-02
[WIP] Add image verification for single images

Itxaka:
When upgrading by passing a single docker image we need to also
verifythat the contents from the image have the same checksum as
the one stored in the metadata

This expands the existing image-mtree-check script to take into
consideration the unpack event and act accordingly


Itxaka:
sent because I want the tests to run :D

Itxaka:
Ready for review!

davidcassany (dismissed):
I am not following all the details, but it looks like a clean and simple refactor. All good to me :+1:

packages/installer/upgrade.sh
```diff
@@ -137,6 +137,13 @@ upgrade() {
         luet util unpack $args $UPGRADE_IMAGE /usr/local/tmp/rootfs
         rsync -aqzAX --exclude='mnt' --exclude='proc' --exclude='sys' --exclude='dev' --exclude='tmp' /usr/local/tmp/rootfs/ /tmp/upgrade
         rm -rf /usr/local/tmp/rootfs
+        # Call image-mtree-check directly with the data until luet supports sending events on unpack
```
mudler, line 140: new 0.16.5 should have this :+1:

mudler:
It looks good, but probably we should add back tests that reflect this feature (we were previously testing DCT) https://github.com/rancher-sandbox/cOS-toolkit/blob/master/tests/upgrades-images-signed/upgrade_test.go

packages/installer/upgrade.sh
Itxaka: already released? wow that was fast!

Itxaka:
> It looks good, but probably we should add back tests that reflect this feature (we were previously testing DCT) https://github.com/rancher-sandbox/cOS-toolkit/blob/master/tests/upgrades-images-signed/upgrade_test.go

Yes, that was the next step. First I wanted to make sure I didn't break everything with the script changes :+1:

mudler (approved):

Itxaka:
Adding wip back, seems that tests are not running as expected but Im blind (having problems running them locally) so I need to investigate further!

tests/upgrades-images-signed/upgrade_test.go
```diff
@@ -15,15 +15,44 @@ var _ = Describe("cOS Upgrade tests - Images signed", func() {
...
+		// Try to gather mtree logs
+		s.GatherLog("/tmp/image-mtree-check.log")
+		s.GatherLog("/tmp/luet_mtree_failures.log")
+		s.GatherLog("/tmp/luet_mtree.log")
 		s.Reset()
 	})
 	Context("After install", func() {
 		When("upgrading", func() {
-			It("fails if verify is enabled on an unsigned version", func() {
-				// Using releases-amd64 as those images are not signed
```
mudler, line 23: I wonder, in this way we aren't checking anymore if we do _fail_ instead when an image is unsigned, but we have the verification turned on?

Itxaka: oh good point!

mudler: I think for sake of testing here, we could even try to upgrade to a `alpine` image, as it will try to pull the metadata file from it, it should fail. That would just test that we are trying to fetch the meta, but not the upgrade itself.

Indeed it's tricky - one way to do this is to push a manual image with a metadata file which have some entries corrupted, and hardcoding that image in the tests

Itxaka: Actually I think 0.5.0 is not signed, or at least not properly, so we can use that to test for failure!

Itxaka: done,test updated!

mudler (approved):

Issue #192, opened 2021-05-17
Split recovery image and system image into separate artifacts


davidcassany:
Currently we are using the same image for the recovery partition than for the active (`COS_ACTIVE`) loop device. For a plain system like the one we are using now it might be reasonable, however for a derivative service specific appliance this approach does not have much sense.

I'd say that generically speaking having production services into the recovery partition is a bad idea (even a dangerous idea) and we should facilitate having some control about what is the recovery system and what is the production system. 

For that I would simply create different images, `system/cos` and `recovery/cos`. `system/cos` used for the `COS_ACTIVE` and `COS_PASSIVE ` loop devices and the  `recovery/cos` one used for the COS_SYSTEM partition. Also the immutability approach could be expressed with different defaults for the recovery system than for the COS_ACTIVE system.

In addition to that and linked with #217 I believe having a recovery image contained in a squashfs filesystem is better for few reasons:

* Immutability as it was well known live image, no persistence at all and whole system gets overlayed.
* No need of applying any sort of custom layout code at boot (just use the well established and robust live boot procedure)
* Compressed images so less space in disk, I'd say the performance penalty for a recovery system is just fine.
* It can be populated without having to mount any filesystem (relevant for containerized builds)

## Action items

- [x] Split `system/cos` into `recovery/cos` and `system/cos`, so that during upgrades it can be specified a different image to upgrade recovery from
- [x] ~~Make the recovery script accept an image to use to recovery from, besides the current behavior which is replicating the recovery to the ACTIVE and PASSIVE~~ / Make sure `cos-upgrade` works from the recovery
- [x] Move the recovery layout to squashfs instead of an image
- [ ] Fix/Add new tests

mudler:
The move to squashfs is more complex than it looks like: at the moment we are relabeling image files and copying them over instead of recreating them, and more over we should ship a squashfs within the final image, containing another image which is squash'fsed.

mudler:
WIP branch at https://github.com/rancher-sandbox/cOS-toolkit/tree/split_recovery

PR #235, opened 2021-06-03
PoC of building disk images inside non privileged containers

### LINKED ISSUE #218: Test and boot the PoC img in AWS
This card is about validating the viability of using the images built in [image_builder](/rancher-sandbox/cOS-toolkit/tree/image_builder) branch and fine tune it. The goal is get up to a point were we can ssh to the recovery system. Documenting the whole process is also part of this card (uploading the image to AWS and booting it with certain cloud-init config).

Finalize the deployment of by re-partitioning and deploying certain cOS like image is out of the scope of this card.

davidcassany:
This is proof of concept for a simple image builder that can be
executed inside a non privileged container.

Fixes #218 

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -158,11 +158,18 @@ jobs:
         if: github.event_name != 'schedule' && !startsWith(github.ref, 'refs/tags/')
         run: |
           sudo -E make local-iso
+          COS_VERSION=$(yq r packages/cos/definition.yaml 'version')
+          mv *.iso cOS-$COS_VERSION.iso
+          mv *.sha256 cOS-$COS_VERSION.iso.sha256
 
       - name: Build ISO from remote repositories 🔧
         if: github.event_name == 'schedule' || startsWith(github.ref, 'refs/tags/')
         run: |
           sudo -E make iso
+          COS_VERSION=$(yq r packages/cos/definition.yaml 'version')
+          mv *.iso cOS-$COS_VERSION.iso
+          mv *.sha256 cOS-$COS_VERSION.iso.sha256
+
       - uses: actions/upload-artifact@v2
         with:
           name: cOS-${{ matrix.flavor }}.iso.zip
@@ -362,6 +369,39 @@ jobs:
           GHR_PATH: release/
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+raw-img:
+    runs-on: ubuntu-latest
+    container: opensuse/leap:15.3
+    needs: build
+
+    steps:
+      - name: Install deps
+        run: |
+          zypper in -y curl e2fsprogs dosfstools mtools squashfs gptfdisk make grub2-x86_64-efi tar gzip xz
+      - uses: actions/checkout@v2
+      - name: Download result for build
+        uses: actions/download-artifact@v2
+        with:
+          name: build-opensuse
+          path: build
+      - name: Build Image
+        run: |
+          # Since sime time /var/lock is a symlink to /run/lock, which doesn't exit in the continer
+          rm -rf /var/lock
+          mkdir -p /var/lock
+          make deps
+          rm /etc/luet/repos.conf.d/*
+          ./images/img-builder.sh
+          COS_VERSION=$(yq r packages/cos/definition.yaml 'version')
+          mv disk.raw cOS_${COS_VERSION}.raw
+          #qemu-img convert -f raw -O qcow2 disk.raw cOS-$COS_VERSION.qcow2
+
+      - uses: actions/upload-artifact@v2
+        with:
+          name: cOS-opensuse
+          path: |
+            *.raw
+          if-no-files-found: error
 
 # Non-squashfs tests
 
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -158,7 +158,7 @@ jobs:
         if: github.event_name != 'schedule' && !startsWith(github.ref, 'refs/tags/')
         run: |
           sudo -E make local-iso
-          COS_VERSION=$(yq r packages/cos/definition.yaml 'version')
+          COS_VERSION=$(yq r packages/cos/collection.yaml 'version')
           mv *.iso cOS-$COS_VERSION.iso
           mv *.sha256 cOS-$COS_VERSION.iso.sha256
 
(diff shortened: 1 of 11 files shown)


davidcassany:
I tested the image with the following procedure:

1. Upload the image to an S3 bucket
```
aws s3 cp <cos-image> s3://cos-images
```

2. Created the disk container JSON (`container.json` file) as:

```
{
  "Description": "cOS Testing image in RAW format",
  "Format": "raw",
  "UserBucket": {
    "S3Bucket": "cos-images",
    "S3Key": "<cos-image>"
  }
}
```

3. Import the disk as snapshot

```
aws ec2 import-snapshot --description "cOS PoC" --disk-container file://container.json
```

4. Followed the procedure described in [AWS docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/creating-an-ami-ebs.html#creating-launching-ami-from-snapshot) to register an AMI from snapshot. Used all default settings unless for the firmware, were I forced it to UEFI boot.

5. Launch instance with this simple userdata:
```
name: "Default ec2 user"
stages:
   network.after:
     - name: "Setup user"
       users:
         ec2user:
           name: "ec2user"
           passwd: "$6$M5bXBW/.7pspU1n7$d3Un967.AG8yf9YK2qlUFJt/3EQR57Vrlhtil866FglGY9dI2/arcBCcbSk7/faSq8pkwf1dkD.tDX7iXLOuG1"
           primary_group: "users"
           shell: "/bin/bash"
           homedir: "/home/ec2user"
```

TODOS:

* It is not obvious to create a ec2 user on the fly and authorize it with the SSH key provided by amazon, by default only `root` user gets the key.
* Monitoring features are missing, AWS status checks are not integrated yet.

davidcassany:
Note tests for this PR will fail as it roughly changes the grub config to embrace an squashfs recovery image. This PR should be easy to rebase on top of #192 outcome, then it will be transparent to tests.

packages/cloud-config/oem/07_cloud-metadata.yaml
```diff
@@ -0,0 +1,14 @@
+# Default cOS OEM configuration file
+#
+# This file is part of cOS and will get reset during upgrades.
+#
+# Before you change this file manually,
+# consider copying this file to /usr/local/cloud-config or
+# copy the file with a prefix starting by 90, e.g. /oem/91_custom.yaml
+name: "Cloud providers metadata"
```
mudler, line 8: I'm not sure if we want to embed this directly :thinking:  would it behave _weirdly looking_ if booted into a baremetal?

maybe can go into a `feature, or just all commented if the purpose is to have it as a sample

packages/grub-config/config/rec_bootargs.cfg
```diff
@@ -0,0 +1,3 @@
+set kernel=/boot/vmlinuz
```
mudler, line 1: I have already those set of changes in my branch for the squashfs change :+1: I think it's safe to merge this after the squashfs change is done, there are also tests to adjust and scripts (which I'm almost done with) to make everything work with the new format

packages/cloud-config/oem/07_cloud-metadata.yaml
davidcassany: It does nothing if the provider is not found. Note this plugin is designed to be used against all supported cloud providers without having to touch it. It basically tests all listed providers one after the other until the list is over or until there is one provider that actually succeeds.

.github/workflows/build.yaml
```diff
@@ -358,3 +358,37 @@ jobs:
         env:
           GHR_PATH: release/
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  raw-img:
```
Itxaka, line 362: can we have a Makefile target to build this locally?

images/img-builder.sh
```diff
@@ -0,0 +1,79 @@
...
+# Create root-tree for COS_RECOVERY
+mkdir -p squashfs
+# `luet install` is the only step requiring root
+luet install --system-target squashfs -y system/cos
+mkdir -p root/grub2/x86_64-efi
+mkdir -p root/cOS
+cp squashfs/usr/share/grub2/x86_64-efi/*.mod root/grub2/x86_64-efi
+cp squashfs/etc/cos/grub.cfg root/grub2
+
+mksquashfs squashfs root/cOS/recovery.squashfs -noappend -b 1M -comp xz -Xbcj x86
```
mudler, line 47: now this one can be consumed directly from the tree e.g. with: `luet install --system-target root/cOS/ -y recovery/cos-squash`

.github/workflows/build.yaml
mudler: good point! 

giving also another thought - the whole script could even be a package itself, like we do for generating the squashfs directly. The pro would be that we keep all the dependencies in the tree, so the user doesn't need any special tool to generate it. But that's just brain food for a follow-up if we think it makes sense

images/img-builder.sh
```diff
@@ -0,0 +1,79 @@
...
+    enable: true
+    urls:
+      - build
+    type: disk
+EOF
+
+# Create root-tree for COS_RECOVERY
+mkdir -p squashfs
+# `luet install` is the only step requiring root
+luet install --system-target squashfs -y system/cos
```
mudler, line 41: mmh, for consistency, would be nice to read those directly from the `manifest.yaml` file, so we can actually specify all of them in one place

images/img-builder.sh
```diff
@@ -0,0 +1,79 @@
...
+mkdir -p root/grub2/x86_64-efi
+mkdir -p root/cOS
+cp squashfs/usr/share/grub2/x86_64-efi/*.mod root/grub2/x86_64-efi
```
mudler, line 44: can we make a package out of this? so we can specify it in the manifest as well for consistency

davidcassany: I wound do it in a follow up issue. I believe the resulting package should also obsolete the `live` bootloader one, so we have a common ground, plus making a cross distro bootloader package might turn into an slightly complex task. I'd vote for a follow up issue.

davidcassany: Opened #279 for that purpose

.github/workflows/build.yaml
davidcassany: target for the makefile created

mudler (approved):

mudler:
@davidcassany I think we can merge it also if tests are not passing, there are no any notable change that could actually break them

davidcassany:
I agree, however I see there is a glitch un raw disk job, locally works but on GHA seams to be problematic

Issue #189, opened 2021-05-17
Don't start services on recovery partition


mudler:
**Is your feature request related to a problem? Please describe.**
Currently, when booting the recovery partition, all OEM files gets loaded, and this means that if we had services defined there, we would start them as well

**Describe the solution you'd like**
The Recovery partition should be used only in case of emergency, and it should not start the derivative services

**Describe alternatives you've considered**

**Additional context**

PR #241, opened 2021-06-04
Split recovery

mudler:
This PR adds another image that _can_ be used optionally as a recovery separately, in a _squashfs_ format.  It is getting installed during the ISO creation and if that is present, it will be used by the installer.

At that point, the scripts/grub will detect if a squashfs is present, and use it instead of the `img` file. This allows derivatives to use a split image with squashfs if needed,  use a single image, or use just `.img` files as before.

To set a default recovery image or package, set `RECOVERY_IMAGE` into /etc/cos-upgrade-image. It allows to override the default image/package for the recovery partition.

~~WIP as need some polishment (recovery and system can be a collection) and test-readaptation~~


packages/cos/build.yaml
```diff
@@ -19,9 +19,10 @@ requires:
 - name: "grub-config"
   category: "system"
   version: ">=0"
-- name: "selinux-policies"
-  category: "system"
-  version: ">=0"
+# XXX: Temp disable
```
mudler, line 22: self-note: leftover

packages/recovery-img/bootargs.cfg
```diff
@@ -0,0 +1,8 @@
+set kernel=/boot/vmlinuz
+if [ -n "$recoverylabel" ]; then
```
mudler, line 2: this allow the image to be used both for booting in active/passive and for recovery

iso/cOS.yaml
```diff
@@ -7,6 +7,7 @@ packages:
   isoimage:
   - live/syslinux
   - live/boot
+  - recovery/cos-img
```
mudler, line 10: from a derivative perspective - it's as easy as adding this package or not. If it's there squashfs will kick-in, otherwise the `img` mechanism should stay untouched.

mudler: @davidcassany this means that in the PoC script there is no need to generate the squashfs file anymore as part of that process, recovery/cos-img in this case it's containing the squashfs file only

packages/cos/collection.yaml
```diff
@@ -0,0 +1,9 @@
+packages:
```
mudler, line 1: this is a way to share the same build yaml file, https://luet-lab.github.io/docs/docs/concepts/packages/collections/

packages/cos/build.yaml
```diff
@@ -19,10 +19,17 @@ requires:
 - name: "grub-config"
   category: "system"
   version: ">=0"
+
+# Skip selinux policies for recovery
```
mudler, line 23: recovery shouldn't need selinux as its not running any core services. At this point it's dropped and only available for the system/cos. We might want to do other changes there as well, regarding for example a split cloud-config set. But I think it's out of purpose of this PR, and we can do a follow up later to iterate on the recovery image with #189

packages/installer/upgrade.sh
```diff
@@ -71,6 +71,18 @@ find_upgrade_channel() {
     if [ -z "$UPGRADE_IMAGE" ]; then
         UPGRADE_IMAGE="system/cos"
     fi
+
+    if [ -n "$UPGRADE_RECOVERY" ] && [ $UPGRADE_RECOVERY == true ] && [ -n "$RECOVERY_IMAGE" ]; then
```
mudler, line 75: This allows to set a `RECOVERY_IMAGE` into `/etc/cos-upgrade-image` allowing to override the default image/package for recovery

packages/recovery-img/bootargs.cfg
mudler: giving another thought, we probably should extend this by default to all images otherwise those can't be used for recovery anymore. It also opens up to the possibilit of just use squashfs as well.

packages/grub-config/definition.yaml
```diff
@@ -1,3 +1,4 @@
 name: "grub-config"
 category: "system"
-version: 0.0.3+3
+version: "0.0.5"
+recovery_type: "squashfs"
```
mudler, line 4: leftover

packages/cloud-config/oem/06_recovery.yaml
```diff
@@ -9,11 +9,19 @@ name: "Recovery partition boot setup"
...
-                echo >> /etc/issue
-                echo "You are booting from recovery mode. Run 'cos-reset' to reset the system to $VERSION" >> /etc/issue
-                echo >> /etc/issue
-            fi
+            echo >> /etc/issue
+            echo "You are booting from recovery mode. Run 'cos-reset' to reset the system to $VERSION" >> /etc/issue
+            echo " or CURRENT=active.img cos-upgrade to upgrade the active partition" >> /etc/issue
+            echo >> /etc/issue
+       files:
+       - path: /run/cos/recovery_mode
```
mudler, line 23: I just added this to have a canonical way to detect if we are in recovery_mode or not

tests/recovery/recovery_test.go
```diff
@@ -51,8 +111,9 @@ var _ = Describe("cOS Recovery upgrade tests", func() {
 			})
 		})
 
-		When("upgrade channel", func() {
-			It("upgrades to latest image", func() {
+		When("using upgrade channel", func() {
+			// TODO: This test cannot be enabled until we have in master a published version of cOS >=0.5.3
+			PIt("upgrades to latest image", func() {
```
mudler, line 116: I disabled this temporarly until we get a published version with this changeset as it wouldn't pass otherwise. I will re-add it later in a follow-up PR

packages/grub-config/config/grub.cfg
```diff
@@ -29,8 +29,13 @@ menuentry "cOS (fallback)" --id fallback {
 }
 
 menuentry "cOS recovery" --id recovery {
+  if search.file /cOS/recovery.squashfs ; then
```
mudler, line 32: I'm not very enthusiastic about this. If no squashfs is found GRUB will display a message that failed to find it which is very confusing for the user, but it works as expected as it boots correctly into recovery. I couldn't find a way to make it "less verbose" and avoid the misleading error message.

davidcassany (commented):
I believe keeping the choice of using squashfs filesystem for the recovery within the installer is causing quite some complexity. Why not assume the recovery is always squashfs and just let the user decide if the recovery system is created from the same active image or not.

This has the counter part that the `active.img` directly can't be reused and requires a mksquashfs call, but I don't think this is that bad, this is just few more minutes, the active partition is already unpacked into some local folder at that stage.

davidcassany:
>  Requiring **always** a squashfs image would force the user to build two images or either one image and post-process it separately, which can be cumbersome for very simple scenarios (and that's why I would keep it).

I am not sure if we are talking about the same. What I mean as **always** squashfs is to image and loop back always an squashfs filesystem on the recovery partition, regardless of the image contained in this filesystem is the same or not as the one contained in `active.img`. So I imagine an scenario where a single image is build `system/cos` and then packed as an ext[2-4] filesystem in `active.img` and as squashfs filesystem in `recovery.img`. The downside is just having to pack twice `system/cos`, but is only once at install time, I don't think this is that bad. This way we always treat the recovery image as a live squashfs system. Then at install time you have the choice to either use the same image for `active.img` or just use another one.

I think this setup would be pretty simple on our side allows us to omit any check on squashfs and easily keep separate boot args in for active and recovery systems as in #235.

packages/grub-config/config/grub.cfg
mudler: yup, that's why I had to end up with a simple if that seems to work just fine: https://github.com/rancher-sandbox/cOS-toolkit/pull/241/files#diff-82bb6cdc894381fa9e4ed0ad09b4bad8d0cbd30079b88abf7aa379d5f9ab881fR2 . the trick part is that each image that is intended to be used as recovery and/or as a bootable system have to deal with the case when it's being used for recovery or not

davidcassany:
> I can't follow what's the difference with the proposed approach in this PR. Here we just make it optional to have the recovery as .img file as it was before, but as soon as the installer/updater detect a squashfs present, then it will pick it up and use it as recovery. The squashfs can come from the same image or not - It's up to the user to setup `RECOVERY_IMAGE` if they want to use a different one

The downside of this setup is that you have to provide two different grub setups for the recovery, one to be used if the image is packed as a regular filesystem and another one to be used when the image is packed into a squashfs and loaded as a live system. So the grub.cfg line of `source (loop0)/etc/cos/bootargs.cfg` should be converted to something like:

```
if search.file /cOS/recovery.squashfs
  recovery=squashfsargs.cfg
...
else
  recovery=extendedfsargs.cfg
...
fi
...
source (loop0)/etc/cos/$recovery
```
And then it is up to the image designer to ensure both config files exist and that they are properly setup.

mudler:
> > I can't follow what's the difference with the proposed approach in this PR. Here we just make it optional to have the recovery as .img file as it was before, but as soon as the installer/updater detect a squashfs present, then it will pick it up and use it as recovery. The squashfs can come from the same image or not - It's up to the user to setup `RECOVERY_IMAGE` if they want to use a different one
> 
> The downside of this setup is that you have to provide two different grub setups for the recovery, one to be used if the image is packed as a regular filesystem and another one to be used when the image is packed into a squashfs and loaded as a live system. So the grub.cfg line of `source (loop0)/etc/cos/bootargs.cfg` should be converted to something like:
> 
> ```
> if search.file /cOS/recovery.squashfs
>   recovery=squashfsargs.cfg
> ...
> else
>   recovery=extendedfsargs.cfg
> ...
> fi
> ...
> source (loop0)/etc/cos/$recovery
> ```
> 
> And then it is up to the image designer to ensure both config files exist and that they are properly setup.

I'm not really convinced that splitting the cfg file between recovery and bootable images it's something that adds a value for the end-user. I expect the image designer to just copy the bootargs.cfg with almost - no need - to touch it at all.

davidcassany:
> I'm not really convinced that splitting the cfg file between recovery and bootable images it's something that adds a value for the end-user. I expect the image designer to just copy the bootargs.cfg with almost - no need - to touch it at all.

Well, the file is just a detail, the if clause is can be inside or outside. I was mostly pointing on the need of having separate bootargs depending on recovery/active instead of having squashrecovery / nonsquashrecovery. I am assuming recovery boot arguments are different to avoid the need of production settings in recovery boot (SElinux, immutable layout, etc.) and do not assume that the bootargs are equivalent as the ones for the active.img even in the case where `active.img` and `recovery.img` are exactly the same image.

IMHO if we want to keep all options bootargs could be:

```
if [ -n "$nonsquashedrecovery" ]; then
  set kernelcmd="console=ttyS0 root=LABEL=$label iso-scan/filename=$img panic=5"
else if [ -n $squashedrecovery ]; then
  set kernelcmd="console=ttyS0 root=live:LABEL=$squashedrecovery rd.live.dir=/cOS rd.live.squashimg=$img panic=5"
else
  set kernelcmd="console=ttyS0 root=LABEL=$label iso-scan/filename=$img panic=5 selinux=1 security=selinux"
fi
```
It can be in multiple files or not, this is a very minor detail. My point is that keep all options leads us to maintain three different sets of boot arguments.

Anyway, don't want to disturb you anymore. This can be discussed later on beyond the scope of this specific PR.

mudler:
> > I'm not really convinced that splitting the cfg file between recovery and bootable images it's something that adds a value for the end-user. I expect the image designer to just copy the bootargs.cfg with almost - no need - to touch it at all.
> 
> Well, the file is just a detail, the if clause is can be inside or outside. I was mostly pointing on the need of having separate bootargs depending on recovery/active instead of having squashrecovery / nonsquashrecovery. I am assuming recovery boot arguments are different to avoid the need of production settings in recovery boot (SElinux, immutable layout, etc.) and do not assume that the bootargs are equivalent as the ones for the active.img even in the case where `active.img` and `recovery.img` are exactly the same image.
> 
> IMHO if we want to keep all options bootargs could be:
> 
> ```
> if [ -n "$nonsquashedrecovery" ]; then
>   set kernelcmd="console=ttyS0 root=LABEL=$label iso-scan/filename=$img panic=5"
> else if [ -n $squashedrecovery ]; then
>   set kernelcmd="console=ttyS0 root=live:LABEL=$squashedrecovery rd.live.dir=/cOS rd.live.squashimg=$img panic=5"
> else
>   set kernelcmd="console=ttyS0 root=LABEL=$label iso-scan/filename=$img panic=5 selinux=1 security=selinux"
> fi
> ```
> 
> It can be in multiple files or not, this is a very minor detail. My point is that keeping all options leads us to maintain three different sets of boot arguments.
> 
> Anyway, don't want to disturb you anymore, I already distracted you more than it was need, sorry smile This can be discussed later on beyond the scope of this specific PR.

Ok, now I see the point :) yes it makes sense to split it even more between recovery/non-recovery. In the first pass implementation actually I was overriding the bootargs of the recovery directly before generating the squashfs. I went back at that approach because it felt more confusing, whereas an image designer should have looked in more places instead of one.

But yup, definetly this can be looked at in a separate PR, this one it's getting already too much big, ideally it should be a quick fixup on top of this one (with its own tests, review, etc :) )

mudler:
Ok, if tests are passing, should be good to merge!

mudler:
I've also added tests for upgrades from the recovery as during this work I've restored the capability to call `cos-upgrade` inside the recovery partition (somehow we have lost that between refactoring)

packages/grub-config/config/bootargs.cfg
```diff
@@ -1,3 +1,8 @@
 set kernel=/boot/vmlinuz
-set kernelcmd="console=tty1 root=LABEL=$label iso-scan/filename=$img panic=5 security=selinux selinux=1"
+if [ -n "$recoverylabel" ]; then
+    set kernelcmd="console=tty1 root=live:CDLABEL=$recoverylabel rd.live.dir=/ rd.live.squashimg=$img panic=5"
```
davidcassany, line 3: I'd say this doesn't boot for two reasons:

* According to the installer I'd say the `rd.live.dir` should be set to `rd.live.dir=/cOS`
* I'd say there is no such a CDLABEL` for the `COS_RECOVERY filesystem, since it is not an ISO. I'd use LABEL instead. Probaby dracut code uses CDLABEL and LABEL as the same thing though...

In case it fails to boot I's suggest to use something like:

```
set kernelcmd="console=tty1 root=live:LABEL=$recoverylabel rd.live.dir=/cOS rd.live.squashimg=$img panic=5"
```

mudler: it does boot, there are tests for it :)

mudler: I didn't touched here where we do put the squashfs files, so not sure if touching `rd.live.dir` makes sense.. or I'm missing something? I think it's just working because `rd.live.squashimg` contains the full path (`/cOS/...`)

mudler: changed from CDLABEL to LABEL done in https://github.com/rancher-sandbox/cOS-toolkit/pull/241/commits/70d4638cb5179d37b2dbc754ef1d83f16fae2149 :+1:

packages/recovery-img/bootargs.cfg
mudler: that was done :+1:

packages/installer/upgrade.sh
mudler: Added in the docs

davidcassany (approved):
Looks good to me! I keep my fingers crossed for the tests

davidcassany:
@mudler I believe this can be merged despite the failed tests. The pull suite passed all green.

PR #246, opened 2021-06-08
Add k9s and nerdctl utils

mudler:
nerdctl and k9s are extremely useful, and could be also reused in the recovery image.

This change also renames the `iso/cOS.yaml` to `manifest.yaml`. It also adds a `build` field with an optional list of packages to build.

See also: https://github.com/ibuildthecloud/os2/blob/master/Dockerfile#L60

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,7 @@ DESTINATION?=$(ROOT_DIR)/build
 #
 # yaml specification of build targets
 #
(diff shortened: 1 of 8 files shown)


mudler:
tricky part for the ci bot: We could keep the old versions around of those as well, and do not trash the old version when we bump them. so they can be also picked up by version when running `luet install`

Itxaka (approved):

mudler:
Finally green!

PR #247, opened 2021-06-08
Add kernel-firmware specific packages for opensuse flavor

mudler:
https://github.com/ibuildthecloud/os2/blob/master/Dockerfile#L29

### DIFF
--- a/values/opensuse.yaml
+++ b/values/opensuse.yaml
@@ -21,11 +21,21 @@ packages: >-
     qemu-guest-agent
     rng-tools
     systemd
+    vim
     parted
     dracut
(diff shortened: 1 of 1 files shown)


Itxaka (approved):

Itxaka:
Is this all firmware packages available? Should we add all so we cover most of the hardware that the kernel supports via firmware packages?

mudler:
> Is this all firmware packages available? Should we add all so we cover most of the hardware that the kernel supports via firmware packages?

would make sense indeed to have a default set to cover most common use cases,  I would say to just add them as needed :+1:

PR #251, opened 2021-06-09
Use the new luet-mtree event for upgrade checks

Itxaka:
We were using a bash script to call luet-mtree on upgrade to check for
valid mtree checksums.

luet-mtree 0.0.6 now supports working with events from luet itself,
which will do the same as the bash script and extract the image metadata
to check the mtree checksums.

This patch removes the old bash script and uses the luet-mtree plugin
directly to check the checksums, it also bumps luet-mtree to 0.0.6


mudler (approved):
Looks good!

Itxaka:
blocked by #254

Itxaka:
test keep failing but locally they pass :/

Itxaka:
Argh, recovery is failing due to the plugin probably! setting this to wip

davidcassany (approved):

PR #253, opened 2021-06-10
golang: add apt-get update before installing deps

mudler:
This avoids errors if repos are out-of-sync


davidcassany (approved):
Oh, good catch. Yes APT does not update automatically.

mudler:
going to merge it, otherwise master is locked. build seems fine, so we should be good!

PR #254, opened 2021-06-10
re-add 'make packer' target

kkaempf:


make/Makefile.test
```diff
@@ -58,6 +58,9 @@ ifeq ("$(BOXFILE)","")
 BOXFILE="$(ROOT_DIR)/packer/cOS.box"
 endif
 
+.PHONY: packer
+packer: $(BOXFILE)
+
 $(BOXFILE): $(PACKER) $(VBOXMANAGE)
```
Itxaka, line 64: wont this make sure that we cannot build other packer images unless we have vbox installed? Not nice for those of us that hate vbox :D

Itxaka:
```
# itxaka @ zeus in ~/projects/cOS-toolkit on git:kkaempf-readd-packer-target o [14:18:55] C:130
$ PACKER_ARGS="-var='feature=vagrant' -only qemu" make packer
'/usr/bin/VBoxManage' is missing and you must be root to install it.
make: *** [make/Makefile.test:50: "/usr/bin/VBoxManage"] Error 1
```

kkaempf: IIRC, `VBoxManage` was a prerequisite before. I am not able to run `make packer` without having Virtualbox installed.

I'd love to get rid of Virtualbox :wink:

Itxaka: No, vboxmanage was only a dependency when building vbox boxes. For qemu boxes there is no need to have vboxmanage installed at all, as I have been building qemu boxes daily without vboxmanage

Itxaka: No, vboxmanage was only a dependency when building vbox boxes. For qemu boxes there is no need to have vboxmanage installed at all, as I have been building qemu boxes daily without vboxmanage

kkaempf: Well, we need more documentation on make targets it seems.

Itxaka: I would say, just drop the vboxmanage in that target and make a `deps-vbox` target with it instead

kkaempf:
merged as #255

PR #255, opened 2021-06-10
re-add 'make packer' target

Itxaka:


Itxaka:
Forcing a remove of vboxmanage for now, we can rework it in later but this is blocking all PRs so hopefully someone approves it today so we can merge and trigger the other dependant PRs, and tomorrow morning we can wake up to green icons @kkaempf @davidcassany @mudler

mudler (approved):

kkaempf (approved):

kkaempf:
I still need to understand why "make packer" fails without `VBoxManage` installed locally. 🤔

Itxaka:
@kkaempf probably because unless you pass args to specify what image you want to build, like `-only qemu` it will try to build all images as defined in the `images.json` so the vbox will still try to build with vboxmanage and fail :)

davidcassany:
Yes, indeed, I usually use the  `-only virtualbox-iso` since I could not manage to run the tests using libvirt. Something fails in my env and I did not dig into, not even sure if this is a problem of my setup or not. @Itxaka did you manage the run `make test` on qemu/libvirt? If you did it encourages me to review my env :sweat_smile: 

I also had some issues with vagrant if I have both, libvirt provider and virtualbox providers installed and available. On my env Vagrant by default picks libvirt provider and apparently this was causing some issues too.

Itxaka:
@davidcassany unfortunately its not possible to run the tests under qemu. When the forward port is set at the end of the vagrant up (local 2222 to machine 22), once the machine is rebooted the forward is terminated and nevere recovered unless you do a vagrant reload (which restarts the machine and somehow fails to connect via ssh ¬_¬) and that is something in the core of the libvirt provider.

So for doing local deployment and manual tests it works, but for any automated tests that restarts the machine, its not valid :(

I had a quick look at the provider code but I could not see where that its setup, should probably have a look at the vbox provider to see what they are doing to maintain the forward at all times....

> I also had some issues with vagrant if I have both, libvirt provider and virtualbox providers installed and available. On my env Vagrant by default picks libvirt provider and apparently this was causing some issues too.

I think this depends on the box that you are using as the box is just an archive with a `provider` file inside that points to the provider that needs to be used. If you launch packer with no args, it builds both boxes, and if Im not mistaken, then end box has the same name for both vbox and qemu. So whoever ends up last will overwrite any boxes in there, and when adding the box it will add the box with the provider attached. We could work around that...in fact we should probably do in case we want to build several boxes in one machine...somehow...

davidcassany:
> once the machine is rebooted the forward is terminated and nevere recovered unless you do a vagrant reload (which restarts the machine and somehow fails to connect via ssh ¬_¬) and that is something in the core of the libvirt provider.

Oh damn! Now I know the problem... :disappointed: It is a problem on our side (or probably wicked side), on each reboot the machine gets a different IP (and a different ssh fingerprint, however this might not be an issue otherwise VBox tests would fail too) I saw that while testing images for AWS. Indeed I noticed it [here](https://github.com/rancher-sandbox/cOS-toolkit/pull/235#issuecomment-853762476), wicked daemon creates on the fly a randomized [DUID](https://en.wikipedia.org/wiki/DHCPv6#DHCP_unique_identifier) (stored in `/var/lib/wicked/duid.xml`) which is a IPv6 concept, however it looks like dnsmasq is also using it as an identifier for IPv4, probably it can be configured not to do so. Anyway, the problem is that wicked creates a random DUID, while IMHO it should be a deterministic ID based on `/etc/machine-id` and the MAC address.

Itxaka:
@davidcassany this actually seems to be a problem of the provider directly, see https://github.com/vagrant-libvirt/vagrant-libvirt/blob/af738b84ebfaccaad6acbfe66e4f810d1d6b13e5/README.md#forwarded-ports

They spawn an ssh command for each forwarded port, so when the machine reboots, the connection is severed and never recovered. So its by "design"? kind of :D

Itxaka:
Just thikning about this, maybe we dont need to do the port forwarding in vagrant and can use `COS_HOST` in combination with `vagrant ssh-info` to get the actual ip if the machine directly....

PR #264, opened 2021-06-14
Allow to upgrade from local directories

mudler:
~~WIP as needs tests~~ added and passing

https://github.com/ibuildthecloud/os2/blob/master/files/usr/sbin/cos-upgrade#L133

### DIFF
--- a/packages/installer/upgrade.sh
+++ b/packages/installer/upgrade.sh
@@ -159,22 +159,25 @@ upgrade() {
     export TMPDIR=$temp_upgrade
     export LUET_PRIVILEGED_EXTRACT=true
 
+    args=""
+    if [ -z "$VERIFY" ]; then
+        args="--enable-logfile --logfile /tmp/luet.log --plugin image-mtree-check"
+    fi
     if [ -n "$CHANNEL_UPGRADES" ] && [ "$CHANNEL_UPGRADES" == true ]; then
-        if [ -z "$VERIFY" ]; then
-          args="--enable-logfile --logfile /tmp/luet.log --plugin luet-mtree"
-        fi
+        echo "Upgrading from release channel"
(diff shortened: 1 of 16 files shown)


packages/installer/upgrade.sh
```diff
@@ -171,23 +171,22 @@ upgrade() {
...
-          args="--plugin image-mtree-check"
-        fi
         luet install $args --system-target $TARGET --system-engine memory -y $UPGRADE_IMAGE
         luet cleanup
+    elif [ "$DIRECTORY" == true ]; then
+        rsync -axq --exclude='host' --exclude='mnt' --exclude='proc' --exclude='sys' --exclude='dev' --exclude='tmp' ${UPGRADE_IMAGE}/ $TARGET
     else
-        args=""
-        if [ -z "$VERIFY" ]; then
-          args="--plugin image-mtree-check"
```
Itxaka, line 184: why removing the mtree check from here?

Itxaka (approved):

packages/installer/upgrade.sh
mudler: it's just a little bit of refactoring, I just moved it up actually to avoid code dups, `args` is just consumed later on need

Itxaka: argh, thats true, didnt see the indentation change, nice!

tests/upgrades-images-unsigned/upgrade_test.go
```diff
@@ -15,12 +15,9 @@ var _ = Describe("cOS Upgrade tests - Images unsigned", func() {
 		s.EventuallyConnects(360)
 	})
 
-	AfterEach(func() {
```
mudler, line 18: This suite is always resetting to verify that the version did actually changed, so I've just removed this.. should go a bit faster

davidcassany (approved):

PR #267, opened 2021-06-14
Prevent multiple job runs

### LINKED ISSUE #136: CI: make sure we have one run of the ci for branch
Currently the CI can run different jobs in parallel for the same branch, which could invalidate the end result.

For example, if we merge 2 PR in sequence, there should be only one job (latest) running.

We should also take extra care to not cancel jobs which are running the publishing step

See also:  https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency

### LINKED ISSUE #265: ci: bump bot is flooding GH actions
**cos-toolkit version:**
Not relevant

Not relevant

**Describe the bug**
our automation bot is creating a commit for each change, which triggers a new workflow for each commit

**To Reproduce**

**Expected behavior**
1 job for PR, or alternatively a PR with 1 commit only containing all the changes

mudler:
Make use of concurrency (https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency) to prevent multiple job runs for the same github reference.

We don't apply it to the whole workflow to prevent publish/release jobs to be canceled while being executed. Otherwise it could invalidate repository states.

Fixes #265
Fixes #136

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -21,8 +21,14 @@ on:
   create:
     tags:
       - v*
+
+
 jobs:
   docker-build:
+    concurrency: 
+      group: ci-docker-build-dismissable-${{ github.head_ref || github.ref }}-${{ github.repository }}
+      cancel-in-progress: true
+
     runs-on: ubuntu-latest
     strategy:
       matrix:
@@ -53,6 +59,9 @@ jobs:
           cos-build $FLAVOR
 
   build:
+    concurrency: 
+      group: ci-dismissable-${{ github.head_ref || github.ref }}-${{ github.repository }}
+      cancel-in-progress: true
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


davidcassany:
I expected to see only `pull_request` tests or `push` tests, but not both of them. Wasn't it the intended behavior?

mudler:
> I expected to see only `pull_request` tests or `push` tests, but not both of them. Wasn't it the intended behavior?

Not really, that's another issue.. I've tried it here, but I cannot find a way with the `concurrency` keyword. It seems to me it needs a separate pipeline for that :shrug: 

With this change we avoid to run multiple jobs for force-pushes, or stacking new commits on a PR/branch. Still there would be a job for PR and one for the branch push

davidcassany:
> With this change we avoid to run multiple jobs for force-pushes, or stacking new commits on a PR/branch. Still there would be a job for PR and one for the branch push (so in case of an external PR not coming from a branch of cos-toolkit, we should see just one job only, even if there are more commits added to it later on)

Oh right, yes, that makes sense :+1:

davidcassany (approved):

mudler:
Ok, let's give it a shot, otherwise CI is really blocking us as the bumpbot manages to starve also the master branch..

PR #269, opened 2021-06-14
ci: split vbox/qemu into separate groups

mudler:
This avoids that vbox/qemu jobs skips each others (as they are running in parallel)


Itxaka (approved):

mudler:
(have to merge it as is, or PRs will be marked red as it skips one of the two jobs.. )

PR #271, opened 2021-06-14
ci: allocate capacity fitting the runner specs

mudler:
hopefully should make tests go faster by allocating (almost) all the available space of the runner

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -249,6 +249,9 @@ jobs:
             packer/*.box
           if-no-files-found: error
   tests:
+    env:
+      VAGRANT_CPU: 3
+      VAGRANT_MEMORY: 10240
     concurrency: 
       group: ci-dismissable-${{ github.head_ref || github.ref }}-${{ github.repository }}-${{ matrix.flavor }}
       cancel-in-progress: true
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -249,6 +249,9 @@ jobs:
             packer/*.box
           if-no-files-found: error
   tests:
+    env:
+      VAGRANT_MEMORY: 3
+      VAGRANT_CPU: 10240
```
mudler, line 254: well, macOS runners have 14GB, but keeping 4gb spare just in case... don't want the whole process to be OOMKilled without notice

Itxaka (approved):
Nice!

davidcassany (approved):

mudler:
@davidcassany @Itxaka I've also tweaked the packer template and boosted cpu/ram there. Default was 1CPU/512MB of ram, that could explain why it was so flaky...

davidcassany (approved):

packer/images.json
```diff
@@ -84,6 +87,8 @@
       "arch": "amd64",
       "build": "dev",
       "disk_size": "50000",
+      "cpus": "3",
+      "memory": "8192",
```
davidcassany, line 91: Do we have this huge amount of memory in GH Actions?  I am wondering if this could lead to swap in some way. I usually run my manual tests with 2048M of memory on qemu and I have not seen any memory issue.

PR #284, opened 2021-06-16
Several speed up for tests

Itxaka:
- Disable key injection for vagrant
 - Enable simple user/pass ssh
 - Enable host dns resolver for vm
 - Disable synced folder for vagrant
 - Disable check for additions in vbox
 - Use vagrant snapshot for restoring the state of the vm after each
test
 - Enable shorter timeout for scp-ing the logs after a test failure


tests/upgrades-images-unsigned/upgrade_test.go
```diff
@@ -10,13 +10,21 @@ import (
...
+	})
+
 	BeforeEach(func() {
 		s = sut.NewSUT()
 		s.EventuallyConnects(360)
 	})
 
 	AfterEach(func() {
-		s.Reset()
+		sut.ResetWithVagrant()
```
davidcassany, line 27: I was also wondering it would it be better to simply remove the `AfterEach` and append its logic to `BeforeEach`, probably it adds an extra step on the first one, but at least it does not run a `Reset` after a failure. On failure I would expect to simpley error out quick and do nothing (so it keeps the failed environment locally and it also exits quick on GHA).

davidcassany (approved):

tests/upgrades-images-unsigned/upgrade_test.go
Itxaka: I can plug that only in case of success we should reset the vm back, good suggestion!

tests/fallback/fallback_test.go
```diff
@@ -9,12 +9,23 @@ import (
...
+		sut.SnapshotVagrantDelete()
+	})
+
 	BeforeEach(func() {
 		s = sut.NewSUT()
 		s.EventuallyConnects()
 	})
 	AfterEach(func() {
-		s.Reset()
+		if CurrentGinkgoTestDescription().Failed == false {
```
mudler, line 26: mmh not sure if this is going to work, I'm pretty sure the AfterEach Reset here is required to have a pristine state on each `It` block, but I could recall wrongly

tests/recovery/recovery_test.go
```diff
@@ -9,16 +9,27 @@ import (
 
 var _ = Describe("cOS Recovery upgrade tests", func() {
 	var s *sut.SUT
+
+	BeforeSuite(func() {
+		sut.SnapshotVagrant()
```
mudler, line 14: Can't we make s.Reset smarter and check for e.g. if we are running on vagrant to use the snapshot feature instead of `cos-reset` directly ?

tests/fallback/fallback_test.go
Itxaka: for this to work we probably have to drop the retry, so when we fail we fail hard and we leave the vm so it can be inspected.

Otherwise, if the test was a success this will reset it to its original state and other tests can continue

tests/recovery/recovery_test.go
Itxaka: yes, no problem there. I just wanted to see if this passes tests and cuts the time by half like in my local machine :o

Will add that!

mudler (approved):

tests/recovery/recovery_test.go
Itxaka: Done!  Now it checks at the suite start if vagrant is in a running state, and stores the value to use in the teardown

tests/fallback/fallback_test.go
Itxaka: retry also dropped!

Itxaka:
finally passes...and its now on need of an update FML :sob:

PR #286, opened 2021-06-16
Persist ssh fingerprint

mudler:


packages/cloud-config/oem/00_rootfs.yaml
```diff
@@ -25,3 +25,23 @@ stages:
...
+    - if: '[ -z "$(blkid -L COS_SYSTEM || true)" ]'
+      name: "Persist ssh fingerprint"
+      commands:
+      - |
+        # persist ssh fingerprint
+        if [ -d /usr/local/etc/ssh ]; then
+          mkdir /etc/ssh || true
+          for i in /usr/local/etc/ssh/*.pub; do cp -rf $i /etc/ssh; done
+        fi
+  network:
```
davidcassany, line 37: I would do it the other way round. In some late step (after network) if there is no `/usr/local/etc/ssh` I would just create it with the contents from `/etc/ssh/` this way we don't have to always restart ssh service. Probably we should review cos-setup stages to make them explicit.

mudler: right, but we don't have a clean way ATM to hook e.g. 'after ssh has been started' , in this way we just start ssh manually once if no fingerprints are present, we aren't restarting it everytime

mudler:
I'm not sure if we actually want this, or rather document something among the lines of:

```yaml
stages:
   initramfs:
     - if: '[ -z "$(blkid -L COS_SYSTEM || true)" ]'
       name: "Persist"
       commands:
       - |
            target=/usr/local/.cos-state

            # Always want the latest update of systemd conf from the image
            mkdir -p ${target}/etc/systemd/
            rsync -av /etc/systemd/ ${target}/etc/systemd/
            # Only populate ssh conf once
            if [ ! -e ${target}/etc/ssh ]; then
            mkdir -p ${target}/etc/ssh/
            rsync -av /etc/ssh/ ${target}/etc/ssh/
            fi
            # make /tmp tmpfs
            cp -f /usr/share/systemd/tmp.mount ${target}/etc/systemd/system/
            # undo /home /opt mount from cos immutable-rootfs module
            sed -i '/overlay \/home /d' /etc/fstab
            sed -i '/overlay \/opt /d' /etc/fstab
            umount /home
            umount /opt
            # setup directories as persistent
            for i in root opt home var/lib/rancher var/lib/kubelet etc/systemd etc/rancher etc/ssh usr/libexec; do
            mkdir -p ${target}/$i /$i
            mount ${target}/$i /$i -t none -o bind
            done
            # This is hidden so that if you run some selinux label checking or relabeling the bind
            # mount won't screw up things.  If you have two files at different paths they will get
            # labeled with two different labels.
            mkdir -p ${target}/empty
            mount ${target}/empty ${target} -o bind,ro
            # persist machine-id
            if [ -s /usr/local/etc/machine-id ]; then
            cat /usr/local/etc/machine-id > /etc/machine-id
            else
            mkdir -p /usr/local/etc
            cp /etc/machine-id /usr/local/etc
            fi
            # ensure /var/log/journal exists so it's labeled correctly
            mkdir -p /var/log/journal
```

Closing for now

PR #288, opened 2021-06-16
Allow upgrades from an extracted folder

ibuildthecloud:
This PR allows upgrades to happen from an extracted folder which allow a simple approach to upgrade using github.com/rancher/system-upgrade-controller/.

Also notice I change the rsync args.  I ran into issue with using AX in a container, but I kind of prefer we don't rely on ACLs and xattrs to be preserved. Are ACLs and xattrs part of the OCI spec? I would think those attributes can't be relies on being the tars from a docker image.

### DIFF
--- a/packages/installer/upgrade.sh
+++ b/packages/installer/upgrade.sh
@@ -165,16 +165,22 @@ upgrade() {
         luet install $args --system-target $TARGET --system-engine memory -y $UPGRADE_IMAGE
         luet cleanup
     else
-        args=""
-        if [ -z "$VERIFY" ]; then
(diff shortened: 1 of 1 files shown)


mudler:
👍  we have to merge this one with the additional tests provided in #288

ibuildthecloud:
@mudler I had no idea you were already merging all the stuff from my repo. Do you even need this PR? Should I close it?

mudler:
> @mudler I had no idea you were already merging all the stuff from my repo. Do you even need this PR? Should I close it?

Yup no worries, I'll take care of merging this one with my branch :+1:

PR #289, opened 2021-06-17
Add nerdctl and k9s

ibuildthecloud:
What I'd like to do is for RancherOS that I pull content from only
three sources

1. Leap
2. cOS Toolkit
3. Rancher built projects

So for example rancherd I can download that binary directly because
I can trust the build process, but for something third party like k9s
I'd prefer that be maintained in the toolkit.

Eventually the cOS toolkit is going to need to use FIPS compliant
go toolchain (essentially https://github.com/golang/go/tree/dev.boringcrypto/misc/boring),
so everything will have to be rebuilt by us.

### DIFF
--- a/packages/toolchain/k9s/build.yaml
+++ b/packages/toolchain/k9s/build.yaml
@@ -1,0 +1,26 @@
+requires:
+- name: "golang"
+  category: "build"
+  version: ">=0"
+env:
+- PATH=$PATH:/usr/local/go/bin
+prelude:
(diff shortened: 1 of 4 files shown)


Itxaka:
both tools are already merged: https://github.com/rancher-sandbox/cOS-toolkit/tree/master/packages/utils

ibuildthecloud:
Yeah, for sure. I had no idea y'all were already packaging up everything.

PR #298, opened 2021-06-18
Rework github workflows

Itxaka:
This is a full rework of our workflows.

The current problem is that our single workflow is a 2 hour long process
that can fail for a variety of reasons, not always having to do with the
actual code changed, and retesting means it will take another 2 full
hours to test and we have to rebuild the whole thing.

This patch splits everything into their own workflows and uses the
workflow_dispatch to triggers those workflows down the line. That way if
only one step fails we can retrigger that workflow only and it should
follow up the rest of the steps with no issues.

This patch also moves the docker-build into a separate workflow of its
own as that doesnt interact with the rest of the workflows.

Also the nosquashfs workflow is kept on a single file in case we want to
deprecate it.

This requires a PERSONAL ACCESS TOKEN to be added to the secrets to
trigger the other workflows as github does not allow using the
GITHUB_TOKEN for that in case there is recursive workflows.

The only downside is that we no longer see the full workflow chart in
one place as we did before, but all workflows should post their status
and link into the PR for easy access, so they wont be lost.


Itxaka:
This wont run until is merged so the only way of testing it is:

 - Get a PERSONAL ACCESS TOKEN and set it as a secret in your own fork (var: PERSONAL_TOKEN)
 - Merge this pull request into your fork
 - Create a PR against your fork

As I have merged it on mine, you can have a look at this PR that should show all the steps: https://github.com/Itxaka/cOS-toolkit/pull/23

kkaempf (approved):
lgtm

davidcassany (approved):
Looks good to me

Itxaka:
closing in favor of #319

PR #304, opened 2021-06-21
[POC] Add github runner scripts for ECP

Itxaka:


kkaempf:
I'd recommend to guard the `self-hosted` actions with conditionals like [here](https://github.com/epinio/epinio/blob/cf995a5d9e1b7c72829c218c5f73b4456bf257cb/.github/workflows/main.yml#L19).
Otherwise 'foreign' PRs (containing malicious code) could run on internal runners.

Issue #220, opened 2021-05-27
Deploy image in COS_ACTIVE in YIP


davidcassany:
This card is about making a YIP plugin to deploy an image to COS_ACTIVE partition.

The image/channel to deploy from should be configurable. And also setting the COS_PASSIVE with same image or not should be configurable. This way it could be used for the first time deployment but also for upgrades.

Note the code from card could be reused for #108 or viceversa

PR #305, opened 2021-06-21
Recfactor immutable rootfs package and initrd creation

### LINKED ISSUE #65: Replace isoscan with a custom dracut script

### LINKED ISSUE #239: Persist dracut configuration
Instead of generating from cli, having the configuration persisted makes it easier to apply changes on top.

https://github.com/rancher-sandbox/cOS-toolkit/blob/master/packages/immutable-rootfs/build.yaml#L29 to persist

**Additional context**
```
# cat /etc/dracut.conf.d/50-immutable-rootfs.conf
hostonly_cmdline="no"
hostonly="no"
compress="xz"
omit_dracutmodules+=" multipath "
add_dracutmodules+=" dmsquash-live cos-immutable-rootfs "
```

davidcassany:
This PR includes three concepts:
    
* Drop isoscan use in favor of a self provided script that is not
  devoted to ISO images and offers some further flexibility. For
  instance this enables `rd.cos.debugrw` kernel parameter to boot on
  readwrite mode. Fixes #65
    
* Drop most dracut hooks in favor of early boot systemd services. This
   is relevant in case we want or need to recode the layout tools so
   we are not tied to shell scripts. This also properly adds the rootfs
   stage into cos-setup. This is relevant for #220 so early boot
   configurations can be easily applied.

* Move initrd creation from immutable-rootfs package an persist the
  dracut configuration in config files. Fixes #239
    

mudler (approved):

Itxaka (approved):
havent tested it but it looks good

Issue #221, opened 2021-05-27
Test and boot the PoC img in Azure


davidcassany:
From the outcome of #218 fine tune the image so that it is also bootable in Azure Cloud.

Requires #218

With the image-builder script in https://github.com/rancher-sandbox/cOS-toolkit/pull/235:

- [x] Try to generate an Image and import it in Azure
- [x] Try to SSH and boot the image

Itxaka:
I was able to boot an image and ssh into it with almost no changes. No ssh key thougth, even when I attached one to the image in azure....

First, the raw image needs to be rounded and converted to vhd format with a FIXED format:

```
rawdisk="disk.raw"
vhddisk="disk.vhd"

MB=$((1024*1024))
size=$(qemu-img info -f raw --output json "$rawdisk" | \
gawk 'match($0, /"virtual-size": ([0-9]+),/, val) {print val[1]}')

rounded_size=$(((($size+$MB-1)/$MB)*$MB))

qemu-img resize $rawdisk $rounded_size

qemu-img convert -f raw -o subformat=fixed,force_size -O vpc $rawdisk $vhddisk
```


Then you create a `Storage Account` and upload the vhd into a `Container` as a storage blob
Then you create an image of type `Linux` and `Gen 2` (UEFI) based of that storage blob
Then you got yourself an image! Launch a virtual machine using that image and boom, cos in azure.

PR #309, opened 2021-06-22
Add azure image generation from raw image

Itxaka:
This patch adds a Makefile target to transform the raw disk that we
generate into a VHD image ready for Azure.

The only things different is that you need the image to be rounded and
set as FIXED vhd instead of dynamic[0]

Closes: #221 

[0] https://docs.microsoft.com/en-us/azure/virtual-machines/linux/create-upload-generic#resizing-vhds


mudler:
Pipelines needs to be rendered now :)

mudler (approved):
Pipelines

Issue #108, opened 2021-04-19
Elemental-cli to unify UX


mudler:
Given the several components involved (and those that will come in the future), it is straightforward to create a single cli that embeds the needed parts from all the involved pieces:

- `luet build` could be easily imported as standalone module and created a separate cli command with the specific defaults set for our use-cases. E.g. `elemental build`
- `luet utils unpack` can be imported / replaced by `elemental pull`. Given that derivatives will have to push their final image to a `:latest` tag for rolling updates, loosing the upgrade channels capabilities. ( see https://github.com/rancher/wharfie for a possible replacement )
- `yip` can be used already as a standalone module and replaced by a `elemental cloudinit`
- Image verification (#107 outcome) by a `elemental verify`
- #56  can become directly `elemental build-iso` and replace the current luet extension, which is just a bash script. This means we could also drop yq and jq from the dependencies
- [cos-setup](https://github.com/rancher-sandbox/cOS-toolkit/blob/master/packages/cos-setup/cos-setup) can be an hypotetical  `elemental run-stage`
- [cos-features](https://github.com/rancher-sandbox/cOS-toolkit/blob/master/packages/cos-features/cos-feature.sh) can become `elemental features enable/disable/list`
- Image builder scripts (arm, raw images, etc) should be part of the CLI. E.g. `elemental build-raw ...`

This have the advantage that we can then tag in a unique piece the moving pieces and decide when to pick new versions of our internal dependencies. 

Besides, we can have one repository containing the CLI and package specs. The package specs can be embedded assets into the binary so they are carried over as part of the binary release. Along with the binaries required by derivatives developers when building a Dockerfile. In this way we have one portable binary which can be both used to rebuild the assets, but also to place them during build in an immutable (and reproducible) way.

## Action Items

To have a replacement for the current feature set, we need:
- [x] #881

Building Images, Packages, Derivatives:

The build part of elemental, should allow us to get rid of the makefiles target, CI specific code to build packages, image building scripts, and luet as a binary dependency.

- [x] #1217
- [x] #1227
- [ ] `elemental build --packages`:  re-use `luet build` code as library to build packages, possibly embedding https://github.com/rancher-sandbox/cOS-toolkit/blob/master/.github/build.go to build package definition from embedded or remote assets
[...]

PR #312, opened 2021-06-22
Unattended deployment on a vanilla image

### LINKED ISSUE #220: Deploy image in COS_ACTIVE in YIP

davidcassany:
This PR allows unattended deployments on top of vanilla images by
using simple cloud init configuration files that can be injected to the
system as part of the user data of public providers.
    
This PR includes an `images/user-data.yaml` file as an example.
    
Fixes #220

### DIFF
--- a/.github/workflows/build-pr.yaml
+++ b/.github/workflows/build-pr.yaml
@@ -232,7 +232,7 @@ jobs:
     needs: vbox-squashfs-opensuse
     strategy:
       matrix:
-        test: ["test-features", "test-smoke", "test-fallback", "test-recovery", "test-upgrades-images-signed", "test-upgrades-images-unsigned", "test-upgrades-local"]
+        test: ["test-features", "test-smoke", "test-fallback", "test-recovery", "test-upgrades-images-signed", "test-upgrades-images-unsigned", "test-upgrades-local", "test-deploys-images"]
     steps:
       - name: Install Go
         uses: actions/setup-go@v2
@@ -385,7 +385,7 @@ jobs:
     needs: vbox-nonsquashfs-opensuse
     strategy:
       matrix:
-        test: ["test-features", "test-smoke", "test-fallback", "test-recovery", "test-upgrades-images-signed", "test-upgrades-images-unsigned", "test-upgrades-local"]
+        test: ["test-features", "test-smoke", "test-fallback", "test-recovery", "test-upgrades-images-signed", "test-upgrades-images-unsigned", "test-upgrades-local", "test-deploys-images"]
     steps:
       - name: Install Go
         uses: actions/setup-go@v2
--- a/images/img-builder.sh
+++ b/images/img-builder.sh
@@ -32,7 +32,6 @@ mkdir -p root/grub2/x86_64-efi
 mkdir -p root/cOS
 cp recovery/usr/share/grub2/x86_64-efi/*.mod root/grub2/x86_64-efi
 cp recovery/etc/cos/grub.cfg root/grub2
-sed -i 's/${saved_entry}/recovery/g' root/grub2/grub.cfg
 luet install --system-target root/cOS -y recovery/cos-img
 
 # Create a 2GB filesystem for COS_RECOVERY including the contents for root (grub config and squasfs container)
--- a/packages/installer/deploy.sh
+++ b/packages/installer/deploy.sh
@@ -1,0 +1,233 @@
+#!/bin/bash
+set -e
+
+CHANNEL_UPGRADES="${CHANNEL_UPGRADES:-true}"
+FORCE=""${FORCE:-false}
+
+# 1. Identify active/passive partitions
+# 2. Deploy active and passive to the same image
+
+find_partitions() {
+    STATE=$(blkid -L COS_STATE || true)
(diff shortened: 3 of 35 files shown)


davidcassany:
This PR is rebased on top of #305, so it is blocked until the base gets merged.

mudler (approved):
Looks good here!

packages/installer/deploy.sh
```diff
@@ -0,0 +1,233 @@
+#!/bin/bash
+set -e
+
+CHANNEL_UPGRADES="${CHANNEL_UPGRADES:-true}"
+FORCE=""${FORCE:-false}
+
+# 1. Identify active/passive partitions
+# 2. Deploy active and passive to the same image
+
+find_partitions() {
```
Itxaka, line 10: everything in here seems very similar to cos-installer. Could we extract most of this functions into a cos-functions file and then source it in both scripts?

mudler: that sounds good, but I would rather do in a separate follow-up PR. The same applies for `cos-upgrade`, `cos-installer` and `cos-reset`

mudler:
base was merged, removing blocked label

mudler:
Can we add a small test here? Just to check we don't break cos-deploy accidentally

packages/cos/build.yaml
```diff
@@ -19,15 +19,10 @@ requires:
 - name: "grub-config"
   category: "system"
   version: ">=0"
-
-# Skip selinux policies for recovery
-{{if ne .Values.category "recovery" }}
```
mudler, line 24: Any specific reason for re-adding it? In case it's needed, we also need to update the recovery bootargs

packages/installer/deploy.sh
davidcassany: Yes I agree I would do it in a separate PR or even leave it as is until we start working on #108 IMHO all this scripts should be substituted by a client in the scope of #108. Probably a fast solution is to build a shell client so we can reuse many portions of the code and we not have to start from scratch, I am not very passionate about this option, but probably this is a good way to start splitting all these logic into pieces and compose some client skeleton.

packages/cos/build.yaml
davidcassany: Yes, because then at deploy time the labels can't be applied, it requires SE Linux binaries to apply the labels.

packages/installer/deploy.sh
davidcassany: Indeed I am not very happy of this `cos-deploy` script, but I though that abusing of `cos-upgrade` to also cover this use case was making the behavior far from obvious (very small code changes were needed, but then the _upgrade_ behavior looked pretty strange to my eyes). That's why I opted for creating its own script by just copying the `cos-upgrade` code, removing many parts and applying few logic changes. I duplicated quite some code, but since this is already the case for all other `cos-*` scripts I assumed we can go with it for now and once the images and CI stories are addressed raise #108 prio

.github/workflows/build-pr.yaml
```diff
@@ -232,7 +232,7 @@ jobs:
     needs: vbox-squashfs-opensuse
     strategy:
       matrix:
-        test: ["test-features", "test-smoke", "test-fallback", "test-recovery", "test-upgrades-images-signed", "test-upgrades-images-unsigned", "test-upgrades-local"]
+        test: ["test-features", "test-smoke", "test-fallback", "test-recovery", "test-upgrades-images-signed", "test-upgrades-images-unsigned", "test-upgrades-local", "test-deploys-images"]
```
mudler, line 235: See here: https://github.com/rancher-sandbox/cOS-toolkit/blob/master/.github/PIPELINES.md, this change needs to be applied in the template, otherwise it would be overwritten each time we regenerate the workflows

images/img-builder.sh
```diff
@@ -32,7 +32,6 @@ mkdir -p root/grub2/x86_64-efi
 mkdir -p root/cOS
 cp recovery/usr/share/grub2/x86_64-efi/*.mod root/grub2/x86_64-efi
 cp recovery/etc/cos/grub.cfg root/grub2
-sed -i 's/${saved_entry}/recovery/g' root/grub2/grub.cfg
```
Itxaka, line 35: umm, I missed this but doesnt this means that the raw image will try to boot to `cos` directly and fail? even if we run user data, we need to boot into a "working" system to deploy that data...

Itxaka:
also from my side testing the local raw image seems that now its broken when running it from recovery, stops trying to mount COS_PERSISTENT

images/img-builder.sh
Itxaka: ah I see, it will keep trying until it falls under cos recovery :+1:

Itxaka:
ahh this seems to be an issue with space, if we dont provide the space needed for the disk expansion to occurr, there doesnt seem to be any errors shown, only the failure to mount cos_persistent...

PR #321, opened 2021-06-24
Re-enable recovery upgrade test

mudler:
We had to disable this because we didn't had published changes
supporting the scenario. 
It also fixes a leftover from a rebase that I believe slipped from the merges due to the automatic PR merge features (seems it is flaky and catches success early if there are no results back from the CI yet.. :shrug: )

### DIFF
--- a/packages/installer/upgrade.sh
+++ b/packages/installer/upgrade.sh
@@ -160,19 +160,24 @@ upgrade() {
     export LUET_PRIVILEGED_EXTRACT=true
 
     args=""
-    if [ -z "$VERIFY" ]; then
-        args="--enable-logfile --logfile /tmp/luet.log --plugin image-mtree-check"
+    if [ -z "$VERIFY" ] || [ "$VERIFY" == true ]; then
+        args="--enable-logfile --logfile /tmp/luet.log --plugin luet-mtree"
     fi
+
     if [ -n "$CHANNEL_UPGRADES" ] && [ "$CHANNEL_UPGRADES" == true ]; then
         echo "Upgrading from release channel"
+        set -x
         luet install $args --system-target $TARGET --system-engine memory -y $UPGRADE_IMAGE
         luet cleanup
+        set +x
     elif [ "$DIRECTORY" == true ]; then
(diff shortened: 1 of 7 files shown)


mudler:
@Itxaka seems it was a rebase mistake that slipped.. apparently GH feature for auto-merge is a bit flaky, if you check the PR #264  was merged even if tests weren't green

packages/installer/upgrade.sh
```diff
@@ -160,19 +160,24 @@ upgrade() {
     export LUET_PRIVILEGED_EXTRACT=true
 
     args=""
-    if [ -z "$VERIFY" ]; then
-        args="--enable-logfile --logfile /tmp/luet.log --plugin image-mtree-check"
+    if [ -z "$VERIFY" ] || [ "$VERIFY" == true ]; then
```
Itxaka, line 163: Is there something that sets this var to true?

packages/installer/upgrade.sh
```diff
@@ -160,19 +160,24 @@ upgrade() {
...
+        args="--enable-logfile --logfile /tmp/luet.log --plugin luet-mtree"
     fi
+
     if [ -n "$CHANNEL_UPGRADES" ] && [ "$CHANNEL_UPGRADES" == true ]; then
         echo "Upgrading from release channel"
+        set -x
```
Itxaka, line 169: Is this a leftover from debugging?

Itxaka (approved):

packages/installer/upgrade.sh
mudler: it could be set to true if someone sets it in the `/etc/cos-upgrade-image` file, that file gets sourced to be able to set custom behavior of `cos-upgrade` by default

packages/installer/upgrade.sh
mudler: yes/no.. I've added it for debugging, but I'm unsure to keep it. It seems useful for debug, but also as a informative output of what's going on.. UX definetly deserve its own story :)

Issue #275, opened 2021-06-15
Add Known issues to our docs


mudler:
**Is your feature request related to a problem? Please describe.**
There are at least two upstream issues that users might face depending on their build environments, specifically https://github.com/moby/moby/issues/37965 and https://github.com/docker/for-linux/issues/480

**Describe the solution you'd like**
A section in the docs that describes workarounds for Known issues. At the moment SELinux building might fail depending on the storage driver being used, and due to https://github.com/moby/moby/issues/37965 multi-stage copy could fail without `DOCKER_BUILDKIT` enabled.

**Describe alternatives you've considered**

**Additional context**
#274

PR #324, opened 2021-06-25
Add packages and update docs

mudler:
This PR is multi-fold:

- It adds `gomplate` as we now use it to generate CI pipelines
- It adds `rancherd`
- It adds `luet-makeiso` to our tree. Once it gets published, we can consume it in https://github.com/rancher-sandbox/cOS-toolkit/blob/master/scripts/get_luet.sh and drop the dependency of the official luet-repo repository.
- Update docs:
   -  Adds a "Known issues section" #275 
   - Adds docs about importing the raw image in AWS
   - Adds docs about creating bootable images that can be used in `cOS` during upgrades. This encompasses key concepts over https://github.com/ibuildthecloud/os2 while providing also examples.

~~PR is draft as I need to test the examples provided first~~


mudler:
I can't test the from scratch example above without a published change of cOS that is in this PR, I will amend in case it's not working in a follow-up. I need another round to consume go-mplate in any case.

examples/scratch/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+COPY --from=luet /usr/bin/luet /usr/bin/luet
+COPY --from=ca /etc/ssl/certs/. /etc/ssl/certs/
+
+# Copy the luet config file pointing to the cOS repository
+ADD conf/luet.yaml /etc/luet/luet.yaml
+ENV USER=root
+ENV LUET_NOLOCK=true
+SHELL ["/usr/bin/luet", "install", "-y", "-d"]
+
+RUN system/cos
```
mudler, line 19: this is the one I couldn't test till the end. It fails during install because `/etc/hostname` is already populated by docker

packages/cos/build.yaml
```diff
@@ -108,3 +108,9 @@ excludes:
 - ^/usr/local/share
 - ^/usr/local/src
 - ^/usr/local/games
+
+- ^/etc/hostname
```
mudler, line 112: I've added those to excludes. Hostname is generated by docker, so not really needed, similarly for the rest.

docs/raw_image.md
```diff
@@ -0,0 +1,47 @@
...
+
+3. Import the disk as snapshot
+
+```
+aws ec2 import-snapshot --description "cOS PoC" --disk-container file://container.json
+```
+
+4. Followed the procedure described in [AWS docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/creating-an-ami-ebs.html#creating-launching-ami-from-snapshot) to register an AMI from snapshot. Used all default settings unless for the firmware, were I forced it to UEFI boot.
+
+5. Launch instance with this simple userdata:
```
Itxaka, line 32: is this really needed? I booted the instance with a different user data, not creating users and root/cos was available.

Also, what is really needed if you want to launch the raw image is to provide a layout file (like in the example user-data) so its possible to use the new cos-deploy to install the system.

Itxaka (approved):

docs/raw_image.md
mudler: good point! I'll update the docs :+1:

mudler: done!

PR #325, opened 2021-06-25
ci: disable trigger on path

mudler:
It is not supposed to filter, but it actually triggers an additional
build (on branches).

See also https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#on


davidcassany (approved):
LGTM

PR #327, opened 2021-06-25
Make ln command idempotent

ibuildthecloud:
I'm not sure why but cos-setup-rootfs seems to be executing
twice on boot in my setup. That may be an issue in and of
itself, but regardless this unit can not be executed twice
because the second time the ln command will fail. On the
second execution it will try to create /system/system but
that will fail since system is pointing to /sysroot/system
which is read only.

### DIFF
--- a/packages/cos-setup/cos-setup-rootfs.service
+++ b/packages/cos-setup/cos-setup-rootfs.service
@@ -7,7 +7,7 @@ Conflicts=initrd-switch-root.target
 
 [Service]
 Type=oneshot
-ExecStartPre=/usr/bin/ln -s /sysroot/system /system
+ExecStartPre=/usr/bin/ln -sf -t / /sysroot/system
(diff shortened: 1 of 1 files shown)


davidcassany:
I am also not sure why is it being executed twice, I also realized about that, the fix is coming in  #312. One possible explanation is that this is a oneshot service that does not remain after exit and this is required by `initrd-root-fs.target`, however after this stage there is `initrd-fs.target` which requires `initrd-root-fs.target`, probably at that point systemd kicks in `cos-setup-rootfs.service` again as part of  `initrd-root-fs.target` requirements (even IMHO it doesn't make much sense doing if the direct dependency is already active...). Anyway, adding `ReamingAfterExit=yes` to cos-setup-rootfs.service seems to fix the issue.

davidcassany (approved):

mudler (approved):

PR #331, opened 2021-06-25
Always enable cdrom datasource

ibuildthecloud:
I'm in the middle of refactoring quite a bit of things so haven't fully tested this change.  One thing that confused me is that the path is set to /system/oem.  Isn't that path read-only?  I'm confused on how this works.

### DIFF
--- a/packages/cloud-config/oem/07_cloud-metadata.yaml
+++ b/packages/cloud-config/oem/07_cloud-metadata.yaml
@@ -10,5 +10,5 @@ stages:
    network:
      - name: "Cloud providers datasources"
        datasource:
-         providers: ["aws", "gcp"]
+         providers: ["aws", "gcp", "cdrom"]
          path: "/system/oem"


davidcassany:
> I'm in the middle of refactoring quite a bit of things so haven't fully tested this change. One thing that confused me is that the path is set to /system/oem. Isn't that path read-only? I'm confused on how this works.

Indeed this `/system/oem` setup does not work as it is read-only. I can't remember how it happened that I did it this way, my bad... I guess these are leftovers from my tests over squashfs where the whole root is overlayed as a live image.

I'd simply remove this file and keep the current `00_rootfs.yaml` that allows to pull user-data from the cloud before setting almost anything, so it give you freedom to configure `/run/cos/cos-layout.env` as part of the data source from cloud providers.

davidcassany:
Closing this PR in favor of #343, feel free to reopen if any doubt or concern.

PR #338, opened 2021-06-28
Add aws image generation based of the raw image

### LINKED ISSUE #302: Build AWS Images with Packer
This card is about having a packer template for AWS by using an image created by #235 

**Additional context**
Part of #208

Itxaka:
Uses the new cos-deploy script to install the image from recovery and
the user-data facility to partition the disk

As a side effect it moves from using json to using HCL for packer as json is no longer receiving any features and HCL is the recommended config language for packer

If HCL is not liked, I can easily revert to using plain json, but it had some issues on selecting the ami image using regex. HCL allows us to use "cos*recovery"  and use latest to pick the latest recovery image and not have any issues with selecting the wrong image. I could not make the same work on json templates.

Fixes #302 


Itxaka:
This has been tested by creating a raw image from master, uploading it to s3, creating a snapshot then turning that snapshot into an ami (cos_0.5.6-recovery.raw)

Then launching the packer process and deploying version 0.5.5 from the repositories snapshooting the image and manually booting from it to test that it was working.

packer/images.json.pkr.hcl
```diff
@@ -0,0 +1,128 @@
...
+      "if [ -n \"${var.feature}\" ]; then mount /dev/disk/by-label/COS_OEM /oem; cos-feature enable ${var.feature}; fi"
+    ]
+    pause_after = "30s"
+  }
+}
+
+  provisioner "shell" {
+    only = ["amazon-ebs.cos"]
+    inline = [
+      "cos-deploy --docker-image quay.io/costoolkit/releases-opensuse:cos-system-${var.cos_version}",
```
mudler, line 114: I would just leave this open as an entire variable, like `container_image`. In this way the template can be used to bootstrap also images from external images too

Itxaka: ah good point!

packer/images.json.pkr.hcl
```diff
@@ -0,0 +1,115 @@
...
+    inline = ["INTERACTIVE=false cos-installer --config /90_custom.yaml /dev/sda",
+      "if [ -n \"${var.feature}\" ]; then mount /dev/disk/by-label/COS_OEM /oem; cos-feature enable ${var.feature}; fi"
+    ]
+    pause_after = "30s"
+  }
+
+  provisioner "shell" {
+    only = ["amazon-ebs.cos"]
+    inline = [
+      "cos-deploy --docker-image ${var.cos_install_url}:cos-system-${var.cos_version}",
```
mudler, line 101: mmh images don't necessarly have to follow that schema,  users might also pick something like `myimage:testing` or `myimage:latest`. I think we should also account that verification might be skipped as well during this process

packer/images.json.pkr.hcl
Itxaka: like this? or the full url?

Thing is, we can take advantage of the cos version for several places

packer/images.json.pkr.hcl
Itxaka: Thats what I though :(

packer/images.json.pkr.hcl
mudler: I know, but this ideally should work also with custom images, not only for cOS ones, and we can't assume they are using `luet` for the building process. I would really let open to a cos-deploy-args even for now

packer/images.json.pkr.hcl
```diff
@@ -0,0 +1,115 @@
...
+  ami_name      = "cos-${var.cos_version}-${formatdate("DDMMYYYY", timestamp())}"
+  ami_description = "cos-${var.cos_version}-${formatdate("DDMMYYYY", timestamp())}"
+  instance_type = "t3.small"
+  region       = var.aws_region
+  secret_key   = var.aws_secret_key
+  ssh_password = "cos"
+  ssh_username = "root"
+  source_ami_filter {
+    filters = {
+      name                = "*cos*recovery*"
```
mudler, line 12: Also probably this one too should be customizable, with a default (like this one)

mudler (approved):

packer/images.json.pkr.hcl
Itxaka: how about now?

Still, I think we are confusing things here. We are providing packer templates for our own use, to publish our images. We dont have an obligation to provide packer templates for derivatives or other customers.

We can provide examples, but those examples are not gonna be anything out of the ordinary from any example packer templates as they will use our published image as base, which is ready for consumption and then they can expand on that to build their own images following the packer documentation on how to provision stuff.

We need a clear separation here, because making OUR own stuff too flexible just in case someone uses it, only causes us identity issues. Is this for us, is this for derivatives? We should have a clear line here, this is for OUR deployment of images, here are also some examples on how to consume our images with packer.

mudler: I thought that was clear from the card, but apparently it wasn't :sweat_smile: sorry ! 

The packer templates for Cloud are for derivatives/users only, as long as we don't need them internally to __provide__ base images. The fact that we use packer to spin up images used for tests is for convenience, but it is flexible enough to be re-used for derivatives too in case of need.

We could also use raw disks for tests for example, but then we don't have a way to see if we break the installation process.

Itxaka: but...wont we need to publish images to cloud providers so others can consume them? ITs a different process altogether I think. In aws, are we gonna provide derivatives a recovery image instead of a fully released image??? Doesn't seem good to me, they will need to actually install the system and then once the image is "working" deploy their stuff.. I havent seen an image that works like that in a cloud provider, they are all ready images so your first launch with user data can configure it completely.

mudler: what we are going to publish is only vanilla images where only the recovery partition is working. 

cOS is not meant to actually be used by per-se, but it's main use is to allow a complete system customization via container images. The recovery image just serves the purpose to bootstrap the minimal required to run the container image you supply (and taking later upgrades from)

Itxaka: Ok, that clarifies it a lot! Then I think down the line we may need a refactor of the templates to separate everything so it's easier for derivatives to work with! Even a readme to provide a pck.vars file by the user with the proper vars instead of cli args and some vars descriptions

PR #342, opened 2021-06-29
Fine tune user-data.yaml example file

davidcassany:
Just few minor tweaks to better align the user-data.yaml example with the installer.

### DIFF
--- a/images/user-data.yaml
+++ b/images/user-data.yaml
@@ -6,23 +6,18 @@ stages:
          # It will partition a device including the given filesystem label or part label (filesystem label matches first)
          device:
            label: COS_RECOVERY
-         # Only last partition can be expanded
-         # expand_partition:
-         #   size: 4096
          add_partitions:
            - fsLabel: COS_STATE
-             size: 8192
+             # 15Gb for COS_STATE, so the disk should be bigger than 17Gb
+             size: 15360
              pLabel: state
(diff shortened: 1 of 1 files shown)


mudler (approved):

davidcassany:
This example data worked fine out of the box in AWS with an instance of  4GB of memory

images/user-data.yaml
```diff
@@ -6,23 +6,18 @@ stages:
          # It will partition a device including the given filesystem label or part label (filesystem label matches first)
          device:
            label: COS_RECOVERY
-         # Only last partition can be expanded
-         # expand_partition:
-         #   size: 4096
          add_partitions:
            - fsLabel: COS_STATE
-             size: 8192
+             # 15GB for COS_STATE, the disk should be at least of 16GB
```
Itxaka, line 11: will this be enough? on my tests the recovery partition was 2Gb so this should be at least 17Gb?

Itxaka (approved):

images/user-data.yaml
davidcassany: Yes you are right, gonna write it should be bigger than 17Gb

Itxaka (approved):

mudler:
No sense to waste CI clocks here, merging :+1:

Issue #334, opened 2021-06-28
ci: Publish AMI images from master


mudler:
**Is your feature request related to a problem? Please describe.**
Consuming vanilla images for #208 

**Describe the solution you'd like**
the CI should publish the images generated to amazon AWS and make them public once tests are passing and the repository has been published.

Images have to be reusable by packer templates, so there is no need to manually update the raw disk anymore for downstream derivatives.

**Describe alternatives you've considered**
N/A

**Additional context**
#302 #217

Itxaka:
some preliminary work on this draft: https://github.com/rancher-sandbox/cOS-toolkit/pull/350

Basically:
 - Runs after publish
 - Only on master
 - For each flavor defined (opensuse, fedora, ubuntu)
 - Uses the latest cos*recovery image as base AMI (this have to be pushed manually from time to time unless we made that part of our workflow?
 - uses packer-aws target and 3 env vars to build the proper flavor
   - PKR_VAR_cos_version -> Sets the version to build the image from
   - PKR_VAR_flavor -> sets the flavor to name the image
   - PKR_VAR_aws_cos_install_args -> Sets the command used to deploy the image, using the --docker-image to pass the repo that we just published

I noticed thanks to this that both the fedora and ubuntu image do not pass the mtree verification, which probbly means that they have some extra files somehow? Have to check that.

Itxaka:
FYI I opened https://github.com/hashicorp/packer-plugin-amazon/issues/103 upstream in case there is interest into allowing the post-processor from packer `amazon-import` to also import and register snapshots as AMIs. That would make this process much simpler for us as we could upload the recovery images automatically to amazon for others to consume.

mudler:
Something seems still not working https://github.com/rancher-sandbox/cOS-toolkit/runs/2975749457?check_suite_focus=true

Itxaka:
Good, because I didnt expect it to work out of the box lmao

Itxaka:
after 5 or 6 fixes, images are now pushed from master to aws, set as public so anyone can consume them

PR #350, opened 2021-06-29
[CI] Publish AMI

Itxaka:
This patch adds a new job on github for the master branch that requires
the publish step to have passed and we use packer to build an AMI based
of the latest master image pushed and store it on aws

Closes: #334 


mudler:
out of curiosity, did you tried also to run the packer template against the published image?  maybe we should update somewhere in the docs what's the public url

mudler (approved):

Itxaka:
> out of curiosity, did you tried also to run the packer template against the published image? maybe we should update somewhere in the docs what's the public url

@mudler do you mean run with base for this the final image? like generate an image from this and then use that as a base for generating another one?

I havent done that...like...why would you do that? If you are generating an image you start from a "clean" one no? And nothing better than the recovery image for that :D

mudler:
> > out of curiosity, did you tried also to run the packer template against the published image? maybe we should update somewhere in the docs what's the public url
> 
> @mudler do you mean run with base for this the final image? like generate an image from this and then use that as a base for generating another one?
> 
> I havent done that...like...why would you do that? If you are generating an image you start from a "clean" one no? And nothing better than the recovery image for that :D

I mean, the published AMI image has the scope of being a base to put something on, we have to keep in mind that https://github.com/ibuildthecloud/os2 is one of the use cases. 

In that case I expect the user to be able to run the packer template on top of our published one - with the only input be the container image reference

Itxaka:
if by published AMI you mean the recovery one, then yes. Using the packer templates and pointing to it should result in a working cos image without touching anything else. If you want to customize then setting the install command to your docker-image should results in a customized AMI with your os on it, like pointing it to install `cos-deploy --no-verify --docker-image docker.io/ibuildthecloud/os2-0.1.0` should end up with a custom os2 AMI published...in theory.


Not tested of course, is there any published artifacts from os2 that I could test with?

mudler:
> if by published AMI you mean the recovery one, then yes. Using the packer templates and pointing to it should result in a working cos image without touching anything else. If you want to customize then setting the install command to your docker-image should results in a customized AMI with your os on it, like pointing it to install `cos-deploy --no-verify --docker-image docker.io/ibuildthecloud/os2-0.1.0` should end up with a custom os2 AMI published...in theory.
> 
> Not tested of course, is there any published artifacts from os2 that I could test with?

That already is enough :+1: but yup btw we have an image `quay.io/costoolkit/test-images:os2`

davidcassany (approved):

Issue #339, opened 2021-06-29
[spike] Using Fleet to automate OS upgrades


kkaempf:
GitOps with [Fleet](https://fleet.rancher.io) can already be used to configure and upgrade the application layer.

Goal of this research spike is to utilize Fleet to trigger OS upgrades.

Apparently Darren is already working in this direction. Get in contact with him first !

### Acceptance criteria

- [ ] reach out to Darren ( https://github.com/ibuildthecloud/os2 )
- [ ] demo showing Fleet used to trigger an OS upgrade
- [ ] documentation about the demo setup
- [ ] demo should partly work on baremetals (TBD)
- [ ] (if needed) creation of further cOS issues to address cOS deficits

mudler:
Darren is unavailable for few days, I'll start to have a look for documenting a Fleet+k3s deployment on top of cOS

mudler:
Note, to build https://github.com/ibuildthecloud/os2 from my PR ( https://github.com/ibuildthecloud/os2/pull/3 ):

```
export REPO=<IMAGE>
export LABEL=<TAG>
make build
docker push $REPO:$LABEL
make iso
```

mudler:
So I've exec'd into the rancher pod and created the admin user with `ensure-default-admin` and password can be reset with `reset-password`. although I'm still seeing sporadic failures in the webui and I can't create/import new cluster due to permission issues.  I think I should have seen the cluster being imported automatically instead, and the rancherd logs are still full of the logs above, investigating.

mudler:
Unrelated note to the above, os2 seems to ship a custom version of https://github.com/rancher/system-upgrade-controller which should take care of the upgrade part to be triggered with fleet.

davidcassany:
>  I'll start to have a look for documenting a Fleet+k3s deployment on top of cOS, following more or less https://github.com/rancher-sandbox/cOS-toolkit/wiki/Github-runner-on-k8s#cos-configuration style

@mudler I had to adapt several part of this yaml to deploy k3s, are you still using this documented yaml? I managed to deploy k3s and manually install fleet on top. I am trying to understand what fleet is how it could hypothetically be used to manage system upgrades. At glance it looks like this is out of scope of Fleet.

mudler:
That's what I've used to deploy k3s+fleet:

```yaml
name: "Fleet master"
stages:
   initramfs:
     - name: "Setting"
       hostname: "rancher-fleet-master"
   network:
     - commands:
[...]

PR #360, opened 2021-07-02
Add suc upgrade support

mudler:
This will make much easy to create images compatible with
[system-upgrade-controller](https://github.com/rancher/system-upgrade-controller).

See #339


davidcassany (approved):
:+1:

mudler:
ok same here, it's just squashfs tests passing for no reason.. but nosquashfs passed, merging it

mudler:
Same here non-squashfs are passing while squashfs tests seems to be broken on master too. Merging

PR #361, opened 2021-07-02
Improve cloud-config

davidcassany:
This commit moves the recovery_mode detection to the very very first
stage where 'yip' comes into play. This is specially helpful to make
use of it in later stages. In addition the default layout is changed for
recovery and system partitions, in recovery mode the COS_PERSISTENT
partition is ignored at boot time. This is is specially helpful for
vanilla images, with the change they can boo the recovery system as
they are, without requiring any partitioning at first boot.

### DIFF
--- a/packages/cloud-config/definition.yaml
+++ b/packages/cloud-config/definition.yaml
@@ -1,3 +1,3 @@
 name: cloud-config
 category: system
-version: "0.6.1"
+version: "0.6.2"
--- a/packages/cloud-config/oem/00_rootfs.yaml
+++ b/packages/cloud-config/oem/00_rootfs.yaml
@@ -8,19 +8,26 @@
 name: "Rootfs Layout Settings"
 stages:
   rootfs.before:
-    - name: "Layout configuration"
+    - name: "Pull data from provider"
+      datasource:
+        providers: ["aws", "gcp", "openstack", "cdrom"]
+        path: "/oem"
+  rootfs:
+    - if: '[ ! -f "/run/cos/recovery_mode" ]'
+      name: "Layout configuration"
       environment_file: /run/cos/cos-layout.env
       environment:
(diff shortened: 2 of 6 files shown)


mudler (approved):

packages/cloud-config/oem/00_rootfs.yaml
```diff
@@ -8,19 +8,26 @@
 name: "Rootfs Layout Settings"
 stages:
   rootfs.before:
-    - name: "Layout configuration"
+    - name: "Pull data from provider"
+      datasource:
+        providers: ["aws", "gcp", "openstack", "cdrom"]
```
Itxaka, line 13: could this datasource somehow affect the partition or expect the partition layout? I.E in the case of cloud-config you can execute stuff

packages/cloud-config/definition.yaml
```diff
@@ -1,3 +1,3 @@
 name: cloud-config
 category: system
-version: "0.6.1"
+version: "0.6.2"
```
mudler, line 3: As a reminder, CI now just builds what we bump, so if we don't bump `system/cos` and the squashfs, they won't be tested

davidcassany: Gonna bump them too then

packages/cloud-config/oem/00_rootfs.yaml
davidcassany: I am not sure I understand the question. The datasource might introduce new cloud config that might call the `layout` plugin and create or expand partitions. The default in-image config does not include anything like changing disk layout and it assumes a COS_OEM for the recovery system and COS_OEM and COS_PERSISTENT for the active and passive systems. The idea is that datasource is pulled at first stage of yip so **any** stage is valid in datasource, even `rootfs.before` can be called from data sources since `/system/oem` has precedence comapred with `/oem`.

PR #382, opened 2021-07-06
[ci] capture serial log on tests

Itxaka:
Sometimes having the serial log output can help us troubleshoot failures
in CI.

This patch enables the serial output to a file during tests and image
creation with vbox and stores it as an artifact as part of the PR
workflow.

Unfortunately as a side effect we cannot longer use the snapshot restore
method after each test to restore the test vm to a pristine state as it
will truncate the file, losing all logs. This patch just reverts to the
old reset method.


mudler (approved):

davidcassany (approved):

davidcassany:
Awesome this works, squashfs tests passed and the serial console is kept as an artifact for the nonsquashfs recovery test that didn't pass. :+1:

The serial console logs of the failed tests looks like the hypervisor failed to reboot... :thinking: but this unrelated to this PR.

Merging since required tests passed and its got a couple of approvals.

PR #366, opened 2021-07-05
Exclude /proc and /sys

mudler:
This fixes the following error while trying to install system/cos from a
container from scratch:
```
 DEBUG (docker.go:#66:github.com/mudler/luet/pkg/installer/client.(*DockerClient).DownloadArtifact) Cache file /var/cache/luet/packages/cos-system-0.5.7+2.package.tar.zst
 DEBUG (docker.go:#81:github.com/mudler/luet/pkg/installer/client.(*DockerClient).DownloadArtifact) Cache hit for artifact cos-system-0.5.7+2.package.tar.zst
 Failed installing package cos Error met while unpacking rootfs: lchown /sys: read-only file system
The command '/usr/bin/luet install -y -d system/cos' returned a non-zero code: 1
```

Signed-off-by: Ettore Di Giacinto <edigiacinto@suse.com>

PR #384, opened 2021-07-06
Add /etc/hosts file back

mudler:
While it fixes things for from scratch installs, it breaks many things:

- we should generate an /etc/hosts file during boot, ideally in
  cloud-init phase, but that looks cumbersome as we can just relay on
what's the default
- going to that route, and then excluding /proc, /sys, /run and /tmp
  requires further work on the bootstrap part, currently tests are not
passing (see #366)

It sounds easier instead to have a separate package for it just for
creating derivatives "from scratch" - in that way we don't interfere
with the standard system/cos which is supposed to run on ISO and other
medium too, where such folders are mandatory.

This addition adds a system/cos-container to just provide a separate package
with /proc, /run, /sys, /dev, /tmp stripped which is meant to be used for
from scratch scenarios.

Supersedes: #366

### DIFF
--- a/docs/creating_bootable_images.md
+++ b/docs/creating_bootable_images.md
@@ -95,7 +95,7 @@ ENV USER=root
 
 SHELL ["/usr/bin/luet", "install", "-y", "-d"]
 
-RUN system/cos
+RUN system/cos-container
 
 SHELL ["/bin/sh", "-c"]
 RUN rm -rf /var/cache/luet/packages/ /var/cache/luet/repos/
--- a/examples/scratch/Dockerfile
+++ b/examples/scratch/Dockerfile
@@ -16,7 +16,7 @@ ENV USER=root
 ENV LUET_NOLOCK=true
 SHELL ["/usr/bin/luet", "install", "-y", "-d"]
 
-RUN system/cos
+RUN system/cos-container
 
 SHELL ["/bin/sh", "-c"]
 RUN rm -rf /var/cache/luet/packages/ /var/cache/luet/repos/
--- a/packages/cos/build.yaml
+++ b/packages/cos/build.yaml
@@ -111,4 +111,21 @@ excludes:
 
 - ^/etc/hostname
 - ^/etc/machine-id
+
+{{ if eq .Values.name "cos-container" }}
+
 - ^/etc/hosts
+- ^/proc
+- ^/sys
+- ^/dev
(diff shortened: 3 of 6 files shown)


davidcassany:
MMm while I am totally in to get back `/etc/hosts` I believe the other folders still need to be cleared, otherwise is a bad design. I believe some failures are because the folder is excluded, but not its contents.

After second look I believe #366 mostly fails because it should have been
```yaml
- ^/tmp/.*
- ^/proc/.*
- ^/sys/.*
- ^/run/.*
- ^/dev/.*
```
instead of 

```yaml
- ^/tmp
- ^/proc
- ^/sys
- ^/run
- ^/dev
```

some of these folders, if not all, are part of the readonly image, so at boot systemd can't create mountpoint.

mudler:
> ```yaml
> - ^/tmp/.*
> - ^/proc/.*
> - ^/sys/.*
> - ^/run/.*
> - ^/dev/.*
> ```

Good point, but for the purpose of #366 excluding its content is not enough, luet will try to `lchown` directories too if they are present in the package. Let's say you do a mkdir `foo`, and `chown` it to something. Luet will try to reconstruct that on the host so even if empty that would hit `lchown` errors again.

That being said, I can try to add nevertheless

```yaml
- ^/tmp/.*
- ^/proc/.*
- ^/sys/.*
- ^/run/.*
- ^/dev/.*
```

to the excludes of `system/cos`, as it sounds a good thing to add anyway :+1:

mudler:
@davidcassany  added in https://github.com/rancher-sandbox/cOS-toolkit/pull/384/commits/6b56092e225f752b1c52def286f57ab2fea03d88 :+1:

davidcassany:
> to the excludes of `system/cos`, as it sounds a good thing to add anyway :+1: 

I am not sure about /dev folder though... Probably I'd start without this one.

mudler:
> > to the excludes of `system/cos`, as it sounds a good thing to add anyway +1
> 
> I am not sure about /dev folder though... Probably I'd start without this one.

Sure! I don't think anyhow we are recreating any of them with mknod, so should be safe

mudler:
ok, something is not ok by excluding those files, I'll split that in a separate PR

davidcassany (approved):

mudler:
All required tests are passing, merging

PR #387, opened 2021-07-07
Make use of grub2 environment block

davidcassany:
This commit sets the grub2 configuration to load /grubenv file, if any,
and checks the `next_entry` variable to set the default grub2 option.

This is helpful to set default boot in grub only for the next boot. From
cOS it is enough to run `grub2-editenv /oem/grubenv set
next_entry=recovery` to set the next reboot to default to recovery system.

### DIFF
--- a/packages/cos/collection.yaml
+++ b/packages/cos/collection.yaml
@@ -1,7 +1,7 @@
 packages:
   - name: "cos"
     category: "system"
-    version: "0.5.7+5"
+    version: "0.5.8"
     description: "cOS base image, used to build cOS live ISOs"
     brand_name: "cOS"
     labels:
@@ -9,7 +9,7 @@ packages:
       autobump.revbump_related: "recovery/cos-img recovery/cos-squash"
   - name: "cos"
     category: "recovery"
-    version: 0.5.7+5
+    version: "0.5.8"
     brand_name: "cOS recovery"
     description: "cOS recovery image, used to boot cOS for troubleshooting"
     labels:
@@ -17,7 +17,7 @@ packages:
       autobump.revbump_related: "recovery/cos-img recovery/cos-squash"
   - name: "cos-container"
     category: "system"
-    version: 0.5.7+6
+    version: "0.5.8"
     brand_name: "cOS"
     description: "cOS container image, used to build cOS derivatives from scratch"
     labels:
--- a/README.md
+++ b/README.md
@@ -90,12 +90,16 @@ If you are looking after only generating a container image that can be used for 
 - [Sample repository](https://github.com/rancher-sandbox/cos-toolkit-sample-repo)
 - [EpinioOS sample repository](https://github.com/rancher-sandbox/epinio-appliance-demo-sample)
 - [Use Fleet to upgrade a cOS derivative](https://github.com/rancher-sandbox/cos-fleet-upgrades-sample)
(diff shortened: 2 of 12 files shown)


davidcassany:
@Itxaka @mudler I could not resist and did a PoC that worked out of the box, so here is the PR :) 

With this grub config change we can run `grub2-editenv /oem/grubenv set next_entry=recovery` and the follow up boot it will default to `recovery`, but it will default to recovery only once, then it will be unset again. If you'd like to change the empty default to something else you could do `grub2-editenv /oem/grubenv set default=fallback` and then the default would be set to `fallback` _permanently_ (obviously only as long as the /oem/grubenv file exists and it is configured as such).

I did not add a tests neither refactored any test since it collides with #382 and I did not want to conflict with it. But it should be pretty straight forward to use that on tests. Also probably a new test for that is not needed since it might be already tested on upgrade or recovery tests.

This PR might also be helpful to define a default reboot behavior in case we detect some obvious error at boot (we are dropped into an emergency shell, no login user is defined, etc.)

mudler (approved):
Nice! Looks good here and seems shouldn't even have side effects on old installs

packages/cos/collection.yaml
```diff
@@ -1,25 +1,25 @@
 packages:
   - name: "cos"
     category: "system"
-    version: "0.5.7+5"
```
mudler, line 4: I'd rather the minor version as there is a change in GRUB which wouldn't be got in from the old installs (unless they do a cos-reset which re-applies grub changes)

davidcassany:
> Nice! Looks good here and seems shouldn't even have side effects on old installs

Correct, that's my understanding too.

mudler:
Maybe also worth it to add to the docs as well? Easy to forget in PRs, but I see this quite handy in the long run also when debugging things going bad from recovery

Itxaka:
This is awesome!

Itxaka:
All tests green, nice change!

davidcassany:
> All tests green, nice change!

Yes, test-recovery had some issues. In fact, I do have concerns about these tests, since we can't get back to the initial state in any after running `cos-upgrade --recovery ...`, if we had a way to `cos-upgrade --docker-image ` then we could make these tests less error prone.

Itxaka:
mmmhhh, we maybe be able to get away with mounting a temporal dir from the runner into the vm, but that would imply doing a create-repo with `--type docker` first and dumping the images into the runner so they are available. Possible but cumbersome.

Or we could rework  the serial port stuff so it saves in different files using `timestamp` and then restore the snapshot functionality, so we get a pristine vm after each test... this seems simpler

mudler:
> mmmhhh, we maybe be able to get away with mounting a temporal dir from the runner into the vm, but that would imply doing a create-repo with `--type docker` first and dumping the images into the runner so they are available. Possible but cumbersome.
> 
> Or we could rework the serial port stuff so it saves in different files using `timestamp` and then restore the snapshot functionality, so we get a pristine vm after each test... this seems simpler

We could also avoid the docker repo type for tests, and just start an http server on the host (there is a small one embedded in luet for that purpose, `luet serve-repo`)  we are already passing the whole repo between the artifacts. It can be used also via http/locally as well

davidcassany:
> We could also avoid the docker repo type for tests, and just start an http server on the host (there is a small one embedded in luet for that purpose, `luet serve-repo`) we are already passing the whole repo between the artifacts. But I'd keep that for few tests, as testing with docker images has its own value as well

That's a nice option indeed.

Issue #297, opened 2021-06-18
recovery/cos-img results to inconsistent builds


davidcassany:
<!-- Thanks for helping us to improve cos-toolkit! We welcome all bug reports. Please fill out each area of the template so we can better help you. Comments like this will be hidden when you post but you can delete them if you wish. -->

**cos-toolkit version:**
<!-- Provide the version of system/cos or git checkout of cos-toolkit -->

**CPU architecture, OS, and Version:**
<!-- Provide the output from "uname -a"  -->
Leap 15.3 x86_64

**Describe the bug**
<!-- A clear and concise description of what the bug is. -->
After changing some package and bumping its version the command `PACKAGES="system/cos recovery/cos recovery/cos-img" make build`  does not result in an updated `recovery/cos-img` package. It does for `system/cos` and `recovery/cos` but it doesn't for the `recovery/cos-img`.

**To Reproduce**
<!-- Steps to reproduce the behavior, including the luet command used -->
recovery/cos-img results to inconsistent builds

Since few days that I am really struggling to figure what is actually included in `recovery/cos` image build.

My process froma clean cache state is the following:

1. Checkout and build required artifacts with:

sudo PACKAGES="system/cos recovery/cos recovery/cos-img" time -p make build create-repo

2. I do some changes (e.g I change `packages/installer/build.yaml` to include a new file and I add a `+1` in version from `packages/installer/definition.yaml`)

3. Rebuild with the same command as I previously did:

sudo PACKAGES="system/cos recovery/cos recovery/cos-img" time -p make build create-repo

Here I can see how `utils/installer` is being rebuild first for `system/cos` and the new version appears logs

```
    5/8 system/cos-0.5.3+4 ⤑ 🔨  build system/grub-config-0.0.6 ✅  Done
 📦  6/8 system/cos-0.5.3+4 ⤑ 🔨  build system/immutable-rootfs-0.0.19+14  starts
 >💧  copy< Package has multi-stage copy, generating required images
 >💧  copy< 1/1 ⤑ 🔨  build toolchain/yip->=0 generating multi-stage images for toolchain/yip->=0
 📦  Compiling toolchain/yip-0.9.2 .... ☕ 
 🌳  Build dependencies for toolchain/yip-0.9.2
  ↪  distro/base-0.20210528+2 🍃 
  ↪  build/golang-1.16.5 🍃 
[...]

PR #391, opened 2021-07-08
Use a different cache per OS flavor

davidcassany:
This PR is to ensure caches between flavor are not mixed. In https://github.com/rancher-sandbox/cOS-toolkit/suites/3187960702/artifacts/73622493 the built ISO is using a Fedora based image for recovery while it should be an openSUSE based one. This PR is to enforce caches are separate for each flavor.

### DIFF
--- a/make/Makefile.build
+++ b/make/Makefile.build
@@ -27,7 +27,7 @@ PUSH_CACHE?=false
 
 BUILD_ARGS?=--pull --no-spinner --only-target-package --live-output
 
-REPO_CACHE?=quay.io/costoolkit/build-cache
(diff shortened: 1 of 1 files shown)


mudler (approved):
While I'm ok with this, it _shouldn't be_ required. I'm puzzled.. If we are facing this it means there is some underlying problem, which makes me think might be related to #297 -  if image hashing is not unique in the cases of images which are a "join" between final images, it might explain also that case as well

mudler:
I think I got the underlying problem, I've opened https://github.com/mudler/luet/issues/234 to track it

PR #398, opened 2021-07-12
Use requires_final_images when building squash image

mudler:
Also move the package under the `cos` directory so it's more clear the linking between
the `cos` packages.

Docs: https://luet-lab.github.io/docs/docs/concepts/packages/specfile/#requires_final_images

See also #297


Itxaka (approved):
Maybe a link to the docs could be added, I had to go around to find what requires_final_images meant :+1:

davidcassany (approved):

Issue #378, opened 2021-07-06
[docs] How to use AWS Packer templates


mudler:
**Is your feature request related to a problem? Please describe.**
Right now a user that builds a container image and wish to use that in the cloud should dig in our templates to figure out how to use it

**Describe the solution you'd like**
A new documentation section to show how to use the packer template. The scenario is the following: the derivative developer has just built a container image, and wants to leverage our published AMI images to generate a new AMI image with his container.

PR #399, opened 2021-07-12
Publish Vanilla AMI in EC2

davidcassany:
This commit adds a makefile target to publish the cOS Vanilla image
that might be later on used on Packer templates.

It also adds an additional CI job to publish it as part of the CI
workflow.

Related to #378


davidcassany:
@mudler @Itxaka JFYI I tested the `publish-vanilla-ami-opensuse` in this [job](https://github.com/davidcassany/cOS-toolkit/actions/runs/1026166209) in fork of mine

.github/build.yaml.gomplate
```diff
@@ -187,7 +187,7 @@ jobs:
           yq w -i $MANIFEST.remote 'luet.repositories[0].type' 'docker'
           yq w -i $MANIFEST.remote 'luet.repositories[0].urls[0]' $FINAL_REPO
           sudo -E MANIFEST=$MANIFEST.remote make local-iso
-          COS_VERSION=$(yq r packages/cos/collection.yaml 'version')
+          COS_VERSION=$(yq r packages/cos/collection.yaml 'packages.[0].version')
```
Itxaka, line 190: I was pretty sure I changed this already, was it lost on some commit?

.github/build.yaml.gomplate
```diff
@@ -501,5 +479,50 @@ jobs:
           path: |
             *.vhd
           if-no-files-found: error
+
+    {{{ if $config.publishing_pipeline }}}
+
+  publish-vanilla-ami-{{{ $flavor }}}:
```
Itxaka, line 485: do we really need to push the raw image per flavor? AFAIK we only use the recovery part of it, are the recoveries different between flavors?

Itxaka: I say this, because we should push ubuntu/fedora images so we should remove the `skip_images_flavor: ["fedora","ubuntu"]` but that would make the vanilla upload 3 times

.github/build.yaml.gomplate
```diff
@@ -501,5 +479,50 @@ jobs:
...
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.AWS_DEFAULT_REGION }}
+      - name: Upload and publish vanilla image
+        run: |
+          make aws_vanilla_ami
+
+  ami-publish-{{{ $flavor }}}:
```
Itxaka, line 506: Currently this will only upload the suse image.

.github/build.yaml.gomplate
davidcassany: See this is only active for `$config.publishing_pipeline` which is only active for `master.yaml` and  `master.yaml` also includes 

```yaml
skip_tests_flavor: ["fedora","ubuntu"]                                          
skip_images_flavor: ["fedora","ubuntu"] 
```

which causes it only to be computed per  `opensuse`

.github/build.yaml.gomplate
davidcassany: Correct

images/aws_upload.sh
```diff
@@ -0,0 +1,58 @@
...
+
+[ -f "${disk}" ] || exit 1
+
+echo "Uploading to S3 Bucket: ${s3_bucket}"
+aws s3 cp "${disk}" "s3://${s3_bucket}" --metadata "Name=${disk_name}"
+
+echo "Import snapshot to EC2"
+task_id=$(aws ec2 import-snapshot --description "${disk_desc}" \
+    --disk-container "Description=${disk_desc},Format=raw,UserBucket={S3Bucket=${s3_bucket},S3Key=$(basename ${disk})}" \
+    --tag-specifications "ResourceType=import-snapshot-task,Tags=[{Key=Name,Value=${disk_name}}]" \
```
Itxaka, line 18: can you also tag it with `Flavor: recovery` ? This is because in the future if we want to introduce auto deletion, we should probably have different policies per flavor image (we may just want to have only 1 recovery at the time) so we can filter by this flavor tag easily :D

images/aws_upload.sh
```diff
@@ -0,0 +1,58 @@
...
+done
+
+[ ! "${task_status}" = "completed" ] && exit 1
+
+snap_id=$(aws ec2 describe-import-snapshot-tasks \
+    --import-task-ids ${task_id} \
+    | jq -r '.ImportSnapshotTasks[0].SnapshotTaskDetail.SnapshotId')
+
+echo "Tagging Snapshot"
+aws ec2 create-tags --resources "${snap_id}" --tags Key=Name,Value=${disk_name} Key=Project,Value=cOS
```
Itxaka, line 41: ditto

images/aws_upload.sh
```diff
@@ -0,0 +1,58 @@
...
+   --architecture x86_64 \
+   --virtualization-type hvm \
+   --ena-support \
+   --boot-mode uefi \
+   --root-device-name "/dev/sda1" \
+   --block-device-mappings "DeviceName=/dev/sda1,Ebs={SnapshotId=${snap_id}}" \
+   | jq -r '.ImageId')
+
+echo "Tagging AMI"
+aws ec2 create-tags --resources "${ami_id}" --tags Key=Name,Value=${disk_name} Key=Project,Value=cOS
```
Itxaka, line 56: ditto

images/aws_upload.sh
Itxaka: Also version would be nice to have :+1:

images/aws_upload.sh
Itxaka: Also version would be nice to have :+1:

images/aws_upload.sh
Itxaka: Also version would be nice to have :+1:

packer/aws/setup-disk.yaml
```diff
@@ -1,13 +1,16 @@
 name: "Default deployment"
 stages:
   rootfs.after:
-    - name: "Repart image"
+    - if: '[ -f "/run/cos/recovery_mode" ]'
```
Itxaka, line 4: :+1: :+1: :+1:

images/aws_upload.sh
davidcassany: These tags are only for the `import-snapshot` task... I thought they would apply for the resulting snapshot but they don't, I just did not remove them, it doesn't hurt and provides some information in case there is anyone watching

Itxaka: ahh indeed!

.github/build.yaml.gomplate
Itxaka: yeah, but we would need to publish ami images for both systems, which means it will trigger the vanilla image upload 3 times

images/aws_upload.sh
davidcassany: sha included

.github/build.yaml.gomplate
davidcassany: Ok I see, you suggest to name it without a flavor

davidcassany: What about now, it is statically forced for opensuse, it is named without a flavor and it only applies for master workflow.

images/aws_upload.sh
```diff
@@ -0,0 +1,61 @@
...
+
+[ ! "${task_status}" = "completed" ] && exit 1
+
+snap_id=$(aws ec2 describe-import-snapshot-tasks \
+    --import-task-ids ${task_id} \
+    | jq -r '.ImportSnapshotTasks[0].SnapshotTaskDetail.SnapshotId')
+
+echo "Tagging Snapshot"
+aws ec2 create-tags --resources "${snap_id}" \
+    --tags Key=Name,Value=${disk_name} Key=Project,Value=cOS Key=GITHUB_SHA,Value=$github_sha
```
Itxaka, line 43: more interesting is the `Key=Flavor,Value=$FLAVOR` here, as it will allow us to create deletion policies based on the flavor images

mudler (approved):

Itxaka (approved):
This is green, no point in blocking it, I can add the extra tags afterwards

PR #405, opened 2021-07-13
amazon: Enhance AMI name and tags

Itxaka:
This patch adds the github.sha to the build name of the AMI so we can
easily identify where the image came from if pushed from CI and to be
able to upload more than one image per day on CI as well.

Also adds it to the tags, and add a new Flavor tag, so we can use data
lifecycle policies to auto remove images and search for an X flavor by
using those tags.

The default value for no git_sha is none.


Itxaka:
This probably breaks https://github.com/rancher-sandbox/cOS-toolkit/pull/399 so dont merge it before that one is merge, this one is simpler to rebase.

davidcassany (approved):
LGTM

Itxaka:
rebased and reworked a bit on adding the same tags to the aws_upload.sh script

davidcassany (approved):

PR #406, opened 2021-07-13
gce: Add script to create ready-to-upload disk to gce

Itxaka:
Same as the other images, this adds a script to prepare the raw disk in
a format that can be uploaded to gce and then imported directly as image

GCE as has some caveats:
 - Image needs to be rounded up to the next Gb
 - Image needs to be tar.gzipped with oldgnu format
 - Image inside the tar needs to be called disk.raw

Those are the main changes from the normal raw disk

### DIFF
--- a/.github/build.yaml.gomplate
+++ b/.github/build.yaml.gomplate
@@ -480,6 +480,37 @@ jobs:
             *.vhd
           if-no-files-found: error
 
+  gce-img-{{{ $flavor }}}:
+    runs-on: ubuntu-latest
+    container: opensuse/leap:15.3
+    needs: raw-img-{{{ $flavor }}}
+
+    steps:
+      - name: Install OS deps
+        run: |
+          zypper in -y qemu-tools make tar gzip xz which curl
+      - uses: actions/checkout@v2
+      - name: Download result for build
+        uses: actions/download-artifact@v2
+        with:
+          name: cOS-raw-img-{{{ $flavor }}}
+          path: .
+      - name: Install toolchain
+        run: |
+          # Since some time /var/lock is a symlink to /run/lock, which doesn't exit in the container
+          rm -rf /var/lock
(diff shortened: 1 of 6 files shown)


Itxaka:
@davidcassany My thougths exactly. When it was 1 it was ok. 2 was starting to stretch. 3 is now way too much, and both gc and azure use raw_img as base so they can use the same file.

Mainly sent this because I was testing a bit and didnt want to leave the script lying around and forget about it. I dont even know if it works (it uploads and you cna import it as image, but Im not sure it works)

Itxaka:
reworked to build all at the same time, rename them properly and add extra info to their naming, both files and artifacts.

GCE image tested to work as expected :o

Itxaka:
Need to leave, writing down here the steps to upload and create an image from the raw disk:

```
$ gsutil cp cos-vanilla-0.5.8+2.tar.gz gs://cos_images/
$ gcloud compute images create cos-vanilla-0-5-8plus2 --project=suse-css-platform --source-uri=https://storage.googleapis.com/cos_images/cos-vanilla-0.5.8\+2.tar.gz --storage-location=eu --guest-os-features=UEFI_COMPATIBLE --family=cos-vanilla --labels="flavor=vanilla,name=cos,version=0-5-8_2"

```

where:
 - --guest-os-features=UEFI_COMPATIBLE is required so machines with this image boot with UEFI, otherwise it tries legacy boot and it fails
 - --family=cos-vanilla : `The family of the image. When creating an instance or disk, specifying
        a family will cause the latest non-deprecated image in the family to be
        used.` so this is used to filter if you were to choose your boot image as cos-vanilla it would pick the latest version.
 - --labels: for easy filtering.

mudler (approved):

mudler:
Nice! :tada:

davidcassany (approved):
Cool, looks good

PR #410, opened 2021-07-14
amazon: Copy vanilla image to all regions

Itxaka:

### DIFF
--- a/.github/build.yaml.gomplate
+++ b/.github/build.yaml.gomplate
@@ -529,5 +529,6 @@ jobs:
       - name: Upload and publish vanilla image
         run: |
           export github_sha="${GITHUB_SHA}"
+          export COPY_AMI_ALL_REGIONS="true"
           make aws_vanilla_ami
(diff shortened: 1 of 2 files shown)


Itxaka:
Not really tested...

mudler (approved):
Looks good, let's see how it behaves on master :+1:

Itxaka:
mmmh...build seems to not have started at all ?

davidcassany (approved):
Looks good.

My only concern, not really exclusively tied to this PR is that we are starting to upload quite some stuff in AWS. Would be nice to have some sort of garbage collector policy, so just keep up to certain amount of artifacts of a kind, or keep binaries only for X time, etc.

Itxaka:
@davidcassany indeed, once we have LifecyclePolicy permissions we can start adding policies to keep only X images or images newere than X days ago

mudler:
The only way to test this is on master :)

PR #415, opened 2021-07-15
ci: fixup raw image artifact name

mudler:
Should fix: https://github.com/rancher-sandbox/cOS-toolkit/runs/3074288128

### DIFF
--- a/.github/build.yaml.gomplate
+++ b/.github/build.yaml.gomplate
@@ -507,10 +507,13 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - name: Export cos version
+        run: |
+          echo "COS_VERSION=$(yq r packages/cos/collection.yaml 'packages.[0].version')" >> $GITHUB_ENV
       - name: Download result for build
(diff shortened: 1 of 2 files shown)


.github/build.yaml.gomplate
```diff
@@ -507,10 +507,13 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - name: Export cos version
+        run: |
+          echo "COS_VERSION=$(yq r packages/cos/collection.yaml 'packages.[0].version')" >> $GITHUB_ENV
```
davidcassany, line 512: It won't work there is no `yq` in this environment :sweat: We overlapped, I did the same an equivalent commit 5mins ago [here](https://github.com/rancher-sandbox/cOS-toolkit/commit/052cf819f4ede0ef954bd84c95513fdf9eabddc2)

davidcassany (changes requested):

.github/build.yaml.gomplate
mudler: closing this one then!

PR #416, opened 2021-07-15
Fix image reference in Vanilla image upload job

davidcassany:


davidcassany:
lets see if we are not missing any further detail :sweat_smile:

mudler (approved):

PR #424, opened 2021-07-16
Ensure sshd runs after deployment

davidcassany:

### DIFF
--- a/images/sample_user-data.yaml
+++ b/images/sample_user-data.yaml
@@ -8,16 +8,25 @@ stages:
            label: COS_RECOVERY
          add_partitions:
            - fsLabel: COS_STATE
-             # 15Gb for COS_STATE, so the disk should be bigger than 17Gb
-             size: 15360
+             # 10Gb for COS_STATE, so the disk should have at least 16Gb
+             size: 10240
              pLabel: state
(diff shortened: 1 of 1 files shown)


davidcassany:
@Itxaka suggested that login prompt should be blocked during the execution of `cos-deploy` in deployments using the Vanilla image.  This way a user can't login via ssh while in the background there is deployment and a followup reboot, this might be confusing and it is easy to collide somehow with the deployment process and break it. I liked the suggestion.
 
This PR does not block all login options, just ssh. Which I believe it sufficient enough for cloud use cases. In addition any user can still login using the serial console or regular ttys if they have access to it. Not blocking serial console logins I think is good, since they are used mostly for debugging purposes.

Itxaka (approved):
This is awesome. We would have to be careful when using this with packer thought as it could lead to timeouts while provisioning the image, but that is easily fixable

davidcassany:
> This is awesome. We would have to be careful when using this with packer thought as it could lead to timeouts while provisioning the image, but that is easily fixable

Well this is why I did not include it in packer user-data files. It is not suited for Packer use case. In fact this is only useful for anyone booting the Vanilla image directly.

PR #425, opened 2021-07-16
azure: Use shared image gallery for cos base

Itxaka:
Instead of using our private images, use the ones in the shared
image gallery which are available to the world. Also make them
configurable so if someone wants to use their own shared image
becuase they modified the vanilla image, they can override the
values


Itxaka:
NOT TESTED

We are out of PublicIpAddress in our account so I could not test this. Image is available and public so it should work.

Commented out the user-data part as it cannot be used currently.

Feel free to take over this.

mudler (approved):

mudler:
If it's not breaking aws build, I'd merge it as-is and we can iterate on top if something is missing

davidcassany (approved):

PR #427, opened 2021-07-19
ci: Add workflow to publish toolchain image

mudler:
This allows us to reuse the image inside docs and also to provide easier steps without having to build the main image.
I've already tested the workflow, and an image was already pushed here: https://quay.io/repository/costoolkit/toolchain?tab=info


davidcassany (approved):
Makes sense, so we have a reference image including the toolchain.

Issue #379, opened 2021-07-06
ci: publish CI artifacts on github releases


mudler:
**Is your feature request related to a problem? Please describe.**
Currently it's hard to consume cOS vanilla images, the user have to dig throughout the CI runs to find a working commit to grab the artefacts 

**Describe the solution you'd like**
We have a CI job (which is currently failing) that releases all the generated CI assets on tag. To note, we should take care also of the artefacts versioning.

**Describe alternatives you've considered**

**Additional context**

**Action Items**

- [x] Adapt luet-makeiso to take a version as input - https://github.com/mudler/luet-makeiso/commit/99672eea4ac386b1702a9aef9d60eec3b262fe60 https://github.com/rancher-sandbox/cOS-toolkit/pull/429
- [ ] Adapt CI to output artifact to specific naming ( Introduce version output in our make targets/refactor as necessary )  and fix the publish job to upload the artifacts we are interested into: https://github.com/rancher-sandbox/cOS-toolkit/pull/433
      - ISOs (?)
      - Raw images: base, gcp, azure
      - OVA, Vagrant, qcow2
      - Docker images links

mudler:
I've opened a separate issue for the next release here #437 - which includes the codenames changes too

PR #433, opened 2021-07-21
Add templates to workflow and rework release job

mudler:
- Split pipeline into template blocks
- Rework how versioning is constructed in the CI, adds a `helpers.sh` file to collect common code around versioning generation and it's used across the CI blocks
- Various minor fixes
- Release job now collects all artifacts including a txt file which is generated pointing to the generated final image of cOS
- Fixes packer builds https://github.com/rancher-sandbox/cOS-toolkit/pull/433/commits/118d8298e434bd5644a3fe51747c1d37616ae929

Part of #379

### DIFF
--- a/.github/PIPELINES.md
+++ b/.github/PIPELINES.md
@@ -6,3 +6,11 @@ Requirements:
 Pipelines are generated from `build.yaml.gomplate`, which is a gomplate template. To generate the pipelines, run `./pipelines.sh`.
 
 Each configuration in `config` is turned to its own pipeline, named `build-$NAME`.
+
+## Github Secrets
+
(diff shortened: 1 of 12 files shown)


mudler:
vhda seems to exceeds limits for releases artifacts (2.09GB, limit is 2GB). I'll see if compressing it is enough, or I'll have to try to see if we can remove some unnecessary packages from our base

mudler:
Ok, having still some inconsistencies with the versioning, but it's almost there: https://github.com/mudler/cOS-toolkit/releases/tag/vtest6

mudler:
Ok, should be good for review now, I'm just waiting another CI cycle to confirm that releases are ok (it should publish here: https://github.com/mudler/cOS-toolkit/releases/tag/vtest6 )

davidcassany (approved):
Looks good :+1:

PR #435, opened 2021-07-21
Add Google Compute Platform Packer template

### LINKED ISSUE #380: Build GCP Images with Packer
This card is about having a packer template for GCP by using an image created by #235 

**Additional context**
Part of #208

davidcassany:
This commit adds a first basic version of a GCP Packer template to
deploy cOS or cOS derivatives systems into Google Cloud Platform
using a cOS Vanilla image as the bootstrap platform.

Fixes #380 


mudler (approved):
Nice! Looks good here!

mudler:
maybe we could also add a small subsection on the docs on how to run the template with packer?

davidcassany:
> maybe we could also add a small subsection on the docs on how to run the template with packer?

Yes we can do that, I was about to start adding a small article about it. However the problem here is that the images are not public, so the instructions will require manually committing the vanilla image to your project first.

PR #436, opened 2021-07-22
Rename variable for consistency

davidcassany:


davidcassany:
Just noticed I documented the template with this variable name, which is consistent with the naming I followed in all other variables.

mudler (approved):

Issue #437, opened 2021-07-22
Release 0.6.0


mudler:
- [x] Create new container repositories and assign permissions to CI (https://quay.io/repository/costoolkit/releases-green)
- [x] #442 
- [x] Check for references to old repos and update them ( release channels, code, sample repos, etc ...)
- [x] Release new version with artifacts https://github.com/rancher-sandbox/cOS-toolkit/pull/446
- [ ] Update docs to point to release page on github and to the new repositories, with a link to a brief explanation on how to use each artifact. e.g. cOS-Seed ISO as a deployer for custom images, how to import qcow, and so on so forth https://github.com/rancher-sandbox/cos-toolkit-docs/pull/14

mudler:
Created #456 and #457 as follow ups

PR #442, opened 2021-07-23
Change flavor names after codenames

mudler:
This commit changes flavors from opensuse, ubuntu, fedora to their respective codenames: green, orange and blue

Part of #437 


davidcassany (approved):
Looks good, it is hard to tell if there is any missing occurrence, however as long as the tests pass I would not fear on doing the change and assume that little leftovers somewhere might pop up in the future.

PR #443, opened 2021-07-23
Include the region of the AMI to tag

davidcassany:


davidcassany:
@mudler the only idea that comes to my mind is adding the region of the image to tag, just in case it filter by the default region... I have not tested it

mudler (approved):
Looks good here, let's see :eyes: !

PR #445, opened 2021-07-23
Fix recovery upgrades

davidcassany:
This commit fixes a bug related environment variables setting for
cos-upgrade command. `cos-upgrade --recovery` was not using
the recovery/cos image for upgrades since at the time the upgrade image
was evaluated and decided the `RECOVERY_UPGRADE` variable was not set
yet.

### DIFF
--- a/packages/installer/upgrade.sh
+++ b/packages/installer/upgrade.sh
@@ -58,17 +58,22 @@ find_upgrade_channel() {
         source /etc/cos-upgrade-image
     fi
 
+    if [ -n "$NO_CHANNEL" ] && [ $NO_CHANNEL == true ]; then
+        CHANNEL_UPGRADES=false
+    fi
+
     if [ -n "$IMAGE" ]; then
         UPGRADE_IMAGE=$IMAGE
         echo "Upgrading to image $UPGRADE_IMAGE"
-    fi
+    else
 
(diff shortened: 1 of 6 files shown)


packages/installer/upgrade.sh
```diff
@@ -58,17 +58,22 @@ find_upgrade_channel() {
...
-    if [ -z "$UPGRADE_IMAGE" ]; then
-        UPGRADE_IMAGE="system/cos"
-    fi
+        if [ -z "$UPGRADE_IMAGE" ]; then
+            UPGRADE_IMAGE="system/cos"
+        fi
 
-    if [ -n "$UPGRADE_RECOVERY" ] && [ $UPGRADE_RECOVERY == true ] && [ -n "$RECOVERY_IMAGE" ]; then
-        UPGRADE_IMAGE=$RECOVERY_IMAGE
+        if [ -n "$UPGRADE_RECOVERY" ] && [ $UPGRADE_RECOVERY == true ] && [ -n "$RECOVERY_IMAGE" ]; then
```
davidcassany, line 74: UPGRADE_RECOVERY was not set here since it gets applied in from the command line parsing which used to happen after this method in former implementation. Hence the recovery image from `/etc/cos-upgrade-image` was never applied and it was defaulting to  `system/cos`

mudler (approved):
good catch!

PR #458, opened 2021-07-27
Use codenames to generate release channel url

mudler:

### DIFF
--- a/packages/cloud-config/build.yaml
+++ b/packages/cloud-config/build.yaml
@@ -6,12 +6,6 @@ steps:
 - mkdir -p /system /oem
 - cp -rfv oem /system
 - chmod -R 600 /system/oem
-{{ if .Values.distribution }}
-{{if eq .Values.distribution "opensuse" }}
-- sed -i 's/:FLAVOR:/opensuse/g' /system/oem/02_upgrades.yaml
-{{else if eq .Values.distribution "fedora" }}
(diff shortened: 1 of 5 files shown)


packages/cloud-config/build.yaml
```diff
@@ -6,12 +6,6 @@ steps:
 - mkdir -p /system /oem
 - cp -rfv oem /system
 - chmod -R 600 /system/oem
-{{ if .Values.distribution }}
-{{if eq .Values.distribution "opensuse" }}
```
mudler, line 10: We moved urls from https://quay.io/repository/costoolkit/releases-opensuse to https://quay.io/repository/costoolkit/releases-green, this was an oversight because the URL were generated by the "distribution" field which was untouched

packages/cloud-config/definition.yaml
```diff
@@ -1,3 +1,3 @@
 name: cloud-config
 category: system
-version: 0.7+1
+version: "0.8"
```
dragonchaser, line 3: why the quotes?

dragonchaser (approved):

packages/cloud-config/definition.yaml
mudler: That's a yaml annoying thing... the version is a string because it is semver compliant ( so can contain things like `+`, `-` and string characters) but gets parsed as float if you don't quote it explictly in this case (the `+` removes the ambiguity between string and numbers, but `0.8` it looks like a float )

PR #472, opened 2021-08-02
ci: Drop setup-docker job, not required anymore

mudler:


Itxaka (approved):
Would be good to explain int he commit why is not needed anymore in case we need to come back to this later.

dragonchaser (approved):
lgtm

mudler:
I can only guess that now it's part of the official VMs, they also added `yq` lately.. we should actually start to use our toolchain image instead, so we don't depend much on the environment

PR #478, opened 2021-08-03
Drop override of /etc/issue on installer

### LINKED ISSUE #163: After install, cos-toolkit issue file is displayed on derivatives
**cos-toolkit version:**
https://github.com/rancher-sandbox/cOS-toolkit/commit/680349531f07406685aaa200915f064c9d926527

Not relevant

**Describe the bug**
https://github.com/rancher-sandbox/cOS-toolkit/blob/680349531f07406685aaa200915f064c9d926527/packages/installer/installer.sh#L215 writes to the persistent storage, so if a derivative gets installed, after boot will get displayed the default password of cos and the content of the /etc/issue of cos-toolkit

**To Reproduce**
Install a derivative, e.g. https://github.com/rancher-sandbox/epinio-appliance-demo-sample

**Expected behavior**
To not see the cos-toolkit message 

## Action Items
- [ ] Drop the hardcoded post-installation cloud-init file installation
- [ ] Have a mechanism in place similar to what we do for `COS_INSTALL_CONFIG_URL`  that copies a custom cloud-config file also to  `/usr/local/cloud-config`. We can then have a default that can be opt-out but mimics the same behavior we have currently

Itxaka:
No need to overwrite anything on install, we already have the branding
on /oem that sets the /etc/issue and can be override by derivatives

Closes #163 

### DIFF
--- a/packages/installer/installer.sh
+++ b/packages/installer/installer.sh
@@ -222,27 +222,6 @@ do_copy()
         get_url "$COS_INSTALL_CONFIG_URL" $OEM
         chmod 600 ${OEM}
     fi
-    mkdir -p $TARGET/usr/local/cloud-config
-cat > $TARGET/usr/local/cloud-config/90_after_install.yaml <<EOF
-# Execute this stage in the boot phase:
-stages:
-   fs.after:
-     - name: "After install"
-       files:
-        - path: /etc/issue
-          content: |
-            Welcome to \S !
-            IP address \4
-
-            Login with user: root, password: cos
-            To upgrade the system, run "cos-upgrade"
-            To change this message permantly on boot, see /usr/local/cloud-config/90_after_install.yaml
-          permissions: 0644
-          owner: 0
-          group: 0
-EOF
-    chmod 640 $TARGET/usr/local/cloud-config
-    chmod 640 $TARGET/usr/local/cloud-config/90_after_install.yaml
 }
 
 SELinux_relabel()
--- a/packages/installer/definition.yaml
+++ b/packages/installer/definition.yaml
@@ -1,3 +1,3 @@
 name: "installer"
 category: "utils"
-version: 0.9+2
+version: 0.9+3


packages/installer/installer.sh
```diff
@@ -222,27 +222,6 @@ do_copy()
         get_url "$COS_INSTALL_CONFIG_URL" $OEM
         chmod 600 ${OEM}
     fi
-    mkdir -p $TARGET/usr/local/cloud-config
```
mudler, line 225: What about having a similar flag for copying a file over here like we do for /oem?

Itxaka: sure, the problem is not that I **dont know** how to do it. The problem is more philosophical, should we actually do it? what brings to the table, what are the advantages to it, what are the cons (more maintenance, another possible failure entrypoint), etc..

As I commented on the issue, I just dont see a user case for this when we have the user-data for deployment configuration AND the COS_INSTALL_CONFIG_URL for the installer, feels like going overboard with this.

mudler: my opinion in this is that there is a location meant for user usage, like /usr/local/.. and /oem. I see it like this: if you are a sysadmin and have some users in the host (well.. always a possibility right) you might want to not give them access to `/oem`, but you might want to give a narrowed down version of it which is inside `COS_PERSISTENT`. This is also a way to circumvent the fact that `COS_OEM` has a fixed size, kept small intentionally, and you could potentially encode chunk of binaries via YAML

mudler (approved):
I would keep around the possibility to write on `/usr/local/cloud-config` during install, but nevertheless the changes looks good and are inline with the card :+1:

packages/installer/installer.sh
Itxaka: I really dont see the use case of a derivative having users in the sytem tbh, like that is so far away from the current mood in the IT world :rofl: 

The space issue could be a problem indeed but...like there is the correct way to add binaries to the system by bulding packages, if someone wants to drop binaries during the install...like we cannot support all the cases in the world. A user could potentially do that, but they could also scp files into COS_OEM, or download an ISO or just create a huge file, etc... there is no way we can protect against all those things except pointing it to the proper way of deploying binaries (create a package, do your stuff in there, will end up as part of the system)

Itxaka: If you really think this is important, lets create a card to follow up on that so we dont forget about that :+1:

PR #484, opened 2021-08-04
Add new system/grub package

### LINKED ISSUE #279: Create a bootloader package that is part of `cos-recovery` and/or `cos-system` image
Current `system/cos` or `recovery/cos` they include grub2 bootloader. This card is about of splitting grub2 related artifacts into a separate package (currently part of the base package) so it can be easily consumed as part of image building procedures (ISOs and disks).

This will facilitate keeping all image related artifacts consistently aligned in `manifest.yaml`.

Itxaka:
Individual package with the grub packages installed so it can be reused
down the line

closes #279 

### DIFF
--- a/packages/grub-artifacts/build.yaml
+++ b/packages/grub-artifacts/build.yaml
@@ -1,0 +1,16 @@
+requires:
+  - name: "cos"
+    category: "system"
+    version: ">=0"
+
+steps:
+  - mkdir /grub-artifacts
+{{ if .Values.distribution }}
+  {{if eq .Values.distribution "opensuse" }}
+  - cp -rf /usr/share/grub2/* /grub-artifacts
+  {{else}}
+  - cp -rf /usr/share/grub/* /grub-artifacts
+  {{end}}
+{{end}}
+
+package_dir: /grub-artifacts
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -26,3 +26,4 @@ build:
  - selinux/rancher
  - selinux/k3s
  - recovery/cos
+ - system/grub-artifacts
--- a/packages/grub-artifacts/definition.yaml
+++ b/packages/grub-artifacts/definition.yaml
@@ -1,0 +1,3 @@
+name: "grub-artifacts"
+category: "system"
+version: "0.0.1"


mudler (changes requested):
I think the card was about the `grub2` bootloader artifacts rather than `grub` itself . The bootloader files are carried over for example in the raw image build process: https://github.com/rancher-sandbox/cOS-toolkit/blob/master/images/img-builder.sh#L33

values/orange.yaml
```diff
@@ -28,3 +24,9 @@ packages: >-
     rsync
 
 kernel_package: linux-image-generic
+
+grub_packages: >-
```
mudler, line 28: No need for a separate section here. You could have a package that extracts certains file from another, see for example https://github.com/rancher-sandbox/cOS-toolkit/blob/master/packages/cos/recovery-img/build.yaml. but  it can be as simple as having a package which `requires` the one that have the files you are interested in, and copy it back in the new

Itxaka:
@mudler does it make sense to have the artifacts separated? Im asking because the full package is just 9Mb and its straightforward (installs grub, thats it) compared against a smallest package but with a include/excludes or a copy from another package, you have to maintain lists of what needs to be copied or included/excluded, etc...

mudler:
> @mudler does it make sense to have the artifacts separated? Im asking because the full package is just 9Mb and its straightforward (installs grub, thats it) compared against a smallest package but with a include/excludes or a copy from another package, you have to maintain lists of what needs to be copied or included/excluded, etc...

yes it does because in this way we can just add a package entry in the manifest and make it consistent instead of hardcoding in several places what we extract, like here https://github.com/rancher-sandbox/cOS-toolkit/blob/master/images/img-builder.sh#L33

packages/grub-artifacts/build.yaml
```diff
@@ -0,0 +1,13 @@
+requires:
+  - name: "cos"
+    category: "system"
+    version: ">=0"
+
+
+unpack: true
+
+includes:
+  - ^/usr$
```
mudler, line 10: It looks good :+1:  I'm just thinking that _maybe_ it makes more sense to have the files in `/usr/share/grub2` in the package as top level instead of `/usr/share/*`.  Because looking at https://github.com/rancher-sandbox/cOS-toolkit/blob/master/images/img-builder.sh#L33 seems the file are just consumed like "blobs"

Itxaka: ahh that is a good point

mudler (approved):
Looks good!

mudler:
ouch, seems in ubuntu the location is different https://github.com/rancher-sandbox/cOS-toolkit/pull/484/checks?check_run_id=3253831513:
```
 
 #9 [5/5] RUN cp -rf /usr/share/grub2/* /grub-artifacts
 #9 sha256:d9a8051e7667eb6bbae5177c62248ae03b3cb886b1dd4feda83a030b7b3be454
 #9 0.314 cp: cannot stat '/usr/share/grub2/*': No such file or directory
#9 ERROR: executor failed running [/bin/sh -c cp -rf /usr/share/grub2/* /grub-artifacts]: exit code: 1
```

Itxaka:
yep, expected...more ugliness incoming to fix this :crying_cat_face:

mudler (approved):

PR #487, opened 2021-08-05
Add cloud config stages for installation/upgrade/deploy

### LINKED ISSUE #154: Installation/Upgrade as cloud-config stages
Customization during install and upgrade phase are not possible, e.g. even displaying a simple informative message

A stage, like we do for `boot`, `network`, ... but for `installation` and `upgrade`. In this way we allow to customize those aspects of the derivatives as well with a generic approach (and we keep everything in cloud-config)

**Additional context**

mudler:
Adds after-install, after-deploy, after-upgrade, before-install,
before-deploy, before-upgrade as cloud-config stages. This means that
now cloud configuration files can hook into those stages as well for
further customization.

Fixes #154

### DIFF
--- a/packages/cos/collection.yaml
+++ b/packages/cos/collection.yaml
@@ -1,7 +1,7 @@
 packages:
   - name: "cos"
     category: "system"
-    version: 0.6.2+1
+    version: "0.6.3"
     description: "cOS base image, used to build cOS live ISOs"
     brand_name: "cOS"
     labels:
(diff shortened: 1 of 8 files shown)


Itxaka (approved):
nice!

Itxaka:
should we also add some before reset? Just thinking that it may be useful for something down the line

mudler:
good point! added :+1:

Itxaka (approved):
was already green before, no point in waiting for tests to merge

PR #488, opened 2021-08-05
Add disk raw test

### LINKED ISSUE #486: Add small test for raw image
When modifying source, sometimes its possible that we break the raw image, but we dont have any tests for it

A simple test added to the tests that uses the raw image as base to check that at least it boots as expected

n/a

**Additional context**

Itxaka:
closes #486

### DIFF
--- a/.github/build.yaml.gomplate
+++ b/.github/build.yaml.gomplate
@@ -505,6 +505,40 @@
           if-no-files-found: error
 {{{ end }}}
 
+{{{define "raw_image_test"}}}
+  {{{ $config := (datasource "config") }}}
+  {{{ $flavor := . }}}
+  raw-image-test-{{{ $flavor }}}:
+    runs-on: macos-10.15
+    needs: raw-images-{{{ $flavor }}}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install yq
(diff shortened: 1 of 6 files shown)


Itxaka:
Lookign for some feedback as this boots the image but it doesnt seem to work as expected (also tested locally with the raw_disk)

Did we break the raw_disk already? Am I doing something weird?

Itxaka:
umm, I migth have broken the raw_disk with the grub artifacts?


```
   Falling back to `cOS recovery'

 hd0,gpt4error: ../../grub-core/loader/i386/efi/linux.c:277:can't allocate kernel.
error: ../../grub-core/loader/i386/efi/linux.c:98:you need to load the kernel first.

Press any key to continue...
```

Itxaka:
ahh seems like I forgot to increase the memory if the vm so grub fails to boot (not able to map the kernel maybe? mount the loop device?)

Itxaka:
can this be reviewed and merged? @mudler @dragonchaser only the vbox-squashfs failed due to virtualbox crapping out, the rest went green without issues :)

mudler (approved):
Looks good! having a simple smoke test for it it's nice, and shouldn't even take long in the test matrix

Issue #491, opened 2021-08-06
FIPS-enabled golang builds


mudler:
**Is your feature request related to a problem? Please describe.**
Binaries that we build on our own aren't FIPS-enabled

**Describe the solution you'd like**
The same set of golang packages that we build off internally but FIPS-enabled. Since FIPS enabled builds requires `CGO_ENABLED=1` we could have or either a set of `-static` packages or either a `-fips` based one, without loosing the static packages that we ship now which are handy in different contexts.

**Additional context**
https://github.com/golang/go/blob/dev.boringcrypto.go1.12/misc/boring/README.md

PR #492, opened 2021-08-06
Add cos-minimal meta package

mudler:
Adds meta/cos-minimal package which requires the minimal set of packages
in order to have a minimal functioning cOS derivative.

This is also in the long-term very handy if we are going to have a
separate set of packages which are fips-enabled (#491)

### DIFF
--- a/packages/meta/collection.yaml
+++ b/packages/meta/collection.yaml
@@ -1,0 +1,26 @@
+packages:
+- category: "meta"
+  name: "cos-minimal"
+  version: "0.6.1"
+  requires:
+  - category: toolchain
+    name: yip
+    version: ">=0"
+  - category: utils
+    name: installer
+    version: ">=0"
+  - category: system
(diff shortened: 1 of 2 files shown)


packages/meta/collection.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - category: system
+    name: immutable-rootfs
+    version: ">=0"
+  - category: system
+    name: grub-config
+    version: ">=0"
+  - category: system
+    name: cloud-config
+    version: ">=0"
+  - category: toolchain
```
mudler, line 24: I gave a thought at this. It's better that the meta explictly requests luet, so it is installing in the system the version from our repository. When building a derivative then you have a transient luet version, but at the end you pick up the one from the toolchain by default. 

The user it's still free to opt-out as they can install the packages singularly as it is right now

Itxaka (approved):

PR #493, opened 2021-08-06
Use templates for golang-based packages

mudler:
That's just a first-pass to introduce shared templates in the repo, I'm pretty sure we will find other cases and improve this one

Docs: https://luet-lab.github.io/docs/docs/concepts/packages/templates/#shared-templates

### DIFF
--- a/packages/templates/golang.yaml
+++ b/packages/templates/golang.yaml
@@ -14,9 +14,10 @@
 {{ $version:= get . "version" }}
 {{ $org:= get . "org" }}
 {{ $repo:= get . "repo" }}
+{{ $host:= get . "host" | default "github.com" }}
 - |
   PV="{{$version}}" && \
   PACKAGE_VERSION=${PV%\+*} && \
-  mkdir -p /luetbuild/go/src/github.com/{{$org}} && cd /luetbuild/go/src/github.com/{{$org}} && \
-  git clone https://github.com/{{$org}}/{{$repo}} && cd {{$repo}} && git checkout "$PACKAGE_VERSION" -b build
+  mkdir -p /luetbuild/go/src/{{$host}}/{{$org}} && cd /luetbuild/go/src/{{$host}}/{{$org}} && \
+  git clone https://{{$host}}/{{$org}}/{{$repo}} && cd {{$repo}} && git checkout "$PACKAGE_VERSION" -b build
 {{end}}
--- a/packages/templates/golang.yaml
+++ b/packages/templates/golang.yaml
@@ -1,0 +1,22 @@
+{{ define "golang_deps" }}
+{{ if .Values.distribution }}
+{{if eq .Values.distribution "opensuse" }}
+- zypper in -y git {{- if .Values.upx -}} upx {{- end }}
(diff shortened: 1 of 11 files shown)


packages/templates/golang.yaml
```diff
@@ -0,0 +1,25 @@
...
+- zypper in -y git {{- if .Values.upx -}} upx {{- end }}
+{{else if eq .Values.distribution "fedora" }}
+- dnf install -y git {{- if .Values.upx -}} upx {{- end }}
+{{else if eq .Values.distribution "ubuntu" }}
+- apt-get install -y git {{- if .Values.upx -}} upx {{- end }}
+{{end}}
+{{end}}
+{{end}}
+
+{{ define "golang_download_packagev" }}
```
mudler, line 13: I really hate this. I need to find a way to make it general so it understands from the context that it needs to put a `v` upfront the version

mudler:
An unrelated issue now: 

```
#8 1.233 	github.com/tredoe/osutil@v1.0.5 requires
#8 1.233 	github.com/tredoe/fileutil@v1.0.0: reading github.com/tredoe/fileutil/go.mod at revision v1.0.0: unknown revision v1.0.0
```

going to update the modules :see_no_evil:

mudler:
Supersedes #503

packages/templates/golang.yaml
```diff
@@ -0,0 +1,22 @@
...
+{{end}}
+
+{{ define "golang_download_package" }}
+{{ $version:= get . "version" }}
+{{ $org:= get . "org" }}
+{{ $repo:= get . "repo" }}
+- |
+  PV="{{$version}}" && \
+  PACKAGE_VERSION=${PV%\+*} && \
+  mkdir -p /luetbuild/go/src/github.com/{{$org}} && cd /luetbuild/go/src/github.com/{{$org}} && \
```
Itxaka, line 20: can we make this more generic? like mayube it doesnt come from github so dest dir shuld not mention github anywhere if possible.

packages/templates/golang.yaml
```diff
@@ -0,0 +1,22 @@
...
+- apt-get install -y git {{- if .Values.upx -}} upx {{- end }}
+{{end}}
+{{end}}
+{{end}}
+
+{{ define "golang_download_package" }}
+{{ $version:= get . "version" }}
+{{ $org:= get . "org" }}
+{{ $repo:= get . "repo" }}
+- |
```
Itxaka, line 17: maybe we should add `{{ $host:= get . "host"| default "github.com" }}` in case we want to use that to generate the full repo below?

Itxaka (approved):
cool!

packages/templates/golang.yaml
mudler: good point! fixed in https://github.com/rancher-sandbox/cOS-toolkit/pull/493/commits/da26962f9c2eca70f1be20dc36395276ab5fa494 . let's see if CI tests are passing :)

Itxaka:
already building packages properly with the templates, I dont think we need to wait for it to finish the whole pipeline :+1:

PR #494, opened 2021-08-06
Split dracut default config to a separate package

mudler:
So users doesn't have to create a default dracut configuration file
unless they really need to.

Blocked by #489 
Signed-off-by: Ettore Di Giacinto <edigiacinto@suse.com>

Issue #231, opened 2021-06-01
cOS light flavor


mudler:
**Is your feature request related to a problem? Please describe.**
A smaller version of cOS would suit in different contexts. For example where a minimal set of feature are required, without taking advantage of the full stack (systemd+grub+dracut) we provide.

**Describe the solution you'd like**
a smaller and light cOS variant that could be used for more portable and embedded purposes

**Describe alternatives you've considered**

**Additional context**
Consider using a lighter stack, e.g. runit+musl+minimal kernel to reduce the final image size

mudler:
This seems even more doable now - the only challange I see here for providing e.g. an alpine variant is providing openrc scripts in place of systemd. I've played with it already a while ago and some WIP version of them are in https://github.com/rancher-sandbox/cOS-toolkit/tree/alpine/packages/cos-setup

PR #495, opened 2021-08-06
Split kernel-related packages

mudler:
Depends on #494 

This allows to consume the kernel and initrd separately. In context of #231 for
example could be tackled by reusing the kernel and (maybe) initrd
generated from other flavors.


mudler:
It actually supersedes #494 , so if tests are passing here should be good to go

Itxaka (approved):
you are on a roll today :D

PR #505, opened 2021-08-10
Upload ami id text file on release

### LINKED ISSUE #457: Add AMI images id list to releases artifact
Having a clear list of the published AMIs helps users to find the image rather have to dig with the aws-cli

An artifact which is uploaded during releasing which contains the AMI IDs

Itxaka:
Output the ami id to a file that then is uploaded to when publishing a
release we can download it and attach it to the release

Closes #457 

### DIFF
--- a/images/aws_upload.sh
+++ b/images/aws_upload.sh
@@ -64,6 +64,7 @@ echo "Making AMI public"
 aws ec2 modify-image-attribute --image-id "${ami_id}" --launch-permission "Add=[{Group=all}]"
 
 echo "AMI Created: ${ami_id}"
+echo "${ami_id},${AWS_DEFAULT_REGION}" >> ami_id.txt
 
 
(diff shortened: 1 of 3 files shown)


images/aws_upload.sh
```diff
@@ -64,6 +64,7 @@ echo "Making AMI public"
 aws ec2 modify-image-attribute --image-id "${ami_id}" --launch-permission "Add=[{Group=all}]"
 
 echo "AMI Created: ${ami_id}"
+echo "${ami_id}" >> ami_id.txt
```
mudler, line 67: mmh maybe we should include AMI's from other regions too?

Itxaka: ummm, yeah, that rigth, should probably be a list of `ami_id, region` instead of a simple ami id

PR #506, opened 2021-08-10
FIPS based toolchain

### LINKED ISSUE #491: FIPS-enabled golang builds

mudler:
Still WIP. I'm trying to move all the golang packages in a collection which is consuming this one, and still need to create a specific meta to pull all FIPS packages in

Fixes #491


mudlers:
Superseded by #513

PR #260, opened 2021-06-11
Bump toolchain/luet to 0.16.6

cOS-cibot:
documentation

PR #507, opened 2021-08-10
Fix #461 && unify upgrade images

### LINKED ISSUE #461: Delete all AMIs which are not 0.6.0 release
Delete all AMIs which are not tagged and versioned after the https://github.com/rancher-sandbox/cOS-toolkit/releases/tag/v0.6.0 release

Blocked by #459

### LINKED ISSUE #481: set LUET_PRIVILEGED_EXTRACT to false make our test suite to fail
**cos-toolkit version:**

**Describe the bug**
At the moment we have enabled privileged extraction: https://github.com/rancher-sandbox/cOS-toolkit/blob/05e150c923fc72c88a74f854c7d831067bdbe058/packages/installer/upgrade.sh#L165
Because it make our test suite to fail, see https://github.com/rancher-sandbox/cOS-toolkit/pull/260 for context.

**To Reproduce**
Disable `LUET_PRIVILEGED_EXTRACT`

**Expected behavior**
To pass our test suite

**Logs**

**Additional context**
https://github.com/rancher-sandbox/cOS-toolkit/pull/260
https://github.com/rancher-sandbox/cOS-toolkit/issues/480

Itxaka:
Looks like we were not creating the destination dir for the extract so it
was failing to upgrade. Removing the switch as its no longer necessary.

Unify upgrade images into a single test, extract image name and image
version into vars to be reused.

Closes #481 


packages/installer/upgrade.sh
```diff
@@ -165,7 +165,6 @@ upgrade() {
     # FIXME: Define default /var/tmp as tmpdir_base in default luet config file
     export XDG_RUNTIME_DIR=$temp_upgrade
     export TMPDIR=$temp_upgrade
-    export LUET_PRIVILEGED_EXTRACT=true
```
mudler, line 168: that looks more like a regression, when I've tested this it was not failing to upgrade at all - but it succeeded in the upgrade - the failing aspect was vagrant ssh not working with pubkeys. If now it works without pubkeys, doesn't necessarly mean its bug free in there :)

mudler: Can you try to check if after an upgrade you can still ssh with the vagrant pubkeys?

mudler: see here: https://github.com/rancher-sandbox/cOS-toolkit/pull/260#issuecomment-862213314 and https://github.com/rancher-sandbox/cOS-toolkit/pull/260#issuecomment-862216012

My test was simply mimicking CI:

- install
- do upgrade (which succeeded, and I could boot just fine afterwards)
- try to vagrant ssh with a priv/pubkey that was previously there already (setted up via cloud-init)

Itxaka: as mentioned on the issue we dont use vagrant keys AFAIK, when you do vagrant ssh it uses the user/pass in the vagrantfile

Maybe when it happened we were not using the ssh user/pass ?

Itxaka: Indeed, original issue had vagrant using its own method for the ssh keys while this PR https://github.com/rancher-sandbox/cOS-toolkit/commit/c3d5ba246358b42830e26ec10b627127d82c5ba0#diff-225f3d44285006fdfb0e6a444342ca7f84f2b8f366ae15fa4360265b535be386 changed that to user/pass to avoid spending time in injecting the keys....

Itxaka: FYI, this was failing without this patch still, so it should be valid. Doesn't fix #481 thought... :D

mudler: yes that's why I'm asking to test this with ssh keys :slightly_smiling_face:  the problem was not present by logging with user/pass, but it was there only while logging in via ssh keys...

Itxaka:
superseeded by https://github.com/rancher-sandbox/cOS-toolkit/pull/509

Issue #510, opened 2021-08-12
Failure in creating AWS images


mudler:
Seems that cos-deploy is not mounting the persistent partition, so it tries to consume `/` instead of `COS_PERSISTENT` as actually was meant to be.

**Logs**
```     amazon-ebs.cos:  Size: 195.7KiB
    amazon-ebs.cos: \   Repository cOS revision: 36 - 2021-08-11 1824 +0000 UTC
    amazon-ebs.cos: \  Repository: cos Priority: 1 Type: docker
    amazon-ebs.cos:  Packages that are going to be inst<sensitive>ed in the system:
    amazon-ebs.cos:   cos-system-0.6.3+4 (cos)
    amazon-ebs.cos: |  Downloading image quay.io/costoolkit/releases-green:cos-system-0.6.3-4
    amazon-ebs.cos: |  failed downloading image quay.io/costoolkit/releases-green:cos-system-0.6.3-4: failed listing images: mount c<sensitive>back failed on /usr/local/.cos-upgrade/tmp/upgrade/containerd-mount455333309: write /usr/local/.cos-upgrade/tmp/upgrade/containerd-mount455333309/boot/System.map-5.3.18-59.16-default: no space left on device
==> amazon-ebs.cos: panic: runtime error: invalid memory address or nil pointer dereference
==> amazon-ebs.cos: [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x192df52]
==> amazon-ebs.cos:
==> amazon-ebs.cos: goroutine 268 [running]:
==> amazon-ebs.cos: github.com/mudler/luet/pkg/compiler/types/artifact.(*Checksums).generateSum(0xc000027cf0, 0x0, 0x20f1ff8, 0xc0004b7d80, 0x1e1510d, 0x6, 0x0, 0x0)
==> amazon-ebs.cos: 	/luetbuild/go/src/github.com/mudler/luet/pkg/compiler/types/artifact/checksum.go:65 +0x52
==> amazon-ebs.cos: github.com/mudler/luet/pkg/compiler/types/artifact.(*Checksums).generateSHA256(0xc000027cf0, 0x0, 0xc0004d26c0, 0xc00070cd18)
==> amazon-ebs.cos: 	/luetbuild/go/src/github.com/mudler/luet/pkg/compiler/types/artifact/checksum.go:60 +0x88
[...]

PR #514, opened 2021-08-12
ci: Add test for deploying from raw image

Itxaka:
Recent events have shown that we do not test cos-deploy from a raw image
as well as we should.

This patch adds a simple test and preparation of a vm based of the raw
image so we can test the the default cos-deploy stanza works as expected
and lays the base so we can expand it with more tests in the future.

Unfortunately as this test is fully modifying the disk image we can only
test one thing per time, once we modify the disk the pristine state that
we started with is gone, so we should probably create one work per test
that we want.

This also means that we can drop the normal raw-image-test that only
checks if we can boot from the raw image properly as this new test already
checks that part.

### DIFF
--- a/make/Makefile.test
+++ b/make/Makefile.test
@@ -96,18 +96,27 @@ $(GINKGO):
 
 
 #
-# ------------ test for raw images  ------------
+# ------------ test for raw image  ------------
 #
 
 RAW?=$(shell ls $(ROOT_DIR)/*.raw 2> /dev/null)
 
-raw_disk_test: as_root $(VBOXMANAGE)
+raw_disk_test_deploy: create_vm_from_raw_image $(GINKGO)
+	# run tests
+	cd $(ROOT_DIR)/tests && $(GINKGO) $(GINKGO_ARGS) ./recovery-raw-disk
+
+create_vm_from_raw_image: as_root $(VBOXMANAGE)
 ifeq ("$(RAW)","")
 	@echo "Raw image does not exists, please run make raw_disk first"
 	@exit 1
 else
 	# transform the raw image
 	VBoxManage convertdd $(RAW) sda.vdi --format VDI
+	# increase disk size to 15Gb
+	VBoxManage modifymedium disk sda.vdi --compact --resize 16000
+	# creatre user-data iso (NoCloud)
+	cp packer/user-data/aws.yaml user-data
+	mkisofs -o cidata.iso -V CIDATA -J -r user-data
 	# create vm
 	VBoxManage createvm --name "test" --register
 	# increase memory, otherwise grub fails to boot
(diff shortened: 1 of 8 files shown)


make/Makefile.test
```diff
@@ -96,18 +96,27 @@ $(GINKGO):
...
+
+create_vm_from_raw_image: as_root $(VBOXMANAGE)
 ifeq ("$(RAW)","")
 	@echo "Raw image does not exists, please run make raw_disk first"
 	@exit 1
 else
 	# transform the raw image
 	VBoxManage convertdd $(RAW) sda.vdi --format VDI
+	# increase disk size to 15Gb
+	VBoxManage modifymedium disk sda.vdi --compact --resize 16000
```
mudlers, line 116: that bugs me a bit, aren't we going to test an image which is a bit larger than the one we provide?

mudlers (approved):
very cool! :+1:  looks good here

make/Makefile.test
```diff
@@ -118,16 +127,18 @@ else
 	VBoxManage storagectl "test" --name "sata controller" --add sata
 	# attach the disk to the machine
 	VBoxManage storageattach "test" --storagectl "sata controller" --port 0 --device 0 --type hdd --medium $(ROOT_DIR)/sda.vdi
+	# attach the user-data iso
+	VBoxManage storageattach "test" --storagectl "sata controller" --port 1 --device 0 --type dvddrive --medium $(ROOT_DIR)/cidata.iso
 	# startvm and detach
 	VBoxManage startvm "test" --type headless
 	# sleep for a bit to let the recovery boot
 	sleep 300
```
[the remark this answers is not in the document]
Itxaka: Good point, there is no need already for this!

make/Makefile.test
Itxaka: We are resizing the image otherwise the user-data cannot partition it correctly (lack of space) this is to be on sync with what we do on cloud operators (resize image on launch)

Itxaka:
gonna wait to change the sleep because this should fail the test due to #510 (this was branched before that was merged)

After, Im gonna rebase and change the sleep so ti should pass, that way we can test this as well :D

mudler:
@Itxaka due to #533 it might fail to succeed, I'd merge this nevertheless so we can have some tests for `cos-deploy`

Itxaka:
As long as it finally works and runs the test without failing, it's ok for me. But currently it doesn't so....😔

mudler:
It just fails for #533 now, I wouldn't wait for the other tests to pass at this point :)

```
         time="2021-08-20T08:23:58Z" level=info msg="Finished yip file execution" stage=after-deploy.after stages=0 success=true
        tune2fs 1.43.8 (1-Jan-2018)
        tune2fs 1.43.8 (1-Jan-2018)
        Grub config file not found
        : Process exited with status 1
    occurred
```

PR #517, opened 2021-08-13
Add STRICT_MODE

### LINKED ISSUE #516: Improve UX of installation/upgrade/reset cloud-init stages
The output of execution of the stages is hidden, and the error is not checked. The stages are running in a very delicate phase - such upgrades, install and reset. On one hand we don't want to bail out in case of failures - those stages are not really mandatory to complete any of those operation successfully, but we do want to offer anyhow a way to be more strict. 

https://github.com/rancher-sandbox/cOS-toolkit/blob/0d402c163aa2e1b6c36ef23ff928c1392579dba8/packages/installer/installer.sh#L387 the stages executed after/before installation and upgrade are not showing output or reporting any error. 

A flag, or an env var which enables a more "strict" mode and displays output and fails over if necessary. Ideally an env var is a good choice, because we could re-use it for a general solution with #224 . By default would make sense to have that enabled, with a way to optionally turn it off.

N/A

**Additional context**
https://github.com/harvester/harvester-installer/pull/93#discussion_r688253302

mudlers:
This allows to fail in case the cloud-init stages are failing to execute. Also introduce `/etc/cos/config` for a general config place where to place customizations

Fixes #516

### DIFF
--- a/packages/cos/collection.yaml
+++ b/packages/cos/collection.yaml
@@ -1,7 +1,7 @@
 packages:
   - name: "cos"
     category: "system"
-    version: "0.6.4"
+    version: "0.6.5"
     description: "cOS base image, used to build cOS live ISOs"
     brand_name: "cOS"
     labels:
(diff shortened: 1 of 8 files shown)


mudler:
Superseded by #518

PR #522, opened 2021-08-17
Allow to install a custom ISO in the installer script

### LINKED ISSUE #515: Add ISO downloader to cos-installer
It might be handy in different occasions to have a parameter on cos-installer which downloads the ISO and consumes that one instead of the squashfs we are actually booting

A new flag to download an iso in cos-installer and consuming that one during install instead of using the hardcoded location

**Additional context**
https://github.com/harvester/harvester-installer/pull/93
https://github.com/harvester/harvester-installer/pull/93/commits/77409ab4b95dff7bd0d7318c6b403d7f82fa2b33

mudlers:
Fixes #515

### DIFF
--- a/packages/cos/collection.yaml
+++ b/packages/cos/collection.yaml
@@ -1,7 +1,7 @@
 packages:
   - name: "cos"
     category: "system"
-    version: 0.6.5+1
+    version: 0.6.5+2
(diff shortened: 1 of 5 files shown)


davidcassany (approved):
LGTM

Issue #532, opened 2021-08-19
Documentation: /etc/cos/config


mudler:
`/etc/cos/config` has been introduced in https://github.com/rancher-sandbox/cOS-toolkit/pull/517 as a single point of configuration.

Documentation and official examples should be updated as necessary, along with a full example of the config file, and how to use it.

See also https://github.com/mudler/c3os/blob/master/files/etc/cos/config for an example

Issue #224, opened 2021-05-28
Mechanism for derivatives branding and customization


mudler:
At the moment is possible to configure the derivative's setting (such as upgrade channels, grub, etc ) via oem file, but there is no common entrypoint for configuration, nor a consistent way to do so

**Is your feature request related to a problem? Please describe.**
Configuring GRUB entries, LIVECD labels, os-release informations, and such are all needed to be applied with OEM or either overriding the values file, but for the pieces that can be tweaked in runtime there should be a common way, like we do for e.g. with channel upgrades (we have a `/etc/cos-upgrade-image` file which can used to tweak `cos-upgrade` behavior)

**Describe the solution you'd like**

Extend  `cos-rebrand`  to cover all the branding use-cases, and consistently use `/etc/cos/config` for the other customization aspects, such as: default upgrade image, default installation partition size,  default image (active/passive/recovery) size, etc.

**Describe alternatives you've considered**
N/A

**Additional context**

Related issues:
- [x] #150 
- [x] #160 
- [x] #565
- [x] #821 
- [x] #238 
- [x] #154 
- [x] #516 
- [x] #543

mudler:
A general `/etc/cos/config` has been introduced in #517

PR #543, opened 2021-08-23
Config fixups

mudler:
While I was working on #532 I noticed that the config files currently in `cos-upgrade` could actually override what is passed explicitly by arguments.

This change also makes the whole more consistent and also allows to use only `/etc/os-release` as a single source of truth.

Relates to #224


packages/installer/upgrade.sh
```diff
@@ -54,10 +54,23 @@ find_recovery() {
...
+    if [ -e "/etc/os-release" ]; then
+        source /etc/os-release
+    fi
+
     if [ -e "/etc/cos-upgrade-image" ]; then
         source /etc/cos-upgrade-image
     fi
 
+    if [ -e "/etc/cos/config" ]; then
+        source /etc/cos/config
```
davidcassany, line 71: Isn't it still overwriting what was passed in the command line? `find_upgrade_channel` happens right after evaluating the command line parameters. If the command line includes `--recovery` and `/etc/cos/config` includes `UPGRADE_RECOVERY=false` the recovery flag turns to be useless. I am unsure of what is expected to be in `/etc/cos/config` though

davidcassany (commented):
Beyond the additional loaded files I am not sure I see a real behavior change.

packages/installer/upgrade.sh
mudler: The image that is being passed from the CLI is being set right afterwards here: https://github.com/rancher-sandbox/cOS-toolkit/blob/f272f6e25e9121ca5a808ee45dd9f8982738bce8/packages/installer/upgrade.sh#L65

UPGRADE_RECOVERY, even if could be set here it shouldn't be part of the config, it's just to drive the CLI selection

davidcassany: OH I see, this is for the image parameter and the config associated with the upgrade image. All clear now, makes complete sense.

davidcassany (approved):
LGTM

kkaempf:
There should be a documentation PR explaining these files and changes.

mudler:
> There should be a documentation PR explaining these files and changes.

post-mortems?! https://github.com/rancher-sandbox/cos-toolkit-docs/pull/24 :)

PR #544, opened 2021-08-23
Implement support for persistent state a area

### LINKED ISSUE #430: Handle persistency layer in immutable rootfs dracut module

Configs like the following are likely to be required in most of use cases and it is curmbersome, error prone and likely to break the upgrades/downgrades (fallback).

```yaml
   initramfs:
     - if: '[ ! -f "/run/cos/recovery_mode" ]'
       name: "Persist"
       commands:
       - |
            target=/usr/local/.cos-state

            # Always want the latest update of systemd conf from the image
            # TODO: This might break the fallback system
            mkdir -p "${target}/etc/systemd/"
            rsync -av /etc/systemd/ "${target}/etc/systemd/"

            # Only populate ssh conf once
            if [ ! -e "${target}/etc/ssh" ]; then
              mkdir -p "${target}/etc/ssh/"
              rsync -av /etc/ssh/ "${target}/etc/ssh/"
            fi

            # undo /home /opt /root mount from cos immutable-rootfs module
            # TODO: we could think of configuring custom overlay paths in
            # immutable rootfs package. So this part could be omitted
            for i in home opt root; do
              sed -i "/overlay \/${i} /d" /etc/fstab
              nsenter -m -t 1 -- umount "/sysroot/${i}"
            done

            # setup directories as persistent
            # TODO: would it make sense defining persistent state overlayfs mounts
            # as part of the immutable rootfs config?
            for i in root opt home var/lib/rancher var/lib/kubelet etc/systemd etc/rancher etc/ssh; do
              mkdir -p "${target}/${i}" "/${i}"
              echo "${target}/${i} /${i} none defaults,bind 0 0" >> /etc/fstab
              nsenter -m -t 1 -- mount -o defaults,bind "/sysroot${target}/${i}" "/sysroot/${i}"
            done
```

Immutable rootfs dracut module is capable to configure some persistent state layer, so the whole logic above could be converted into some simple config file applied at `rootfs` stage. By default the module should not do anything like that, only on demand, so the current behavior stays and the chance to cook your own soup regarding the persistent layer is still there.

Also we could allow the possibility to distinguish between fallback and active systems state if desired to prevent upgrade incompatibilities.

davidcassany:
This commit adds the concept of a persistent state area in
immutable-rootfs package. It adds support for adding an additional
non ephemereal overlayfs on top of configurable locations,
alternatively, instead of using overlayfs in can simple bind mount
a mirrored location from a persistent area.

Fixes #430

### DIFF
--- a/packages/cos/collection.yaml
+++ b/packages/cos/collection.yaml
@@ -1,7 +1,7 @@
 packages:
   - name: "cos"
     category: "system"
-    version: "0.6.6"
+    version: "0.6.7"
     description: "cOS base image, used to build cOS live ISOs"
     brand_name: "cOS"
     labels:
@@ -9,7 +9,7 @@ packages:
       autobump.revbump_related: "recovery/cos-img recovery/cos-squash"
   - name: "cos"
     category: "recovery"
-    version: "0.6.6"
+    version: "0.6.7"
     brand_name: "cOS recovery"
     description: "cOS recovery image, used to boot cOS for troubleshooting"
     labels:
@@ -17,7 +17,7 @@ packages:
       autobump.revbump_related: "recovery/cos-img recovery/cos-squash"
   - name: "cos-container"
     category: "system"
-    version: "0.6.6"
+    version: "0.6.7"
     brand_name: "cOS"
     description: "cOS container image, used to build cOS derivatives from scratch"
     labels:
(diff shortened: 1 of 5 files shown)


mudler (approved):
Looks good here! shall we also update the docs to specify how to use it? (we can also start to replace it in our samples with the cloud-init mechanism we have spread all over already..)

davidcassany:
> Looks good here! shall we also update the docs to specify how to use it? (we can also start to replace it in our samples with the cloud-init mechanism we have spread all over already..)

Sure I am already working on documenting it. Then I was also thinking on updating the examples we have.

PR #545, opened 2021-08-23
Update example config with inline documentation

mudler:
I'm going to reuse the example by linking it in the docs directly to avoid dups


davidcassany (approved):
LGTM

Issue #525, opened 2021-08-18
Remove hardcoded arches


dragonchaser:
Remove all hardcoded arch definitions.
e.g.

- [x] https://github.com/rancher-sandbox/cOS-toolkit/blob/master/scripts/get_luet.sh#L17
- [ ] https://github.com/rancher-sandbox/cOS-toolkit/blob/9ff8323c3ec981b264d33874433829812333bb75/.github/build.yaml.gomplate#L35
- [x] https://github.com/rancher-sandbox/cOS-toolkit/blob/5e4b098f73daeee96e7a50c0f47b4f099fe5ec11/packages/cos-features/features/carrier.yaml#L32
- [x] https://github.com/rancher-sandbox/cOS-toolkit/blob/9af8b3f4e714d4e4d8299ac073d62f8aacbc9566/.github/workflows/build-pr.yaml#L247
- [ ] ...

mudlers:
Re: the values file, there should be a value file for each specific arch we intend to support (green-arm, blue-arm, etc)

Itxaka:
As part of the arm64 story, all of the hardcoded values here were moved to a var linked to the current arch.

PR #549, opened 2021-08-25
Fix uname arg to get arch

Ludea:


mudler:
Thanks @Ludea for opening a PR!

@dragonchaser can you have a look as it is related to #525 ?

mudler:
@Ludea can you please also sign-off the commits? otherwise the checks will automatically fail, https://github.com/rancher-sandbox/cOS-toolkit/pull/549/checks?check_run_id=3421417597

PR #551, opened 2021-08-25
ci: add build test for examples dir 

Itxaka:


Itxaka:
Maybe partial of #144 ? We could use this as base to clone out that repo, change references to the current commit and rebuild...maybe

.github/workflows/build-pr.yaml
```diff
@@ -1,47 +1,4 @@
-
-
-
-
```
mudler, line 4: Looks like an unwanted effect?

mudler (changes requested):
seems the workflows were removed :man_shrugging:

.github/workflows/build-pr.yaml
mudler: ah my bad. It looked like the file was completely removed :smile:

mudler (approved):
Looks good, it looked like at first glance the files were removed.. :smile:

.github/pipelines.sh
```diff
@@ -3,4 +3,5 @@ set -ex
 
 for i in $(ls config); do 
     gomplate --left-delim "{{{" --right-delim "}}}" -V --datasource config=config/$i --file build.yaml.gomplate > workflows/build-$i
+    sed -i '/^[[:space:]]*$/d' workflows/build-$i
```
mudler, line 6: I think mostly comes from templates block not having `{{-` prepended, so this shouldn't be really required.. but if it works fine by me :)

PR #553, opened 2021-08-27
Update carrier to epinio

Ludea:


kkaempf:
[Epinio v0.0.23](https://github.com/epinio/epinio/releases/tag/v0.0.23) was released yesterday :wink:

mudler:
Thanks @Ludea for looking at this!

Although since `epinio` and `k3s` are not strictly `cOS` features we decided to drop them, as they were meant only as sample and for testing purposes.

We are closing this PR in favor of https://github.com/rancher-sandbox/cOS-toolkit/pull/554 and by documenting #555 how derivatives can ship their own features like those ones that can be enabled in runtime, with examples along.

PR #580, opened 2021-09-01
Live mode

mudler:
Docs: https://github.com/rancher-sandbox/cos-toolkit-docs/pull/29

### DIFF
--- a/packages/cloud-config/definition.yaml
+++ b/packages/cloud-config/definition.yaml
@@ -1,3 +1,3 @@
 name: cloud-config
 category: system
-version: 0.8+3
+version: "0.9"
--- a/packages/cloud-config/oem/07_live.yaml
+++ b/packages/cloud-config/oem/07_live.yaml
@@ -1,0 +1,19 @@
+# Default cOS OEM configuration file
+#
+# This file is part of cOS and will get reset during upgrades.
+#
+# Before you change this file manually,
+# consider copying this file to /usr/local/cloud-config or
+# copy the file with a prefix starting by 90, e.g. /oem/91_custom.yaml
+name: "LiveCD Detection"
+stages:
+   rootfs.before:
+     - if: |
+            cat /proc/cmdline | grep -q "CDLABEL" || cat /proc/cmdline | grep -q "rd.cos.disable"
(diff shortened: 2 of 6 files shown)


kkaempf (commented):
Where do you test existance of `/run/cos/live_mode` and its content ?

mudler:
> Where do you test existance of `/run/cos/live_mode` and its content ?

we don't, I've tested this manually as we don't have tests running on livecd

kkaempf:
Hmm, livecd testing probably cannot be automated with github workers 🤔 
Then we should have a test document for manual testing where the sentinel file is checked.

mudler:
> Hmm, livecd testing probably cannot be automated with github workers thinking
> Then we should have a test document for manual testing where the sentinel file is checked.

We don't have a test document, or I am missing something?

Itxaka:
> Hmm, livecd testing probably cannot be automated with github workers


Why not? We are already testing recovery images which are also "live". The sky is the limit!

mudler:
I've created https://github.com/rancher-sandbox/cOS-toolkit/issues/581 with what I had in mind now, but pretty sure the list isn't exhaustive at all :slightly_smiling_face:

mudler:
> > Hmm, livecd testing probably cannot be automated with github workers
> 
> Why not? We are already testing recovery images which are also "live". The sky is the limit!

the limit is also how many concurrent job that uses virtualization we can have...

Itxaka (approved):

Itxaka:
> the limit is also how many concurrent job that uses virtualization we can have...
@kkaempf is gonna get us a few of those AMD Epyc servers, so that is fixed already :P

kkaempf:
> @kkaempf is gonna get us a few of those AMD Epyc servers, so that is fixed already :P

You provide the cooling, I provide the servers. Deal ?

Itxaka:
> You provide the cooling, I provide the servers. Deal ?

Send them my way, I'll store them at home. Did you said that they had nice GPUs? BRB, looking for the GPU requirements to run cyperpunk 2077...

Itxaka (approved):

PR #557, opened 2021-08-27
Use grub2 in ISOs

davidcassany:
This commit refactors grub2 usage in cOS. It adds a collection of
packages related to grub2 and configures ISOs generation to make
use of grub2 packages.

Fixes #521

Signed-off-by: David Cassany <dcassany@suse.com>

PR #584, opened 2021-09-02
ci: Reintroduce the check on raw test vm boot

Itxaka:
Lets wait after launching the vm for testing until we see the welcome
message

### DIFF
--- a/make/Makefile.test
+++ b/make/Makefile.test
@@ -131,6 +131,10 @@ else
 	VBoxManage storageattach "test" --storagectl "sata controller" --port 1 --device 0 --type dvddrive --medium $(ROOT_DIR)/cidata.iso
 	# startvm and detach
 	VBoxManage startvm "test" --type headless
+	# sleep for a bit to let the recovery boot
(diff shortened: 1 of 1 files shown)


mudler (approved):
Even if hardcoded sleeps are unpleasant, I'd merge this as-is. Better have stable tests than flaky ones

davidcassany:
I'd wait for my PR (https://github.com/rancher-sandbox/cOS-toolkit/pull/557) to finalize before merging this if you don't mind. There I did a couple of changes to speed up boot. Default SATA ports number is 30 and this has a huge impact on boot and causes at least a failure on initial `OEM` mount (even this should not hurt it might have impact on boot time).

Itxaka:
oh I see, so if you dont specify it just  creates 30 ports?? Wow.

mudler:
> oh I see, so if you dont specify it just creates 30 ports?? Wow.

you never know right.. :rofl:

Itxaka:
closed in favour of #591

PR #590, opened 2021-09-02
Limit SATA ports to speed up boot

davidcassany:


Itxaka:
I dont think we need to wait for tests in order to merge this, one nit only, I would like to keep the reboot in place in order to check that cos-deploy from raw image still works, as that was tested before. with #591 the reboot should not be a problem anymore.

davidcassany:
> I dont think we need to wait for tests in order to merge this, one nit only, I would like to keep the reboot in place in order to check that cos-deploy from raw image still works, as that was tested before. with #591 the reboot should not be a problem anymore.

I don't see the problem here, I am not deleting the reboot, I am just deleting the grub setting to Active. I just removed that because this is already the default. So if it does not boot to Active this should be considered a bug.

PR #591, opened 2021-09-02
ci: use time param for reboot

Itxaka:
Currently we had the reboot function on SUT with a hardcoded timeout
which was hitting its limits on the raw disk test

This modifies the reboot function so it acceps a timeout and passes it
to the EventuallyConnects function


davidcassany (approved):
Looks good to me.

davidcassany:
According to the serial logs I checked a while ago on some failed tests I saw that first boot took around 250 seconds and it was failing due to a timeout on reboot. So increasing the reboot time might definitely help to fix the issue, even though more 180s for a single reboot without having to install anything at boot time is close to the unacceptable area, but we can analyze that later on.

davidcassany:
@mudler the raw image deploy test passed. I'd say there is no need to wait for all the tests, if possible I'd force merge this PR as it has potential to unblock others.

PR #596, opened 2021-09-03
makefile: check yq version for package list

Itxaka:
Because yq version 3 and 4 differ in the cli arguments if you are using
version 3, the packages list will be empty and everything will be built.

Instead have a small check in the makefile to run the proper yq command
based on the major yq version on build


mudler (approved):
nice!

dragonchaser (approved):

PR #606, opened 2021-09-06
grub-artifacts: add provides to respect old packages

### LINKED ISSUE #604: grub-config package was renamed
**cos-toolkit version:**
0.6.7

**Describe the bug**
grub2-config was introduced in place of grub-config, causing breakage of `grub-config` consumers. We should have added `grub-config` to the `grub2-config` provides in order to handle a smooth switching without any breakage

**To Reproduce**
`luet install -y system/grub-config`

**Expected behavior**
luet to correctly redirect to grub2-config instead

**Additional context**
#602

Itxaka:
When moving to the new name of grub2-config, the old package grub-config
is no longer present which will cause breakage on grub-config consumers.

This just adds the provides stanza so consumers will get redirected to
the newer grub2-config package

Closes #604

### DIFF
--- a/packages/grub2/collection.yaml
+++ b/packages/grub2/collection.yaml
@@ -4,4 +4,7 @@ packages:
     version: 0.0.2+1
   - name: "grub2-config"
     category: "system"
-    version: 0.0.11+1
+    version: 0.0.11+2
+    provides:
+    - name: "grub-config"
+      category: "system"


Itxaka:
Requires https://github.com/mudler/luet/pull/252 and a new luet release to work properly, but it can be merged already.

mudler (approved):

PR #607, opened 2021-09-06
ci: change triggers for github_runner job

Itxaka:
Currently there is no point in running it on PRs, especially because we
cant access the proper secrets for the workflow to work.

It also doesnt make sense to run it on each master push, instead just
run it when something has changed in the job files and its a push to
master


mudler (approved):
:+1: No need to wait for CI I guess! :)

PR #608, opened 2021-09-06
Support image builder script in all flavors

davidcassany:

### DIFF
--- a/images/img-builder.sh
+++ b/images/img-builder.sh
@@ -1,9 +1,10 @@
 #!/bin/bash 
 
 # This is PoC for building images without requiring admin capabilities (CAP_SYS_ADMIN)
-FINAL_REPO="${FINAL_REPO:-quay.io/costoolkit/releases-green}"
+FLAVOR="${1:-green}"
+FINAL_REPO="${FINAL_REPO:-quay.io/costoolkit/releases-${FLAVOR}}"
 
-rm -rf ./*.part disk.raw grub_efi.cfg recovery root .luet.yaml oem
+rm -rf ./*.part disk.raw grub_efi.cfg recovery root .luet.yaml oem efi
 
 set -e
 
@@ -25,6 +26,7 @@ repositories:
 EOF
 
 # Create root-tree for COS_RECOVERY
+luet install --system-target efi -y system/grub2-efi-image
 luet install --system-target root -y system/grub2-config
 luet install --system-target root/grub2 -y system/grub2-artifacts
 luet install --system-target root/cOS -y recovery/cos-img
@@ -37,12 +39,7 @@ mkfs.ext2 -L COS_RECOVERY -d root rootfs.part
 truncate -s $((20*1024*1024)) efi.part
 
 mkfs.fat -F16 -n EFI efi.part
-mmd -i efi.part ::EFI
-mmd -i efi.part ::EFI/BOOT
-mcopy -i efi.part root/grub2/x86_64-efi/grub.efi ::EFI/BOOT/bootx64.efi
-
-# Copy grub.cfg in EFI partition
-mcopy -i efi.part root/etc/cos/grub_efi.cfg ::EFI/BOOT/grub.cfg
+mcopy -s -i efi.part efi/EFI ::EFI
 
 # Create the grubenv forcing first boot to be on recovery system
 mkdir -p oem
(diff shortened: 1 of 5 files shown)


Itxaka:
What would be the advantage of this? Is just using a different efi boot loader from the different flavors? After all the recovery should be the same on all 3 no?

Itxaka:
like are the files different? I can see the modules part which seems very nice to add/remove support for stuff (filesystems maybe?) but are the efi files different?

Itxaka (approved):

packages/grub2/collection.yaml
```diff
@@ -1,7 +1,21 @@
...
+  provides:
+  - name: "grub-artifacts"
+    version: ">0.0.2"
+
+- name: "grub2-config"
+  category: "system"
+  version: "0.0.13"
+  provides:
+  - name: "grub-config"
+    version: ">0.0.12"
```
Itxaka, line 14: I think this will clash if anyone is consuming grub-config 0.0.10 for example no? should we just `version: ">=0"` and call it a day?

davidcassany: I am not so sure, since this is 0.0.13 and 0.0.10 was provided by another package definition, most likely providing, the same, but still a different package. So if someone call for grub-config 0.0.10 probably I prefer to make explicit they require the old build.

davidcassany:
> What would be the advantage of this? Is just using a different efi boot loader from the different flavors? After all the recovery should be the same on all 3 no?

The advantage of this is the `image/img-builder.sh` script now it won't fail if you use any of the three flavors. Previously it was implicit the use of `green` flavor. So this prepares the EFI binaries within the grub2-efi-image package and the script just applies it. It is not that we need to provide image builder for all flavors, this just makes the script generic.

davidcassany:
> like are the files different? I can see the modules part which seems very nice to add/remove support for stuff (filesystems maybe?) but are the efi files different?

yes they are, the pre-built binary for opensuse seams to be enough to boot. However I could not find equivalent ones in Ubuntu or Fedora. Moreover there are little differences between the paths (`/usr/lib/grub` vs `/usr/share/grub2`) and the binary names (`grub2-mkimage` vs `grub-mkimage`). It is so damn annoying how grub package differs across distros.

packages/grub2/build.yaml
```diff
@@ -16,13 +16,30 @@ steps:
...
+{{end}}
+
 {{if eq .Values.name "grub2-artifacts"}}
 - mkdir /grub-artifacts
   {{if .Values.distribution}}
     {{if eq .Values.distribution "opensuse"}}
 - cp -rf /usr/share/grub2/* /grub-artifacts
     {{else}}
-- cp -rf /usr/share/grub/* /grub-artifacts
+- cp -rfL /usr/lib/grub/* /grub-artifacts
```
davidcassany, line 42: @mudler I had to include `-L` because in Ubuntu there is symlink here under this path and the when installing/extracting the resulting package at install time luet was complaining of a broken symlink. Is this the intended behavior? If so I'd say it should fail at build time rather than install time.

Itxaka (approved):

packages/grub2/build.yaml
mudler: If the package ships a symlink, then luet will recreate a symlink when running `luet install`, but if it points to a path that doesn't exist then it shows a warning, but it will nevertheless recreate it.

So it really depends on the wanted behavior, luet will just try to reconstruct what it finds out

images/img-builder.sh
```diff
@@ -25,6 +26,7 @@ repositories:
 EOF
 
 # Create root-tree for COS_RECOVERY
+luet install --system-target efi -y system/grub2-efi-image
```
mudler, line 29: I'd really like to see at this point those packages being sourced in the `manifest.yaml` file in a specific field. So it's really easy to switch off default recovery packages, grub artifacts and configs in case of need

packages/grub2/build.yaml
davidcassany: well it did print stuff in stdout or stderr, but return code was not 0, so the builder script was failing, I did not actually verify if the installation happened or not. I guess it did.

images/img-builder.sh
davidcassany: Well I am not sure if it will make the manifest a bit confusing, since these will be four new sub sections of packages. So I'd list them in manifest with something like:

```yaml
packages:                                                     
  rootfs:                                                                       
  - system/cos                                                                  
  uefi:                                                                         
  - live/grub2-efi-image                                                        
  isoimage:                                                                     
  - live/grub2                                                                  
  - live/grub2-efi-image                                                        
  - recovery/cos-img 

disk_packages:
  efi:
  - system/grub2
  grub2:
  - system/grub2-artifacts
  system-img:
  - recovery/cos-img
```

All in different sections since they all have a different `--system-target`.  Or do you have something different in mind?

mudler: I see, yep indeed `packages` should be `iso_packages` at this point to avoid any confusion. But I think for the time being it's fine also if we just annotate in comments along the lines to explain what are those sections about.

Yup that would make sense, but I would not block this PR from being merged at this point as it looks outside the scope of this PR and can be iterated over in a second round of changes :+1:

mudler (approved):

images/img-builder.sh
davidcassany: Agree, I just created a follow up issue so we do not forget.

PR #614, opened 2021-09-07
ci: Add arch to CI builds

Itxaka:
This adds an arch value to the ci templates in order to differentiate
between the current amd64 and future arches that we may add.

On top of that it expands the ci templates with the labels value which
would add an extra step to the workflow in order to check for the given
labels in the PR and fail the workflow otherwise (Useful for arches,
do-not-test, etc...)

It also adds an arm64 workflow that runs on self runners for PRs labeled
with the arm64 label. Currently a separated one as its experimental..and
we dont have arm54 runners yet :D


.github/build.yaml.gomplate
```diff
@@ -672,6 +672,31 @@
...
+{{{end}}}
+
+{{{define "local_runner_deps"}}}
+{{{ $config := (datasource "config") }}}
+    {{{- if $config.local_runner }}}
+  local-runner-deps:
+    runs-on: [{{{ join $config.runs_on ", "}}}]
+    steps:
+      - run: |
+          apt-get install sudo
```
mudler, line 696: shouldn't those changes be part of the Docker image instead?

Itxaka: probably yes, but currently is faster to test it like this :D

mudler: I just thought that it would be impossible to apt-get something from the runner user xD my bad

dragonchaser: I have added sudo to the Dockerfile see PR #615

Itxaka: > I just thought that it would be impossible to apt-get something from the runner user xD my bad

it shouldnt but who knows :D

it fails so....

mudler:
just a reminder, we would need to have for both arm64 and armv7 (32bit hfp)

.github/build.yaml.gomplate
```diff
@@ -95,12 +109,22 @@
   {{{ $flavor := . }}}
   build-{{{ $flavor }}}:
     {{{ tmpl.Exec "runner" }}}
+    {{{ tmpl.Exec "needs" }}}
     env:
+      LUET_ARCH: {{{ $config.arch }}}
       FLAVOR: {{{ $flavor }}}
-      FINAL_REPO: {{{$config.organization}}}/{{{$config.repository}}}-{{{ $flavor }}}
+      ARCH: {{{ $config.arch }}}
+      FINAL_REPO: {{{$config.organization}}}/{{{$config.repository}}}-{{{ $flavor }}}{{{- if ne $config.arch "x86_64"}}}-{{{$config.arch}}}{{{end}}}
```
mudler, line 117: :+1:

.github/build.yaml.gomplate
```diff
@@ -95,12 +109,22 @@
...
       DOWNLOAD_METADATA: false
       PUSH_CACHE: {{{$config.push_cache}}}
-      REPO_CACHE: {{{$config.organization}}}/{{{$config.cache_repository}}}-{{{ $flavor }}}-cache
+      REPO_CACHE: {{{$config.organization}}}/{{{$config.cache_repository}}}-{{{ $flavor }}}-cache{{{- if ne $config.arch "x86_64"}}}-{{{$config.arch}}}{{{end}}}
+      {{{- if has $config "luet_install_from_cos_repo" }}}
+      LUET_INSTALL_FROM_COS_REPO: {{{ $config.luet_install_from_cos_repo }}}
+      {{{- end }}}
+      {{{- if ne $config.arch "x86_64"}}}
+      # For non x86_64 archs, skip the *-fips build as there are no golang releases in other arches
+      SKIP_PACKAGES: "build/golang-fips toolchain-fips/yip toolchain-fips/luet-makeiso toolchain-fips/luet live/grub2 live/syslinux live/systemd-boot live/boot"
```
mudler, line 126: A thought for a later iteration, maybe shall we include that as part of the `manifest.yaml` file as `skip_packages` for each arch? In that way we can control those more easily rather than relying on the CI

.github/build.go
```diff
@@ -118,6 +118,21 @@ func main() {
 		fmt.Println("-", m.String())
 	}
 
+	if os.Getenv("SKIP_PACKAGES") != "" {
+		for _,skip := range strings.Split(os.Getenv("SKIP_PACKAGES"), " ") {
+			for index, pkg := range missingPackages {
+				name := fmt.Sprintf("%s/%s", pkg.Category, pkg.Name)
+				if name == skip {
+					fmt.Println("- Skipping build of package due to SKIP_PACKAGES: ", pkg.String())
+					// how absurd is this just to pop one element from a slice ¬_¬
```
mudler, line 127: Just a nit on the style, so feel free to ignore it. It is more common instead to recreate a new slice and omitting the ones that should be skipped while repopulating it. There is a bit of payoff on memory, but it is more readable

Itxaka: Well, I only see a problem with that approach.

I dont know how to do it any other way :rofl:

.github/build.yaml.gomplate
Itxaka: yes, that is a good point. I might send an extra PR afterwards to move this there.

Itxaka:
> That's nice! It looks good here, just few nitpicks but shouldn't block from merging. I also noticed armv7 is not present, I hope won't be much trouble running armv7 builds on the aarch64 runner


Yeah, this thing grew a bit so I didnt want to start including armv7 on top, especially because it will require more profound changes on the templates probably to deal with the arches growing, so I wanted to keep it "simple" LMAO

.github/build.go
mudler: That's not really _so_ important, but here it is a small example: https://play.golang.org/p/2L2WPk9kR6E

Itxaka:
All Green!

PR #619, opened 2021-09-09
ci: Only run green flavor on master arm64

Itxaka:


mudler (approved):

Itxaka:
FFFFFFFFFF

We should remove the actual non-green values from there. We are only releasing the green stuff so no point in having those values in there

mudler:
> We should remove the actual non-green values from there. We are only releasing the green stuff so no point in having those values in there

On the long run it still makes sense imho to provide packages for blue and orange as well, so we keep a door open to build derivatives for those flavors. But if that's a problem for now, we can create follow-ups when needed

mudler:
I'd say to not wait for CI here, we will catch more issues when merging to master :)

PR #622, opened 2021-09-09
arm64: disable manifest creation

Itxaka:
Currently no version of luet-mtree is published for arm64 so ci cant
install.

Manually removed so we can regenerate the workflows afterwards


Itxaka:
@mudler force-pushing this, job was green on the build part

PR #624, opened 2021-09-09
ci: dont install from repo on arm64

Itxaka:


Itxaka:
should affect only master jobs, merging...

PR #626, opened 2021-09-10
ci: publish job should not use luet_arch

Itxaka:
Its running on ubuntu so it should default to x86_64


Itxaka:
I promise this is the last one @mudler

PR #628, opened 2021-09-10
ci: support more than one arch per config

Itxaka:
Instead of having separated configs, lets be more intelligent and define
the values per arch in the same config files


mudler (approved):
Nice! :tada:

PR #629, opened 2021-09-10
ci: enable manifest and luet repo for arm64

Itxaka:
Now that we have published the artifacts, we can rely again in the cos
repo to install luet and packages in the system with the side effect of
being able to enable the manifest generation for arm64


Itxaka:
only arm64 job should pass for this PR

Itxaka:
nice, we got all packages from the releases-green-arm64 repo :D

mudler (approved):

PR #633, opened 2021-09-10
ci: Fix images link for x86_64

Itxaka:
They didnt include the arch but the artifact uploader and github releaser was looking for that


mudler (approved):

Itxaka:
Again, this affects only the release job so no point in waiting around

PR #635, opened 2021-09-13
ci: Drop label checking for arm64 jobs

mudler:
This defeates automation for automatic PRs for bumps


Itxaka:
@mudler could you expand a bit on that? I dont see how a job in a workflow separated from the autobump could affect the bumper???

mudler:
> @mudler could you expand a bit on that? I dont see how a job in a workflow separated from the autobump could affect the bumper???

At the moment it makes a fail the build requirement of the pr jobs, but we should enable it anyhow when opening up PRS

PR #641, opened 2021-09-14
ci: Tag toolchain images

mudler:


mudler:
I gave it a quick shot by rendering the pipelines on the PR jobs and looks fine. we can just tell how it behaves during releases, will create a testing tag later.

PR #642, opened 2021-09-14
Labels

Itxaka:
revert label checking for arm64


Itxaka:
This means that we need to check the requirements for the jobs in the master branch as the names have changed but thats some clicks away

mudler (approved):

PR #647, opened 2021-09-15
ci: Add cloud checker for leftover instances

Itxaka:
First part is just checking and storing the instances we have running
for more than 8 hours on each region


Itxaka:
probably wont be able to test it until is merged. Script is tested locally and works as expected.

Needs:
 - exclude list (so we can exclude instances that we know are running and we want to keep)
 - Auto terminate?? (would rather implement this after running it a couple of times on master to be able to cehck that is picking correctly everything, we dont want to screw up :D )

mudler (approved):
Looks good! Let's try it out then :+1:

PR #650, opened 2021-09-15
grub2-config: allow overriding neednet

Itxaka:


Itxaka:
Im not even sure this is supposed to work...are we overriding the grubenv by the OEM dir somehow? That would allow this to be setup but not after the package is rebuild with `neednet="0"` :thinking:

Itxaka:
ok seems like we are searching for the file grubenv on any disk, so overriding by oem with yip would work...maybe?

mudler:
> ok seems like we are searching for the file grubenv on any disk, so overriding by oem with yip would work...maybe?

Yes, the point there is to do that in runtime, rather while building. It is somewhat imposed by our dracut modules as requires net in order to process certains boot options. The problem is that situation might not apply, and in such case we just not boot at all.

PR #651, opened 2021-09-15
cos: unify all cos packages

### LINKED ISSUE #613: move the cOS package as a single collection
Currently we have recovery and squash-img file as a separate package which makes it really hard to bump manually.

The build.yaml file of recovery and squash-img should be merged with cOS ones so we have a single collection for cOS

n/a

**Additional context**
See the cOS package definition

Itxaka:
Unify all packages under the same collection.

closes #613 


packages/cos/build.yaml
```diff
@@ -131,3 +131,38 @@ excludes:
 - ^/run/.*
 
 {{ end }}
+{{- end}}
+#### system/cos system/cos-container and recovery/cos end here
+
+#### recovery/cos-img starts here
+{{ if eq .Values.name "cos-img" }}
+requires:
```
mudler, line 139: maybe this requires block can be shared with `cos-squash`?

Itxaka: it can but...do we want to? Like I rather have 5 duplicated lines if it means one less if statement. I kind of like the current state of having the 3 different templates one after another instead of mangling everything with ifs.

Itxaka:
blocked cuz I want to build locally and compare packages with those pushed just in case.

mudler (approved):

Itxaka:
I confirmed by using pkgdiff that nothing has changed by introducing this patch :D

I also discovered that we are bundling `/luetbuild` with the cos-squash image, which we should probably exclude :D

mudler:
> I also discovered that we are bundling `/luetbuild` with the cos-squash image, which we should probably exclude :D

That must have been a leftover, please do!

PR #653, opened 2021-09-16
ci: retain old package versions in the published tree

### LINKED ISSUE #602: Can't pin to specified versions because out-of-dated packages are not in the repository
We'd like to have the ability to pin cOS-toolkit packages on specified versions. Because:
- Packages might be removed. e.g.: https://github.com/harvester/harvester/issues/1160
- New packages might include breaking changes and are not tested (from Harvester side).

We can't put a `@version` after packages now because older or removed packages are not indexed in the Green repository.

mudler:
Adds DOWNLOAD_ALL to add old packages back in the index, and enables it
on the publishing pipeline. This will cause the tree to see the old
packages and make them available for installation.

```
--> BUILD --> PULL all metadata 
           -> Create repository (with `--from-metadata`) -> PUBLISH
```
Requires luet 0.17.10 https://github.com/mudler/luet/commit/e4fff77d4390a43552ecd02e4988f6eeb5d49139

Fixes #602

Blocked as requires luet 0.17.10 and needs some manual sanity check as it touches internals and could cause issues.. 

### DIFF
--- a/.github/build.go
+++ b/.github/build.go
@@ -22,7 +22,7 @@ type resultData struct {
 	Exists  bool
 }
 
-func downloadMeta(p Package, o opData) error {
+func downloadImage(img, dst string) error {
 	tmpdir, err := ioutil.TempDir(os.TempDir(), "ci")
 	if err != nil {
 		return err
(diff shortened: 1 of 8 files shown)


Itxaka (approved):
I guess? Need to check the luet change to be sure :D

mudler:
> I guess? Need to check the luet change to be sure :D

sorry, updated the issue. the change is actually really small https://github.com/mudler/luet/commit/e4fff77d4390a43552ecd02e4988f6eeb5d49139 , it's more about testing it. I'm concerned we might see some issues here with packages that are not really semver compliant, so double checking that.

mudler:
mh. couldn't work as nicely as I thought, could I :) seems to _work_ but it takes build dependencies as runtime https://github.com/mudler/luet/commit/e4fff77d4390a43552ecd02e4988f6eeb5d49139. so it needs further work

mudler:
ok, tested also manually and should be good. Unfortunately we can't test this on PRs workflow, the publish step runs only on master

PR #656, opened 2021-09-16
Delete EBS root volumes by default

ibuildthecloud:
By default set DeleteOnTermination so that EBS root volumes are deleted
with the EC2 instance. This is the default value of most public AMIs.


mudler (approved):

mudler:
looks good, 🙏 thanks!

PR #663, opened 2021-09-17
raw_disk: move packages to manifest

### LINKED ISSUE #616: Consume manifest.yaml in image builder script

Make use of `manifest.yaml` file to list all the packages needed for the image builder script.

Consider suggestions form https://github.com/rancher-sandbox/cOS-toolkit/pull/608#discussion_r702968349

This way we can think of disk customization as we do with ISOs.

**Additional context**

Itxaka:
Closes #616 

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -8,6 +8,25 @@ packages:
   - live/grub2-efi-image
   - recovery/cos-img
 
+# Raw disk creation values start
+
+
+raw_disk:
+  # Where to get the packages from if the local build doesn't have the needed packages
+  repo: quay.io/costoolkit/releases-green
+  # which packages to install and the target to install them at
+  packages:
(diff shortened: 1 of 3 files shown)


Itxaka:
Not sure about the format

This way its expandable in the future if we need more parameters per package, but on the other side its more complicated than a list of packages...

mudler (approved):
Like it!

manifest.yaml
```diff
@@ -8,6 +8,24 @@ packages:
   - live/grub2-efi-image
   - recovery/cos-img
 
+# Raw disk creation values start
+
+# Where to get the packages from
```
mudler, line 13: maybe we can have as part of a general `raw_disk` section?

Itxaka: works for me

PR #666, opened 2021-09-20
jq: build arm64 version

Itxaka:
Default version was taking the x86_64 binary release from github for all
arches.

This patch downloads the git repo and builds the binary for arm64 if the
arch is set

### DIFF
--- a/packages/utils/jq/build.yaml
+++ b/packages/utils/jq/build.yaml
@@ -19,7 +19,7 @@ prelude:
 
 steps:
 - |
-  PACKAGE_VERSION=${PACKAGE_VERSION%\+*} && \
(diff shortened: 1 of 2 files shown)


mudler (approved):

mudler:
Nice! Maybe we can also go building with amd64 too?

Itxaka:
cancelled arm64 job until master pushed latest packages of all artifacts. No point building everything here again because we are missing artifacts.

mudler:
We had a green build with both arm and amd64, and there are no tests explictly leveraging jq as it is not installed. It is just consumed by the CI, so I would merge it without waiting tests

PR #680, opened 2021-09-21
tests: introduce GatherAllLogs function

### LINKED ISSUE #674: Recover all logs on failed tests
When a test fails, we have no visibility on what happened so we need to reproduce it locally which can be difficult

On test failure, we should recover logs for several services and systems

A snapshot of the running instance uploaded as an artifact so we can download and launch it locally to examine it fully

**Additional context**

Itxaka:
Instead of adding to each test the files that we waznt to gather, we can
centralize everything in a single function for easy expansion.

GatherAllLogs expands the gathering to service files, full journalctl,
dmesg, uname, disks and users.

closes #674 

### DIFF
--- a/tests/deploys-images/deploy_test.go
+++ b/tests/deploys-images/deploy_test.go
@@ -17,10 +17,7 @@ var _ = Describe("cOS Deploy tests", func() {
 	AfterEach(func() {
 		// Try to gather mtree logs on failure
 		if CurrentGinkgoTestDescription().Failed {
-			s.GatherLog("/tmp/image-mtree-check.log")
-			s.GatherLog("/tmp/luet_mtree_failures.log")
-			s.GatherLog("/tmp/luet_mtree.log")
-			s.GatherLog("/tmp/luet.log")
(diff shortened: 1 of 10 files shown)


tests/deploys-images/deploy_test.go
```diff
@@ -17,10 +17,7 @@ var _ = Describe("cOS Deploy tests", func() {
 	AfterEach(func() {
 		// Try to gather mtree logs on failure
 		if CurrentGinkgoTestDescription().Failed {
-			s.GatherLog("/tmp/image-mtree-check.log")
```
mudler, line 20: I'm a bit unsure , maybe I'm missing something, but now we don't catch those anymore?

mudler: ah, just saw it down below

mudler (approved):
Looks good here!

Itxaka:
squashfs tests passed, this was already tested with a CI failure and picks the correct logs. force pushing as it makes no sense to spend more CI cycles on it.

PR #681, opened 2021-09-21
ci: split docker builds into itw own workflow

Itxaka:
Docker builds take way too much time. If for some reason we need to
relaunch the workflow it means that they need to be also rebuild from
scratch which makes no sense at all.

This patch splits them into theyr own workflow so our main PR workflow
sts ligther and can be retriggered without affecting others

### DIFF
--- a/.github/config/pr-docker.yaml
+++ b/.github/config/pr-docker.yaml
@@ -1,0 +1,30 @@
+---
+
+arches:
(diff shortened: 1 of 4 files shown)


mudler (approved):

mudler:
Let's merge this, there is no need to wait CI here

PR #682, opened 2021-09-21
Switch to dashes instead of plus for revbumps

### LINKED ISSUE #679: Semver comparison between revbumped packages
Currently, our automation handles revbump by incrementing the number after the `+`, although now that with #653 we index all the packages - this means that the `luet` solver will both see e.g. `1.0+1` and `1.0+2` available. 

The problem is that in semantic version `1.0+1` and `1.0+2` are identical from an ordering perspective. Everything after the `+` is considered metadata and can have arbitrary characters, so it's not possible to order them without defining a specific criteria

**Expected behavior**
Ordering should take into account of numbers and pick the latest, or either move away from `+` and use `-` which should be used for internal bumps

**Alternatives**
Do not use `+` for revbumps

**See also**
https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Version

mudler:
"+" is for metadata and not numeric versioning, in fact it cannot be
compared and for e.g. a+1 and a+2 are at the same version when comparing
in semver notation.
By using dashes we refer to internal numbering, and instead can be
ordered

Fixes #679


Itxaka (approved):
Looks good, have you checkd the github CI naming and aws/azure/etc naming?

IIRC in packer we change the last + into a dash because it was not compatible, that may be broken now? Or probably not and we can drop it.

mudler:
> Looks good, have you checkd the github CI naming and aws/azure/etc naming?
> 
> IIRC in packer we change the last + into a dash because it was not compatible, that may be broken now? Or probably not and we can drop it.

I've checked and we just replace + with -, so we should be fine.. let's see if anything breaks

PR #700, opened 2021-09-22
ci: Cancel pending jobs on PR close/merge

Itxaka:
No point in having jobs running once the PR is merged or closed

### DIFF
--- a/.github/workflows/close_jobs.yaml
+++ b/.github/workflows/close_jobs.yaml
@@ -1,0 +1,17 @@
+on:
+  # Runs on PR against master that are closed, that includes merged and closed PRs
+  pull_request:
+    types:
+      - closed
+    branches:
+      - master
+jobs:
+  clean:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Cancel build runs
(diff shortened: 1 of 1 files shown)


Itxaka:
will have to be updated once https://github.com/rancher-sandbox/cOS-toolkit/pull/681 is merged

mudler:
question: does it cancel just jobs associated with the PR, or all jobs matching the workflow ?

Itxaka:
> question: does it cancel just jobs associated with the PR, or all jobs matching the workflow ?

jobs matching the **branch**

Usually is matching the branch and sha, so it can cancel old jobs for a branch, but in our case we want to cancel all jobs from branch X ignoring the sha, as it wont change on close event.

mudler (approved):

mudler:
sounds great! let's go for it.. no sense to wait for CI here

Itxaka:
Need to update to add the new docker workflow

Itxaka:
I guess we may see if it works rigth now or crashes and burns and cancels the whole org jobs LMAO

PR #701, opened 2021-09-22
ci: Actually be able to terminate jobs from forks

Itxaka:
Becuase our Prs come from fork, the actions runner has no perms to close
those jobs.

Instead we need to trigger the cancel job from the  workflows and check
if its a pr that was closed/merged


mudler (approved):

Itxaka:
Im not even sure if this works...it should have triggered already...

PR #706, opened 2021-09-22
Reattempt to download image list

mudler:
with a backoff in case of failure


Itxaka (approved):

mudler:
maybe we should get this in as well. the part which are touched are not exercised in the PR workflow

PR #707, opened 2021-09-22
ci: Require labels for certain jobs without failing

Itxaka:
Instead of checking and failing all the time, we can just use a if on
the main jobs that trigger the rest, as if those are skipped, the rest
should be and we wont get the ugly red X everywhere.

This also should allow us to use more labels in PRs

### DIFF
--- a/.github/build.yaml.gomplate
+++ b/.github/build.yaml.gomplate
@@ -67,18 +67,29 @@
   {{{ $config := (datasource "config") }}}
   {{{- if $config.local_runner }}}
     runs-on: [{{{ join $config.runs_on ", "}}}]
-    if: contains(["mudler", "davidcassany", "itxaka", kkaempf", "cOS-cibot"], "${{ github.actor }}") || contains(github.event.pull_request.labels.*.name, 'safe to test')
   {{{- else }}}
     runs-on: ubuntu-latest
   {{{ end }}}
 {{{end}}}
 
-{{{define "needs"}}}
-    {{{ $config := (datasource "config") }}}
-    {{{if has $config "labels"}}}
-    needs:
-      - check-labels
-    {{{end}}}
+{{{- define "generate_labels"}}}
+{{{- $config := (datasource "config") }}}
+{{{- range $config.labels}}} && contains(github.event.pull_request.labels.*.name, '{{{ . }}}'){{{- end}}}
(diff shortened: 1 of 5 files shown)


mudler (approved):
yess! :partying_face:

.github/workflows/build-master-arm64.yaml
```diff
@@ -9,7 +9,7 @@ concurrency:
 jobs:
   build-green-arm64:
     runs-on: [self-hosted, arm64]
-    if: contains(["mudler", "davidcassany", "itxaka", kkaempf", "cOS-cibot"], "${{ github.actor }}") || contains(github.event.pull_request.labels.*.name, 'safe to test')
```
Itxaka, line 12: how the F did this work?????

Itxaka: id didnt work since the start lmao. The expression engine does not accept array literals but it just doesnt fail LMAO

Itxaka:
yay, now its skipped and accepts more than one label!

Itxaka:
@mudler asking for a re-review as it has changed from the time it was approved to deal with more than one label

mudler (approved):

.github/build.yaml.gomplate
```diff
@@ -67,26 +67,37 @@
...
+{{{- define "generate_labels"}}}
+{{{- $config := (datasource "config") }}}
+{{{- range $config.labels}}} && contains(github.event.pull_request.labels.*.name, '{{{ . }}}'){{{- end}}}
+{{{- end}}}
+
+{{{define "if_with_labels"}}}
+{{{ $config := (datasource "config") }}}
+{{{if has $config "labels"}}}
+  {{{- if $config.local_runner }}}
+    if: contains(fromJson('["mudler", "davidcassany", "itxaka", "kkaempf", "cOS-cibot"]'), github.actor) {{{- tmpl.Exec "generate_labels" }}}
```
mudler, line 84: That's neat, guessing that later on we could switch to handle users in the same way for labels

Itxaka: would be great if we could automagically get the members of the cos team and use it here :D

Itxaka:
not touching normal workflows and tested with labels already

PR #708, opened 2021-09-23
ci: set default download retries to 10

mudler:
Maybe this should be enough...


Itxaka (approved):

mudler:
Let's try it out!

PR #711, opened 2021-09-23
Iso arm64

Itxaka:


Itxaka:
Ready for review @mudler

make/Makefile.iso
```diff
@@ -95,6 +95,7 @@ else
 	$(YQ) e -i '.luet.repositories[0].enable=true' $(MANIFEST).remote
 	$(YQ) e -i '.luet.repositories[0].priority=90' $(MANIFEST).remote
 	$(YQ) e -i '.luet.repositories[0].type="docker"' $(MANIFEST).remote
+	$(YQ) e -i '.luet.repositories[0].type="docker"' $(MANIFEST).remote
```
mudler, line 98: mm this looks like a leftover?

Itxaka: :thinking: weird

mudler (approved):
Looks good! That's great!

packages/livecd/grub2/build.yaml
```diff
@@ -1,28 +1,35 @@
-image: opensuse/leap:15.3
+image: {{ .Values.image }}
```
mudler, line 1: If we are switching image here, the `zypper` call below should be adjusted to take into account blue and orange

Itxaka: indeed, Im wondering if it makes sense to have this package in other flavors....

mudler (changes requested):
Just some nits, otherwise looks good

packages/livecd/grub2/build.yaml
mudler: Previously we were using files from opensuse for the other flavors too, we did that to let the user just specify one repo for flavor.. maybe we can do the same on aarch64 too?

Itxaka:
arm iso built!

mudler (approved):

PR #716, opened 2021-09-24
ci: Add arm64 iso building to master/releases

Itxaka:


Itxaka:
should only affect master and release arm64 versions

mudler (approved):

PR #719, opened 2021-09-24
installer: User arch variables for installation

### LINKED ISSUE #718: cos-installer fails in arm64
**cos-toolkit version:**
master

arm64

**Describe the bug**
running cos-installer when booting from the raw disk image results into an error due to a hardcoded path:

```
grub2-install: error: /usr/share/grub2/x86_64-efi/modinfo.sh doesn't exist. Please specify --target or --directory.
```

**To Reproduce**
`cos-installer /dev/vdb`

**Expected behavior**
the script takes into account the arch

**Logs**

**Additional context**

Itxaka:
Use the current ARCH for things like installing grub.

Also changes the squashfs filter for compression depending on arch

Fixes #718 ... I think...

### DIFF
--- a/packages/installer/definition.yaml
+++ b/packages/installer/definition.yaml
@@ -1,4 +1,4 @@
 name: "installer"
 category: "utils"
-version: "0.17"
+version: "0.17.1"
 description: "Installer, Upgrade and reset utilities to manage cOS derivatives"
--- a/packages/cos/collection.yaml
+++ b/packages/cos/collection.yaml
@@ -2,7 +2,7 @@ packages:
   - &cos
     name: "cos"
     category: "system"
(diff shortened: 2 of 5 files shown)


packages/installer/definition.yaml
```diff
@@ -1,4 +1,4 @@
 name: "installer"
 category: "utils"
-version: "0.17"
+version: "0.17-1"
```
mudler, line 3:
```suggestion
version: "0.17.1"
```

Itxaka: :sob:

Itxaka:
Manually tested. This makes the livecd installer work

mudler (approved):

PR #720, opened 2021-09-24
ci: Always store logs on failure

Itxaka:


Itxaka:
Directly merging as it only affects failures in tests

PR #728, opened 2021-09-27
ci: Run examples on master push

Itxaka:


Itxaka:
This only affects master so force-merging

PR #740, opened 2021-09-29
make: fix FINAL_REPO order

Itxaka:
It makes sense to first define your var and then check fir its value
rather than the reverse. Otherwise your var will always be empty :/

Only affected local runs of `publish-repo` and `iso`


mudler (approved):
Good catch!

PR #748, opened 2021-09-30
CI: Run tests for arm64

Itxaka:


mudler (approved):
Not a vagrant/libvirt user, but looks good here overall

PR #759, opened 2021-10-05
Support GPT on legacy BIOS system

johnliu55tw:
The main reason behind this PR is that we have users running on legacy BIOS systems with storage greater than MBR limit (2.2TB), so we need to create GPT for their system.

**Some additional thoughts**: I'm thinking maybe GPT should be the default option, regardless of booting from EFI or Legacy BIOS, because GRUB can use GPT for both. Maybe MBR is the one that should be `--forced`, only when users encountered some issue.

### DIFF
--- a/packages/installer/installer.sh
+++ b/packages/installer/installer.sh
@@ -50,7 +50,7 @@ cleanup()
 
 usage()
 {
-    echo "Usage: $PROG [--force-efi] [--iso https://.../OS.iso] [--debug] [--tty TTY] [--poweroff] [--no-format] [--config https://.../config.yaml] DEVICE"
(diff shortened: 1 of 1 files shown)


mudler (changes requested):
It looks good to me, however it needs a version bump into the installer definition file here: https://github.com/rancher-sandbox/cOS-toolkit/blob/master/packages/installer/definition.yaml#L3. Otherwise there is no new installer version with your changes included (and neither CI will see them)

mudler (approved):
I've added the package bumps on top of the PR, looks good here, thanks! :+1:

johnliu55tw:
Thank you so much @mudler!

tscswcn:
has harvester support GPT as default within 0.30 release？

PR #762, opened 2021-10-06
ci: Add new instance to exclude list

Itxaka:
I screwed up the old instance yesterday while hacking away at the arm64 tests so we need to prepare a new one. 

This patch adds it to the exclude list so its not cleaned automatically


Itxaka:
Merging this directly as it doesnt affect anything else than the cloud cleaner job

PR #768, opened 2021-10-08
ci: Build qemu boxes in ubuntu

Itxaka:
This allows us to free more macos instances on the CI (Max 5 at a time) for the tests, so if there is more than one job in parallel they are not blocked by this


mudler (approved):

Itxaka:
oh, it failed the test only, good. Kind of.

PR #773, opened 2021-10-08
ci: Push arm64 artifacts for other flavors

Itxaka:


Itxaka:
Requires the repos to be available first.

Itxaka:
Dracut is not building properly for orange-arm64

initdir on module-setup.sh seems to be empty somehow

mudler:
> looks like the same issue is happening on x86_64 version for orange darcut pakcage, so ignoraring that and opening a ticket to follow it up. Maybe the resulting artifacts are broken...it will require a manual check thought.

could be. Maybe we should add a comment in the package explaining the rationale. @davidcassany I think it was because at the end we are always consuming the green package, right?

PR #821, opened 2021-10-28
Customize grub menu

mudler:
This PR adds in the GRUB configuration context:

- a `extra_cmdline` that can be specified as grub env config to specify additional booting kernel parameters (applied to all entries, `extra_active_cmdline`, `extra_passive_cmdline` and `extra_recovery_cmdline` for specific entrypoints)
- a `display_name` var that can be specified to tweak the default booting menu entry (in place of the current sed logic)
- a way to add additional config to the grub menu. If a file `/grubmenu` is found (e.g. in `COS_STATE`) it is sourced, and eventual menuentries declared inside are displayed in the boot menu
- a way to customize the fallback logic

It still needs tests and docs to be updated too

### DIFF
--- a/packages/grub2/config/grub.cfg
+++ b/packages/grub2/config/grub.cfg
@@ -57,7 +57,7 @@ menuentry "${display_name}" --id cos {
   loopback loop0 /$img
   set root=($root)
   source (loop0)/etc/cos/bootargs.cfg
-  linux (loop0)$kernel $kernelcmd ${extra_cmdline}
+  linux (loop0)$kernel $kernelcmd ${extra_cmdline} ${extra_active_cmdline}
   initrd (loop0)$initramfs
 }
 
@@ -68,7 +68,7 @@ menuentry "${display_name} (fallback)" --id fallback {
   loopback loop0 /$img
   set root=($root)
   source (loop0)/etc/cos/bootargs.cfg
-  linux (loop0)$kernel $kernelcmd ${extra_cmdline}
+  linux (loop0)$kernel $kernelcmd ${extra_cmdline} ${extra_passive_cmdline}
   initrd (loop0)$initramfs
 }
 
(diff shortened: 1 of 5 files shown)


mudler:
@Itxaka needs still docs update, but I think I've figured why tests were failing, but should be good for a first review!

packages/grub2/config/grub.cfg
```diff
@@ -15,37 +29,50 @@ else
...
-menuentry "cOS" --id cos {
+menuentry "${display_name}" --id cos {
   search.fs_label COS_STATE root
   set img=/cOS/active.img
   set label=COS_ACTIVE
   loopback loop0 /$img
   set root=($root)
   source (loop0)/etc/cos/bootargs.cfg
-  linux (loop0)$kernel $kernelcmd
+  linux (loop0)$kernel $kernelcmd ${extra_cmdline} ${extra_active_cmdline}
```
Itxaka, line 60: Not directly related, but should we just load everything, including kernelcmd from env files directly so the whole thing is overwritable?

packages/grub2/config/grub.cfg
```diff
@@ -1,12 +1,26 @@
 set timeout=10
 
+# Load custom env file
 set env_file="/grubenv"
 search --file --set=env_blk "${env_file}"
 
+# Load custom env file
+set oem_env_file="/grub_oem_env"
```
mudler, line 8: The rebranding logic writes in this file instead of `/grubenv` or `/grubmenu`. Those are left for users to specify, while during install we instrument only `/grub_oem_env`

Itxaka (approved):
Looks good to me

packages/grub2/config/grub.cfg
mudler: the kernelcmd comes from the derivative itself already as an env file, so the derivative have full control over the boot commands. However, there is no explicit way for users to pin specific flags, while now you could:

For all:
```
grub2-editenv /oem/grubenv set extra_cmdline=...
``` 

For the single entry, e.g. active:
```
grub2-editenv /oem/grubenv set extra_active_cmdline=...
```

mudler:
Docs PR at https://github.com/rancher-sandbox/cos-toolkit-docs/pull/55

PR #833, opened 2021-11-02
Meta cos verify

Itxaka:


examples/standard/Dockerfile
```diff
@@ -66,7 +70,10 @@ COPY conf/luet.yaml /etc/luet/luet.yaml
 # Copy luet from the official images
 COPY --from=luet /usr/bin/luet /usr/bin/luet
 
-RUN luet install -y meta/cos-minimal \
+# TODO: change this to meta/cos-verify once the meta package is published
+RUN luet install -y toolchain/cosign toolchain/luet-cosign
```
mudler, line 74:
```suggestion
RUN luet install -y meta/cos-verify
```

packages/meta/collection.yaml
```diff
@@ -83,3 +83,13 @@ packages:
       - category: system
         name: cloud-config
         version: ">=0"
+  - category: "meta"
```
mudler, line 86: :+1:

examples/standard/Dockerfile
mudler: Something that "puzzles" me is that in this way, when we get cosign itself and the plugin, those are not verified (via signatures, etc). 

Shall we maybe create a signed container where we take this tools from instead?

Itxaka: wont work :D until this same PR is merged and the meta package published

I mean, it wont fail as luet will fail to install the packages and luet will also fail to find the plugins, so the dockerfile will be built correctly

But I want this to be tested on CI in case its broken (it worked on my machine, yadayadayada :P)

Itxaka: tested, so updating this to use the meta package. It will eventually work, like cluster synchronization LOL

mudler (approved):

Itxaka:
No need to wait for tests, this shouldn't affect existing packages/tests only the dockerfile and those run

PR #836, opened 2021-11-03
ci: Fixup retrieving missing packages

mudler:
This had the unpleasant effect of making each PR/build to rebuild the
whole repository each time


mudler:
@Itxaka this also caused a lot of slowdown into jobs.. each build job was taking always 50m, even if had nothing to build :(

Itxaka:
Heh, i was wondering why the build time shoot up lately 😁

Itxaka (approved):

mudler:
Right, to check if we grab all the metadata correctly we can just check out on master.. merging it

Issue #734, opened 2021-09-28
ci: testing the installer


mudler:
Currently we just run a shallow test on the installer: the test process starts by creating a vagrant image by using `cos-installer`.

It would be nice to have a test suite for `cos-installer`, as likely would also help with #581 and increase the test coverage area.

## Action items

- [ ] Find out a way to run automated test for the installer ( vagrant, insert iso on runtime and reboot )
- [ ] Create a simple test suite to leverage the installer options ( e.g. test `after-install-chroot` hook )

Itxaka:
Doesn't cover all tests but it's a start and a framework to cover more in the future.

I will create s follow up card with some of the expected tests.

One thing though is that efi testing seems to take waaaay longer for some weird reason...but most of the time spent there is during cos-install for some weird reason. I will create another card to follow up on that, in case it's just not just a coincidence but something wrong with the installer. Both run in macos and the virtual drives should be the same, but the times are veeeery different (2 minutes Vs 8/13 minutes!)

PR #839, opened 2021-11-03
Partitioning layout

mudler:
This makes the installer to accept a cloud-init file, - only with GPT partition types - like the following:

```yaml
stages:
   partitioning:
     - name: "Repart disk"
       layout:
         device:
           path: /dev/sda
         add_partitions:
           - fsLabel: COS_STATE
             size: 8192
             pLabel: state
           - fsLabel: COS_OEM
             size: 10
             pLabel: oem
           - fsLabel: COS_RECOVERY
             # default filesystem is ext2 if omitted
             filesystem: ext4
             size: 40000
             pLabel: recovery
           - fsLabel: COS_PERSISTENT
             pLabel: persistent
             # default filesystem is ext2 if omitted
             filesystem: ext4
             size: 40000
```

Of course, it can be also used to create additional partitions, or either create those into a different device (by setting `path`), having more stages, etc.

That can be supplied to the installer with `cos-installer --partition-layout <file> <device>`

VirtualBox_cOS teststt_04_11_2021_12_55_21

### DIFF
--- a/examples/standard/files/etc/cos/config
+++ b/examples/standard/files/etc/cos/config
@@ -37,3 +37,6 @@ COSIGN_EXPERIMENTAL=1
 # This can be set to file, URL, KMS URI or Kubernetes Secret
 # This is only used if COSIGN_EXPERIMENTAL is set to 0
 COSIGN_PUBLIC_KEY_LOCATION=""
+
+# Default size (in MB) of disk image files (.img) created during upgrades
+DEFAULT_IMAGE_SIZE=3240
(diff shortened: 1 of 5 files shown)


mudler:
Not possible to add tests due to #734

mudler:
There is a small change required into yip to make it work. Between mkfs calls, `yip` tries to detect the partition location and calls `partprobe` to recall tables, however that doesn't seems to be sufficient while testing locally. If I define more than 2/3 partitions, it starts to error out that can't find the partition number. A call to `sync` in between seems to fix the issue

mudler:
Done, included in `0.9.21`

Itxaka (approved):

mudler:
Docs PR: https://github.com/rancher-sandbox/cos-toolkit-docs/pull/56

PR #848, opened 2021-11-05
Setup installer testing

Itxaka:


tests/installer/installer_test.go
```diff
@@ -0,0 +1,68 @@
...
+		s = sut.NewSUT()
+		s.EventuallyConnects()
+	})
+
+	AfterEach(func() {
+		if CurrentGinkgoTestDescription().Failed {
+			s.GatherAllLogs()
+		} else {
+			// Trash disk so we boot from iso after each test
+			s.EmptyDisk("/dev/sda")
```
mudler, line 23: maybe we should empty the disk before each one? Otherwise it depends from the first one to be able to run successfully (and can be a bit confusing?)

tests/installer/installer_test.go
```diff
@@ -0,0 +1,68 @@
...
+			// Reboot so it boots from iso
+			s.Reboot()
+		}
+	})
+	Context("Using legacy bios", func() {
+		Context("generic tests", func() {
+			It("booted from iso", func() {
```
mudler, line 30: mmm and maybe we can do these assertion right after we empty the disk, before each suite?

mudler (approved):

mudler:
Overall looks good! :100:

PR #851, opened 2021-11-08
Allow to specify an image to cos-installer

mudler:
By making `cos-installer` capable to take container images in place of the ISO content, we are untying the installer from the ISO. This has two benefit:

1) the installer can be used to deploy cOS (and any derivative) outside of the ISO context, also standalone. For instance, IF in a booting (e.g. a 3rd party) LiveCD we can run docker, we could install cOS by using the installer inside our container image, specifying the partition and the image to deploy (I'll work on instructions later)
2) remove code dups. `cos-deploy` was partially crossing with `cos-installer`. The only difference is that `cos-deploy` doesn't format, and just copy an image during install, which now `cos-installer` is capable of too as well.

This commit drops cos-deploy * logics, but keeps the wrapper by calling `cos-installer` accordingly.

Additionally, cos-installer now accepts: `--docker-image`, `--no-verify` and `--no-cosign`

### DIFF
--- a/packages/installer/cos.sh
+++ b/packages/installer/cos.sh
@@ -94,17 +94,12 @@ is_booting_from_squashfs() {
     fi
 }
 
-prepare_deploy_target() {
-    if [ -f  ${STATEDIR}/cOS/active.img ] && [ "$FORCE" != "true" ]; then
-        echo "There is already an active deployment in the system, use '--force' flag to overwrite it"
-        exit 1
+is_booting_from_live() {
+    if cat /proc/cmdline | grep -q "CDLABEL"; then
+        return 0
+    else
+        return 1
     fi
-
-    mkdir -p ${STATEDIR}/cOS || true
-    rm -rf ${STATEDIR}/cOS/active.img || true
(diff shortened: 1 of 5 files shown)


mudler:
@Itxaka should be good for a first review. Tests are still not passing, but I think I've got almost all of them fixed

packages/installer/cos.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -e
+set -ex
```
mudler, line 2: Going to drop `-x` as soon as tests are green :)

Itxaka:
Looking good to me for now. Do you plan to add tests to the installer for this new feature? 😉

mudler:
> Looking good to me for now. Do you plan to add tests to the installer for this new feature? wink

I thought at it, but then, the feature is actually tested by cos-deploy tests

mudler:
deploy-image and raw deploy tests passed :tada:  (there was just some flakyness, as some failed for network issues :/ )

Itxaka (approved):
Sorry, I thought I already reviewed this!

mudler:
Docs at https://github.com/rancher-sandbox/cos-toolkit-docs/pull/59

PR #866, opened 2021-11-15
ci: Enable vbox recording on tests

Itxaka:

### DIFF
--- a/packer/images.json.pkr.hcl
+++ b/packer/images.json.pkr.hcl
@@ -151,6 +151,9 @@ source "virtualbox-iso" "cos" {
   ssh_timeout            = "5m"
   ssh_username           = "${var.root_username}"
   vm_name                = "cOS"
+  vboxmanage = [
+    ["modifyvm", "{{.Name}}", "--recording", "${var.enable_video_capture}", "--recordingscreens", "0","--recordingfile", "../capture.webm"],
+  ]
 }
 
 build {
--- a/.github/build.yaml.gomplate
+++ b/.github/build.yaml.gomplate
@@ -324,6 +324,12 @@
           path: |
             packer/*.box
           if-no-files-found: error
+      - uses: actions/upload-artifact@v2
+        if: failure()
+        with:
+          name: cOS-Packer-{{{$subset}}}-{{{ $flavor }}}-vbox-{{{ $config.arch }}}.capture.zip
+          path: capture.webm
+          if-no-files-found: error
 {{{ end }}}
 
 {{{define "test"}}}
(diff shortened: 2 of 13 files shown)


Itxaka:
Seems to work :D

packer/images.json.pkr.hcl
```diff
@@ -151,6 +151,9 @@ source "virtualbox-iso" "cos" {
   ssh_timeout            = "5m"
   ssh_username           = "${var.root_username}"
   vm_name                = "cOS"
+  vboxmanage = [
+    ["modifyvm", "{{.Name}}", "--recording", "on", "--recordingscreens", "0","--recordingfile", "../capture.webm"],
```
mudler, line 155: mmm just a silly question, is this going to record nevertheless when we ran the packer template? there is no way to optionally turn it off/on?

mudler (approved):
Looks good here!

mudler:
mmm seeing this in test-raw disks:

```
Run actions/upload-artifact@v2
With the provided path, there will be 1 file uploaded
events.js:187
      throw er; // Unhandled 'error' event
      ^

Error: EACCES: permission denied, open '/Users/runner/work/cOS-toolkit/cOS-toolkit/capture.webm'
Emitted 'error' event on ReadStream instance at:
    at internal/fs/streams.js:120:12
    at FSReqCallback.oncomplete (fs.js:146:23) {
  errno: -13,
  code: 'EACCES',
  syscall: 'open',
  path: '/Users/runner/work/cOS-toolkit/cOS-toolkit/capture.webm'
}
```
https://github.com/rancher-sandbox/cOS-toolkit/runs/4211735613?check_suite_focus=true

packer/images.json.pkr.hcl
Itxaka: no way of turning off for now. Do we want to skip it?

Itxaka: I mean, we could just not do it in the packer or hide it under a debug env var I guess, packer is very consistent and usually never fails, and its easy to reproduce locally

mudler: :+1: fine by me, was just curious.. I just wondered about webm support. Is something backed inside Vbox I guess, so should be fine!

Itxaka:
> mmm seeing this in test-raw disks:

should be fixed, permissions issue

packer/images.json.pkr.hcl
Itxaka: there, updated with a switch so its off by default

mudler: Awesome! 👍

Itxaka:
mmhh, this actually introduces a kind of heavy impact on the speed, with some tests taking up to 400 more seconds which I dont think its viable.

Im thinking Im gonna set it behind a debug keyword like on packer (capture_enabled) and have it disabled by default.

Issue #859, opened 2021-11-11
Expand tests for installer


Itxaka:
**Is your feature request related to a problem? Please describe.**
Currently we have minimal tests for installer, we should expand them

**Describe the solution you'd like**
Expanding tests for the installer

**Describe alternatives you've considered**
not testing anything and living la vida loca

**Additional context**
The installer tests already has some test descriptions on them as guidance.

mudler:
#581 contains already a small list of scenarios which aren't currently covered, but I guess most likely we would like to cover what we have documented: https://rancher-sandbox.github.io/cos-toolkit-docs/docs/getting-started/install/

Itxaka:
Part of this is done, missing installer options

    –iso (How to serve the iso?)
    –poweroff (Can probably be only tested on bios)
    –no-format (Requires a pre-format)
    –docker-image
    –debug ( needed?)
    –no-verify
    –no-cosign

Itxaka:
docker-image is probably the most important currently and should be straightforward to implement

PR #871, opened 2021-11-17
Refactor internal vars installer

### LINKED ISSUE #797: Upgrading from a directory fails if the IMAGE variable exists in environment files
**cos-toolkit version:**
cOS-Seed-green-0.7.0-12-g3e2e084a-x86_64.is

Linux cos 5.3.18-59.24-default #1 SMP Mon Sep 13 15:06:42 UTC 2021 (2f872ea) x86_64 x86_64 x86_64 GNU/Linux

**Describe the bug**

If there is an `IMAGE=xxx` line in `/etc/os-release` file, running `cos-upgrade` command to upgrade from a directory fails.

**To Reproduce**

- Add an `IMAGE=xxxx` line:
```
cos:~ # cp /etc/os-release /tmp/
cos:~ # echo >> /tmp/os-release
cos:~ # echo 'IMAGE="quay.io/costoolkit/releases-green:cos-system-0.7.0-14"' >> /tmp/os-release
cos:~ # mount -o bind /tmp/os-release /etc/os-release
cos:~ # cat /etc/os-release
ANSI_COLOR="0;32"
BUG_REPORT_URL="https://github.com/mudler/cOS/issues"
DOCUMENTATION_URL="https://github.com/mudler/cOS"
HOME_URL="https://github.com/mudler/cOS"
ID="cOS"
NAME="cOS"
PRETTY_NAME="cOS v0.7.0-12"
VERSION="0.7.0-12"
IMAGE="quay.io/costoolkit/releases-green:cos-system-0.7.0-14"
```

- Mount a squashfs image to a directory and upgrade from it
```
# I upload the squashfs from an ISO to /usr/local 
cos:~ # ls -l /usr/local/rootfs.squashfs
-rw-r--r-- 1 root root 314286080 Oct 14 08:10 /usr/local/rootfs.squashfs
cos:~ # mkdir /tmp/new
cos:~ # mount /usr/local/rootfs.squashfs /tmp/new/
cos:~ # cos-upgrade --directory /tmp/new/
Upgrading to image quay.io/costoolkit/releases-green:cos-system-0.7.0-14
Upgrading system..
-> Upgrade target: active.img
3240+0 records in
3240+0 records out
3397386240 bytes (3.4 GB, 3.2 GiB) copied, 1.54999 s, 2.2 GB/s
mke2fs 1.43.8 (1-Jan-2018)
Discarding device blocks: done
Creating filesystem with 829440 4k blocks and 207584 inodes
Filesystem UUID: 41da2a4e-4fc3-4c06-acfc-89f5756f4e92
Superblock backups stored on blocks:
	32768, 98304, 163840, 229376, 294912, 819200

Allocating group tables: done
Writing inode tables: done
Writing superblocks and filesystem accounting information: done

mkdir: cannot create directory ‘/tmp/upgrade/usr/local’: No such file or directory
INFO[0000] yip version 0.9.14-g8cc8e0e3e42aa4d70715cc5ef97693f0d490d61d 2021-10-11 05:03:41 UTC
INFO[0000] Executing /system/oem/00_rootfs.yaml
INFO[0000] Executing yip file                            name="Rootfs Layout Settings" stage=before-upgrade.before stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade.before stages=0 success=true
INFO[0000] Executing /system/oem/01_defaults.yaml
INFO[0000] Executing yip file                            name="General settings" stage=before-upgrade.before stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade.before stages=0 success=true
INFO[0000] Executing /system/oem/02_upgrades.yaml
INFO[0000] Executing yip file                            name="Upgrades settings" stage=before-upgrade.before stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade.before stages=0 success=true
INFO[0000] Executing /system/oem/03_branding.yaml
INFO[0000] Executing yip file                            name=Branding stage=before-upgrade.before stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade.before stages=0 success=true
INFO[0000] Executing /system/oem/04_accounting.yaml
INFO[0000] Executing yip file                            name="Default user" stage=before-upgrade.before stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade.before stages=0 success=true
INFO[0000] Executing /system/oem/05_network.yaml
INFO[0000] Executing yip file                            name="Default network configuration" stage=before-upgrade.before stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade.before stages=0 success=true
INFO[0000] Executing /system/oem/06_recovery.yaml
INFO[0000] Executing yip file                            name="Recovery partition boot setup" stage=before-upgrade.before stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade.before stages=0 success=true
INFO[0000] Executing /system/oem/07_live.yaml
INFO[0000] Executing yip file                            name="LiveCD Detection" stage=before-upgrade.before stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade.before stages=0 success=true
INFO[0000] yip version 0.9.14-g8cc8e0e3e42aa4d70715cc5ef97693f0d490d61d 2021-10-11 05:03:41 UTC
INFO[0000] Executing /system/oem/00_rootfs.yaml
INFO[0000] Executing yip file                            name="Rootfs Layout Settings" stage=before-upgrade stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade stages=0 success=true
INFO[0000] Executing /system/oem/01_defaults.yaml
INFO[0000] Executing yip file                            name="General settings" stage=before-upgrade stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade stages=0 success=true
INFO[0000] Executing /system/oem/02_upgrades.yaml
INFO[0000] Executing yip file                            name="Upgrades settings" stage=before-upgrade stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade stages=0 success=true
INFO[0000] Executing /system/oem/03_branding.yaml
INFO[0000] Executing yip file                            name=Branding stage=before-upgrade stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade stages=0 success=true
INFO[0000] Executing /system/oem/04_accounting.yaml
INFO[0000] Executing yip file                            name="Default user" stage=before-upgrade stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade stages=0 success=true
INFO[0000] Executing /system/oem/05_network.yaml
INFO[0000] Executing yip file                            name="Default network configuration" stage=before-upgrade stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade stages=0 success=true
INFO[0000] Executing /system/oem/06_recovery.yaml
INFO[0000] Executing yip file                            name="Recovery partition boot setup" stage=before-upgrade stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade stages=0 success=true
INFO[0000] Executing /system/oem/07_live.yaml
INFO[0000] Executing yip file                            name="LiveCD Detection" stage=before-upgrade stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade stages=0 success=true
INFO[0000] yip version 0.9.14-g8cc8e0e3e42aa4d70715cc5ef97693f0d490d61d 2021-10-11 05:03:41 UTC
INFO[0000] Executing /system/oem/00_rootfs.yaml
INFO[0000] Executing yip file                            name="Rootfs Layout Settings" stage=before-upgrade.after stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade.after stages=0 success=true
INFO[0000] Executing /system/oem/01_defaults.yaml
INFO[0000] Executing yip file                            name="General settings" stage=before-upgrade.after stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade.after stages=0 success=true
INFO[0000] Executing /system/oem/02_upgrades.yaml
INFO[0000] Executing yip file                            name="Upgrades settings" stage=before-upgrade.after stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade.after stages=0 success=true
INFO[0000] Executing /system/oem/03_branding.yaml
INFO[0000] Executing yip file                            name=Branding stage=before-upgrade.after stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade.after stages=0 success=true
INFO[0000] Executing /system/oem/04_accounting.yaml
INFO[0000] Executing yip file                            name="Default user" stage=before-upgrade.after stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade.after stages=0 success=true
INFO[0000] Executing /system/oem/05_network.yaml
INFO[0000] Executing yip file                            name="Default network configuration" stage=before-upgrade.after stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade.after stages=0 success=true
INFO[0000] Executing /system/oem/06_recovery.yaml
INFO[0000] Executing yip file                            name="Recovery partition boot setup" stage=before-upgrade.after stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade.after stages=0 success=true
INFO[0000] Executing /system/oem/07_live.yaml
INFO[0000] Executing yip file                            name="LiveCD Detection" stage=before-upgrade.after stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade.after stages=0 success=true
INFO[0000] yip version 0.9.14-g8cc8e0e3e42aa4d70715cc5ef97693f0d490d61d 2021-10-11 05:03:41 UTC
INFO[0000] yip version 0.9.14-g8cc8e0e3e42aa4d70715cc5ef97693f0d490d61d 2021-10-11 05:03:41 UTC
INFO[0000] yip version 0.9.14-g8cc8e0e3e42aa4d70715cc5ef97693f0d490d61d 2021-10-11 05:03:41 UTC
INFO[0000] yip version 0.9.14-g8cc8e0e3e42aa4d70715cc5ef97693f0d490d61d 2021-10-11 05:03:41 UTC
INFO[0000] Executing /proc/cmdline
INFO[0000] Executing yip file                            name= stage=before-upgrade.before stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade.before stages=0 success=true
INFO[0000] yip version 0.9.14-g8cc8e0e3e42aa4d70715cc5ef97693f0d490d61d 2021-10-11 05:03:41 UTC
INFO[0000] Executing /proc/cmdline
INFO[0000] Executing yip file                            name= stage=before-upgrade stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade stages=0 success=true
INFO[0000] yip version 0.9.14-g8cc8e0e3e42aa4d70715cc5ef97693f0d490d61d 2021-10-11 05:03:41 UTC
INFO[0000] Executing /proc/cmdline
INFO[0000] Executing yip file                            name= stage=before-upgrade.after stages=0
INFO[0000] Finished yip file execution                   stage=before-upgrade.after stages=0 success=true
Upgrading from local folder: quay.io/costoolkit/releases-green:cos-system-0.7.0-14
rsync: change_dir "/root//quay.io/costoolkit/releases-green:cos-system-0.7.0-14" failed: No such file or directory (2)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1189) [sender=3.1.3]
```

**Expected behavior**

The upgrade should succeed.
The upgrade source should be the specified directory rather than the IMAGE variable value.

**Logs**

**Additional context**

The [`IMAGE` variable set in the parse opts loop](https://github.com/rancher-sandbox/cOS-toolkit/blob/49eb210468fa7a70a94e1c0cf8ced3ec2d66cd11/packages/installer/upgrade.sh#L352) is overrided after sourcing some environment files: https://github.com/rancher-sandbox/cOS-toolkit/blob/49eb210468fa7a70a94e1c0cf8ced3ec2d66cd11/packages/installer/upgrade.sh#L66-L89

### LINKED ISSUE #799: Differentiate internal variables
#797 shows how easy is to interact with internal variables from config files. the `IMAGE` variable is not intended to be configured or set via config file, but if that happens the upgrade process is broken.

Variables that are meant to be used internally inside the scripts shouldn't conflict with global variables. Use variable prefixes (e.g. `cos_image`, `_cos_image` instead of `IMAGE`) or whatever necessary to avoid clashes

**Additional context**
#797

Itxaka:
Change all vars used in the script to have a `_` in front so no collisions with existing vars
Use more locals
Set proper override of values in order Defaults -> config -> env vars -> switches
The change above means that I was able to remove some of the assignments and move them all to the same load_env_vars function

Fixes #799 
Fixes #797


Itxaka:
soooo

It has 63 commits because in case that it fails its easier to bisect. It will be squashed

packages/installer/cos.sh
```diff
@@ -4,17 +4,16 @@ set -e
...
-RECOVERYDIR=/run/cos/recovery
-RECOVERYSQUASHFS=${ISOMNT}/recovery.squashfs
-GRUBCONF=/etc/cos/grub.cfg
+_PROGS="dd curl mkfs.ext4 mkfs.vfat fatlabel parted partprobe grub2-install grub2-editenv"
+_DISTRO=/run/rootfsbase
+_ISOMNT=/run/initramfs/live
+_TARGET=/run/cos/target
+_RECOVERYDIR=/run/cos/recovery
+_RECOVERYSQUASHFS=${_ISOMNT}/recovery.squashfs
+_GRUBCONF=/etc/cos/grub.cfg
```
mudler, line 13: This one is configurable by a user. Shall we maybe prefix all of the ones that are configurable?

For backward compatibility we should also respect `GRUBCONF` in this case too.

packages/installer/cos.sh
```diff
@@ -4,17 +4,16 @@ set -e
...
 
 # Default size (in MB) of disk image files (.img) created during upgrades
-DEFAULT_IMAGE_SIZE=3240
+_DEFAULT_IMAGE_SIZE=3240
```
mudler, line 16: This one is intended to be tweakable too

packages/installer/cos.sh
```diff
@@ -539,47 +545,47 @@ validate_device()
...
-        CURRENT=passive.img
+    _COS_PASSIVE=$(blkid -L COS_PASSIVE || true)
+    if [ -n "$_COS_PASSIVE" ]; then
+        _CURRENT=passive.img
     fi
 
-    if [ -z "$CURRENT" ]; then
+    if [ -z "$_CURRENT" ]; then
         # We booted from an ISO or some else medium. We assume we want to fixup the current label
         read -p "Could not determine current partition. Do you want to overwrite your current active partition? (CURRENT=active.img) [y/N] : " -n 1 -r
```
mudler, line 574: See here. `CURRENT` can actually be used to avoid prompting and select the target image. Maybe this one needs a prefix too.

packages/installer/cos.sh
```diff
@@ -735,30 +753,31 @@ ensure_dir_structure() {
 }
 
 luet_args() {
+    local args
     args="--enable-logfile --logfile /tmp/luet.log"
-    if [ -z "$VERIFY" ] || [ "$VERIFY" == true ]; then
+    if [ -z "$_VERIFY" ] || [ "$_VERIFY" == true ]; then
```
mudler, line 758: Those can be specified via config as well. It's a way to have defaults.

A specific prefix as well here makes sense

mudler (changes requested):
We have to be a bit careful here. There are variables that are meant to be configured (e.g. the one we have in our default example configuration file), and some which can be specified via CLI, e.g. `CURRENT` while selecting default targets (mostly used in scripting). 

Besides, we have to offer some backward compatibility or we are going to break user configurations default.

packages/installer/cos.sh
Itxaka: That is true.

My idea is to move the configurable ones to be set on the `load_config` and use only internal ones. So no touching any external ones or using them at all, so they only entrypoints for the script are the switches or the supported configs mentioned under https://rancher-sandbox.github.io/cos-toolkit-docs/docs/customizing/general_configuration/

packages/installer/cos.sh
```diff
@@ -40,7 +35,102 @@ fi
 
 ## COMMON
 
+load_full_config() {
```
Itxaka, line 38: tested this manually to check the order:

```
$ ./cos.sh upgrade
where is verify coming from: config
$ VERIFY=ENV ./cos.sh upgrade
where is verify coming from: ENV
$ VERIFY=ENV ./cos.sh upgrade --no-verify
where is verify coming from: flag
$ ./cos.sh upgrade --no-verify 
where is verify coming from: flag
```

Message output was a temp change to set the proper value on the `_VERIFY` var, print that and exit

packages/installer/cos.sh
Itxaka: should be fixed by the change below

packages/installer/cos.sh
Itxaka: should be fixed by the change below

packages/installer/cos.sh
```diff
@@ -539,47 +691,47 @@ validate_device()
...
-        CURRENT=passive.img
+    _COS_PASSIVE=$(blkid -L COS_PASSIVE || true)
+    if [ -n "$_COS_PASSIVE" ]; then
+        _CURRENT=passive.img
     fi
 
-    if [ -z "$CURRENT" ]; then
+    if [ -z "$_CURRENT" ]; then
         # We booted from an ISO or some else medium. We assume we want to fixup the current label
         read -p "Could not determine current partition. Do you want to overwrite your current active partition? (CURRENT=active.img) [y/N] : " -n 1 -r
```
mudler, line 720: Do we need to update the help message ?

Itxaka: why? am Im missing something here? Is CURRENT able to be overwritten? 'cause its not documented

packages/installer/cos.sh
```diff
@@ -588,45 +740,38 @@ find_recovery() {
 # cos-upgrade-image: system/cos
 find_upgrade_channel() {
 
-    load_config
```
mudler, line 591: mmh too bad we don't have tests for suc-upgrade. I remember we loaded config here because otherwise the config file would override what was the user provided flags from CLI, breaking it. https://github.com/rancher-sandbox/cOS-toolkit/pull/543/files#r693834604

packages/installer/cos.sh
mudler: `    # We booted from an ISO or some else medium. We assume we want to fixup the current label
        read -p "Could not determine current partition. Do you want to overwrite your current active partition? (CURRENT=active.img) [y/N] : 
`

it is few lines below if no `CURRENT` was found. Basically that is the only way to recover other images (besides active) for e.g. when booting from a LiveCD

packages/installer/cos.sh
Itxaka: isnt this then fixed with the new order load?

packages/installer/cos.sh
Itxaka: Sorry, Im still not getting the problem here, could you elaborate further?

Is current supposed to be able to overwrite via env var?

packages/installer/cos.sh
mudler: I think it is indeed, currently having a closer look :eyes:

packages/installer/cos.sh
mudler: yep, the script tries to find a "current" image as target for the upgrade process but it is a best effort, hence why the command prompt. Since the prompt could break automatic flows (scripts, etc) it is meant to be possible to set it to avoid any interaction

Itxaka: ah, ok. Do we support the CURRENT in the config file or only via env vars?

mudler: I don't think it makes sense really sense in the config file as it is a corner-case here. Besides defaults wouldn't be relevant as we try to auto-detect it

Itxaka: updated!

mudler (approved):
Looks good!

mudler (changes requested):
right, it needs a bump of cOS packages or.. tests wouldn't run on this PR :)

Itxaka:
LOOOL

mudler (approved):

mudler:
Needs a tag of 0.7.4 before moving ahead and merging this

packages/installer/cos.sh
```diff
@@ -1242,11 +1395,19 @@ install() {
...
+    if ! is_booting_from_live && [ -z "$_COS_INSTALL_ISO_URL" ] || [ -n "$_COS_IMAGE" ] && [ -z "$_COS_INSTALL_ISO_URL" ]; then
         find_upgrade_channel
+    else
+      # Beforehand the config was only loaded if the path above was set, so we loaded some vars needed for the upgrade
+      # But now the config is loaded on start, so we need to unset some of those loaded vars if we dont need them, otherwise
+      # we can wrongly install stuff
+      unset _UPGRADE_IMAGE
+      unset _RECOVERY_IMAGE
+      # Set channel upgrades to default value
+      _CHANNEL_UPGRADES=true
```
mudler, line 1407: mmh maybe this is not necessary, as it makes sense only if we have UPGRADE_IMAGE set

mudler:
This PR is (yet) another case where https://github.com/rancher-sandbox/cOS-toolkit/issues/108 would fit in the scope instead

packages/installer/cos.sh
Itxaka: indeed, no matter if this is set or not in this path

Itxaka:
Maybe this should wait until the installer tests on #859 are expanded so we can make sure that we are covering the installer correctly and not breaking stuff. After all #797 has a workaround, so its not high priority.

Itxaka:
Blocked by #859

mudler:
Having a tagged release in between should be sufficient too, so we can always point to those artifacts if needed.. we could tackle regressions individually later

Itxaka:
> Having a tagged release in between should be sufficient too, so we can always point to those artifacts if needed.. we could tackle regressions individually later

Sounds good but...do we need to? I rather not merge broken code, and fix in this PR any issues that I can find while running the tests. Plus, we gotta add more tests anyway so...

Also helps me figure out if I missed something in the rework and maybe fix it better (like the config stuff!)

Itxaka:
as soon as https://github.com/rancher-sandbox/cOS-toolkit/pull/888 is merged, I will rebase this and see if it passes...

Itxaka:
Rebased. Lets see if those installer test also pass now....

PR #885, opened 2021-11-24
Fixups for odroidc2 example

mudler:
Add symlink to vmlinuz and reduce image size by running zypper cc.
It also override default layout (don't require COS_PERSISTENT) and disable requiring network on boot.
Drops ubuntu sample - it's not working, and green one it's enough, hooks the image creation to the ci

Next:
- Annotate generated docker images and publish them during release


.github/workflows/examples_publish.yaml
```diff
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - master
+      - odroid_c2
```
mudler, line 8: Gotta remove this after it passes :)

.github/workflows/examples_publish.yaml
```diff
@@ -68,4 +71,34 @@ jobs:
...
+             echo "COS_VERSION=$(cos_version)" >> $GITHUB_ENV
+      - name: Build Image
+        id: build_image
+        run: |
+          mkdir build
+          sudo ./images/arm-img-builder.sh --docker-image ${{ steps.prep.outputs.current }} build/cOS-example-${{ matrix.tag }}$COS_VERSION.img
+      - name: Upload Image
+        run: |
+          sudo tar cvf build.tar build
+          sudo luet util pack quay.io/costoolkit/examples:${{ matrix.tag }}$COS_VERSION.img build.tar image.tar
```
mudler, line 87: This creates a docker image for the raw `img` file. I suspect those files are too big to be released over Github, so we might end up just distributing images that can be pulled locally

Itxaka (commented):
looks good.

the artifact name seems to have a missing field `odroid-c2--image`
Cant the image be compressed somehow? maybe convert it to qcow? I know its not the same format but they are easily interchanged and can results in a much smaller image, with some docs on how to decompress it should be ok

Itxaka:
Maybe we should call the workflow differently? Like this are device example images no? As to not mistake them with our own docker example images

.github/workflows/examples_publish.yaml
mudler: It passed here: https://github.com/rancher-sandbox/cOS-toolkit/runs/4314722836?check_suite_focus=true

examples/odroid-c2/files/system/oem/01_layout.yaml
```diff
@@ -0,0 +1,9 @@
+name: "Immutable Layout"
+stages:
+  rootfs:
+    - if: '[ ! -f "/run/cos/recovery_mode" ]'
```
Itxaka, line 4: Maybe a comment here on `why` we are doing this would be good for future us.

Itxaka (approved):

mudler:
> looks good.
> 
> the artifact name seems to have a missing field `odroid-c2--image` 

Where? I see it being uploaded as:

```
 build/cOS-example-odroid-c2-0.7.4-5-gb0a380c9.img.sha256
build/cOS-example-odroid-c2-0.7.4-5-gb0a380c9.img
```

> Cant the image be compressed somehow? maybe convert it to qcow? 

Why qcow? we want to provide images that can be directly used for SD card etc

> I know its not the same format but they are easily interchanged and can results in a much smaller image, with some docs on how to decompress it should be ok

We could apply xz, but I avoided as the last step of the release already compress them. Besides the container image which is being uploaded is compressed already (see ` odroid-c2-0.7.4-5-gb0a380c9.img ` in https://quay.io/repository/costoolkit/examples?tab=tags)

Itxaka (approved):

examples/odroid-c2/files/system/oem/01_layout.yaml
mudler: Addressed in https://github.com/rancher-sandbox/cOS-toolkit/pull/885/commits/3ba3bf43dcb161304218e48129044cdf21c97997 !

mudler:
No CI jobs will run for this as it runs just on master and its a new pipeline, merging :t-rex:

Issue #887, opened 2021-11-25
Create multiple partitions might fail


mudler:
Seems there is a flakyness while creating multiple repository in sequence in a disk. Seems `yip` fails to find the partitions after creation 

**cos-toolkit version:**
Applies to yip from https://github.com/rancher-sandbox/cOS-toolkit/commit/108a64bcf7ed8ee106d8aa770982871e7c7a2c0c

**Additional context**
https://github.com/harvester/harvester/issues/1583

Itxaka:
gofmthttps://github.com/mudler/yip/pull/27

mudler:
Reopening as we are still seeing this issue

Itxaka:
https://github.com/mudler/yip/releases/tag/0.9.25 should fix this

Itxaka:
Harvester folks closed their issue as resolved, they tested positive and negative with great results, so closing this.

https://github.com/harvester/harvester/issues/1583

PR #889, opened 2021-11-25
bump yip and dependants

Itxaka:
New yip version uses  retries for re-reading the partitions so it doesn't fail or skip on the first run.

Also adds sync into the inmutable rootfs

Fixes: https://github.com/rancher-sandbox/cOS-toolkit/issues/887

### DIFF
--- a/packages/immutable-rootfs/30cos-immutable-rootfs/module-setup.sh
+++ b/packages/immutable-rootfs/30cos-immutable-rootfs/module-setup.sh
@@ -29,7 +29,7 @@ install() {
     # Include utilities required for cos-setup services,
     # probably a devoted cos-setup dracut module makes sense
     inst_multiple -o \
-        partprobe lsblk sgdisk mkfs.ext2 mkfs.ext3 mkfs.ext4 mkfs.vfat mkfs.fat mkfs.xfs blkid e2fsck resize2fs mount xfs_growfs umount
+        partprobe sync lsblk sgdisk mkfs.ext2 mkfs.ext3 mkfs.ext4 mkfs.vfat mkfs.fat mkfs.xfs blkid e2fsck resize2fs mount xfs_growfs umount
     inst_hook cmdline 30 "${moddir}/parse-cos-cmdline.sh"
     inst_script "${moddir}/cos-generator.sh" \
         "${systemdutildir}/system-generators/dracut-cos-generator"
--- a/packages/immutable-rootfs/build.yaml
+++ b/packages/immutable-rootfs/build.yaml
@@ -13,7 +13,7 @@ copy:
 - package: 
     category: "toolchain"
     name: "yip"
-    version: ">=0"
+    version: ">=0.9.23"
   source: "/usr/bin/yip"
   destination: "/usr/bin/yip"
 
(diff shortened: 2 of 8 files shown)


packages/installer/cos.sh
```diff
@@ -192,7 +192,7 @@ prepare_passive() {
 
 part_probe() {
     local dev=$1
-    partprobe ${dev} 2>/dev/null || true
+    blockdev --rereadpt ${dev} 2>/dev/null || true
```
Itxaka, line 195: This may also fix the issues I was seeing on the tests

mudler (approved):
Looks good here!

packages/installer/cos.sh
```diff
@@ -405,7 +405,6 @@ get_iso()
 get_image()
 {
     if [ -n "$UPGRADE_IMAGE" ]; then
-        part_probe
```
Itxaka, line 408: @mudler any idea why was this here? This did nothing as there was no `$dev` passed to the call, but didnt fail due to the `|| true` in the function.

mudler: looks like a leftover, `blkid_probe` was actually probably meant here

packages/immutable-rootfs/30cos-immutable-rootfs/module-setup.sh
```diff
@@ -29,7 +29,7 @@ install() {
     # Include utilities required for cos-setup services,
     # probably a devoted cos-setup dracut module makes sense
     inst_multiple -o \
-        partprobe lsblk sgdisk mkfs.ext2 mkfs.ext3 mkfs.ext4 mkfs.vfat mkfs.fat mkfs.xfs blkid e2fsck resize2fs mount xfs_growfs umount
+        blockdev lsblk sgdisk mkfs.ext2 mkfs.ext3 mkfs.ext4 mkfs.vfat mkfs.fat mkfs.xfs blkid e2fsck resize2fs mount xfs_growfs umount
```
mudler, line 32: I just thought at it. But this is the kind of changes that forces the package to depend on a version of yip >= 0.9.22 (in runtime)

Itxaka: good point! wait for tests to pass then addd it + force-merge? 

Not only the installer but also the inmutablefs package should have a minimum yip dependency as part of this

packages/immutable-rootfs/build.yaml
```diff
@@ -13,7 +13,7 @@ copy:
 - package: 
     category: "toolchain"
     name: "yip"
-    version: ">=0"
+    version: ">=0.9.22"
```
mudler, line 16: This should go in the collection.yaml file of the package as it is a runtime deps, not really a build dep (as we don't run any `yip` command inside the `build.yaml` file

Itxaka: woops, sorry, screwed it up due checking that raw image test issue, will fix it later after checking that one

Itxaka: wait, isnt it necessary during build so it can be using during module-setup.sh? thats what it says on the comment no? what would be the difference here on build VS runtime?

PR #894, opened 2021-11-26
Make OEM partition label customizable at build time

davidcassany:
This commit sets the basis to make OEM partition label customizable.
Currently it can only be customized a build time, it already was,
however the value was hardcoded in several places. Now the label
value is included in /etc/cos/config file and all other packages
read it from there at build time and install time.

In the future it should not be complex to make it configurable at
install time. For that some little extensions to the installer are
required.


images/arm-img-builder.sh
```diff
@@ -183,6 +188,18 @@ while [ "$#" -gt 0 ]; do
     shift 1
 done
 
+if [ -z "$cos_config"]; then
+  echo "No cos-config file specified"
```
mudler, line 192: Maybe sticking to default instead of existing here? I guess it's fine to keep the current labels as default ( so no need to pass by a default config each time)

images/img-builder.sh
```diff
@@ -2,8 +2,31 @@
...
 MANIFEST=$1
-if [[ -z "${MANIFEST}" ]]; then
-  echo "Manifest required as parameter, cannot continue"
+if [ -z "${MANIFEST}" ]; then
+  >&2 echo "Manifest required as parameter, cannot continue"
+  exit 1
+fi
+
+CONFIG_FILE=$2
+if [ -z "${CONFIG_FILE}" ]; then
```
mudler, line 11: same here I guess, would make sense to ship with sane defaults that can be override to avoid too many inputs from the user at start. If they want more fine-grained control they can still by supplying an optional config

packages/cloud-config/oem/00_rootfs.yaml.tmpl
```diff
@@ -17,14 +17,14 @@ stages:
       name: "Layout configuration"
       environment_file: /run/cos/cos-layout.env
       environment:
-        VOLUMES: "LABEL=COS_OEM:/oem LABEL=COS_PERSISTENT:/usr/local"
+        VOLUMES: "LABEL=@OEM_LABEL@:/oem LABEL=@PERSISTENT_LABEL@:/usr/local"
```
mudler, line 20: :100:

images/arm-img-builder.sh
davidcassany: I coded it to be consistent with the other options and because I am not sure in which context this script it used, so I am not sure what would be a sane default cos-config file value ( it could be `packages/cos-config/cos-config`) but I feels odd to me here using some relative path making the script dependent on the working dir. For the x86 script this is solved at Makefile level. The makefile manages the default.

mudler: Sorry my bad I wasn't clear! I mean defaults for `${ACTIVE_LABEL}` , `${PASSIVE_LABEL}`, etc so the config file is not mandatory, but rather optional overriding defaults values (`like COS_ACTIVE` for `ACTIVE_LABEL` and so on)

davidcassany: I am not really a fan of this idea, to be honest, since this means going back to hard coded values and having the same default written in more than one place, but I could live with it as long as our CI does not makes use of them and we enforce passing the config file, which is already the case for the x86 builder. Gonna add a commit for it.

Anyway I hope we can get rid of all this once we have defaults defined within the elemental client and some structured configuration files to alter them. Then these scripts should only make use of the elemental client.

mudler: yep, I do agree it's not ideal now that we have all the defaults fragmented, but with the new cli this is going to be eased out quite a lot as we will have a central place (so no dups).

The idea behind the arm script is that we can re-use them as a base for the user-facing ones in the future, and ideally the minimum input from the user would be an image (at the moment it's hardcoded to the odroid, but should probably just point to the generic image instead) to get started

mudler (approved):
Looks good here! nice!

Itxaka (approved):

PR #900, opened 2021-11-29
Add packages

mudler:
Adding packages while I tried to get OdroidC2 working locally. Sadly I don't get eth0 after boot (altought, I can see u-boot trying to get an IP )  so I cannot test upgrades, resets and the full cOS featureset. I'm suspecting kernel, but requires quite some time to experiment.  This change adds firmware packages, firmware and dtbs to the example image

Another approach would be to test with the u-boot and old kernel, but would be suboptimal to release with those ( for the records, here is a spec that build the Odroid kernel https://github.com/rancher-sandbox/cOS-toolkit/commit/c31ebbd1ac151883b49a74fbf90955c38841eb3f).

Also, this PR adds packages to the toolchain image in order to run the arm image script successfully. For reference, that's where I was experimenting with : https://github.com/mudler/cos-embedded-images/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM opensuse/leap
 
 ENV LUET_NOLOCK=true
 
-RUN zypper in -y docker curl squashfs xorriso make which mtools dosfstools jq gptfdisk
+RUN zypper in -y docker curl squashfs xorriso make which mtools dosfstools jq gptfdisk git parted kpartx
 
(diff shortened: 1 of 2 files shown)


Itxaka (approved):

davidcassany (approved):

mudler:
shouldn't affect CI, and OdroidC2 images are built only on master.. merging!

PR #901, opened 2021-11-29
bump yip

Itxaka:


packages/cos-setup/definition.yaml
```diff
@@ -1,7 +1,7 @@
 name: cos-setup
 category: system
-version: 0.4-12
+version: 0.4-13
```
mudler, line 3: Let's leave revbumps for the bot unless strictly necessary

```suggestion
version: "0.5"
```

mudler (approved):

PR #906, opened 2021-11-30
Build cos package by stacking packages rather than build envs

davidcassany:
This commit changes the build approach for cos package. Instead of
stacking build environments for the build it stacks all dependency
packages. This allows control over the distinction of build time
requirements and run time requirements.

### DIFF
--- a/packages/cos/build.yaml
+++ b/packages/cos/build.yaml
@@ -1,6 +1,11 @@
 #### system/cos system/cos-container and recovery/cos start here
 {{ if or (eq .Values.name "cos-container") (eq .Values.name "cos" ) }}
 requires:
+# Base distro is not required by other packages to facilitate their usage
+# as stand alone packages (e.g. in Dockerfiles), so that it is included here.
+- name: "base"
+  category: "distro"
+  version: ">=0"
 - name: "cos-setup"
   category: "system"
   version: ">=0"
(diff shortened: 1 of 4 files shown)


davidcassany:
I did this PR to raise an issue I found during #894 work. In that mentioned PR I took care to clearly divide between build time requirements and run time requirements. My idea is that the templating script should not be part of the base system since the current structure does not allow to modify any of the templated files at runtime. However I noted that `cos` package is build up on top of the stacked build environments of its dependencies, so basically buildtime dependencies of listed packages are also included as part of the image. I consider this a limitation and also error prone (we might easily confuse buildtime and runtime deps with this approach).

I am not sure if there is any explicit reason to build `cos` the way we do now. Since these are few lines of code and there is a simple example of the effects of it I thought to better discuss the convenience of it or not over the PR itself. So that the `Do not merge` label.

mudler:
Looks good here!
 
> I am not sure if there is any explicit reason to build `cos` the way we do now. Since these are few lines of code and there is a simple example of the effects of it I thought to better discuss the convenience of it or not over the PR itself. So that the `Do not merge` label.

I don't think there is a specific reason except that the feature was introduced later, before the cOS packages have been written

mudler (approved):

mudler:
I think similarly we can also drop the `copy` part in favor of a single requires block

Itxaka (approved):

davidcassany:
> I think similarly we can also drop the `copy` part in favor of a requires block

I agree, lets do that in a follow up PR if the current managed to pass tests.

packages/cos/build.yaml
```diff
@@ -1,6 +1,11 @@
 #### system/cos system/cos-container and recovery/cos start here
```
mudler, line 1: Needs a bump to the cos versions in the collection file too

Issue #757, opened 2021-10-05
Document core packages


mudler:
**Is your feature request related to a problem? Please describe.**
A way for users to identify the cos core packages and the features they provide

**Describe the solution you'd like**
A documentation page, or a set of pages that describe the core packages (`meta/core`) and the features they provides contextualized with the docs.

Would be preferable to keep specific package documentation together the package sources.

mudler:
Starting to thinker with it:

Our packages already support a `description` field, which main usage is to be displayed over CLI and the [package browser](https://rancher-sandbox.github.io/cos-toolkit-package-browser/cos-toolkit-green/).

For specific documentation, seems golang yaml support for reading comments is very limited https://ubuntu.com/blog/api-v3-of-the-yaml-package-for-go-is-available and seems also a route to a very complex task. 

I would go over creating something simple, living in cos-docs which generates `md file` from a template over this repository content. 
Preferably, for easy of maintaining/writing it should be reading from a specific field in the `definition.yaml` files, for e.g. `documentation` (which could contain valid md/html code), so we have:

```yaml
name: "foo"
category: "bar"
version: "1.0"
description: "What this package do in simple words (keep it under 100 chars)"
documentation: |
                          ## Package usage
                          Here goes a long detailed description on what the package provides and how to use it
```

In this way we can re-use the information already present in the `definition.yaml` file to to generate the appropriate page name and metadata (package name, category, etc)

PR #907, opened 2021-11-30
Add packages documentation

mudler:
This is just adding some placeholders, I'll add up docs but wanted to have all in place and see how it looks like
 before going ahead

Part of: https://github.com/rancher-sandbox/cOS-toolkit/issues/757


davidcassany:
@mudler JFYI in #909 I am doing little cleanup on packages deps and I saw there is not need for a collection here as we are building a single package. I moved the `collection.yaml` to a `definition.yaml`.

mudler:
yep, noticed the same here indeed, no worries, I can re-adapt after your PR is merged :+1:

mudler:
ok, reworked out so it just applies the README.md file and leaves the specs untouched so it's easier to rebase later

mudler:
Just moved some docs and added placeholders. merging because it is tight with the other PR, and docs were now missing from the website :sweat_smile:

PR #910, opened 2021-12-01
ci: Drop autobump.revbump_related from specs

mudler:
This is not required anymore as now revdeps should be clearly displayed
also for specs that requires final packages during build time.

I think this should also fixups revbumps versions from the bot


Itxaka (approved):
Niiice

mudler:
Doesn't have an impact on build, merging

Issue #917, opened 2021-12-02
Too many CI failures


mudler:
**cos-toolkit version:**
Since at least https://github.com/rancher-sandbox/cOS-toolkit/commit/a684abb22e8b91dda22428c14fee4f6b466d203c

**CPU architecture, OS, and Version:**

**Describe the bug**
CI jobs are sporadically failing more frequently, preventing time to time to go for the full flow. we should look after into optimizing the jobs, or relax our test suites (maybe splitting?)

**To Reproduce**
There was no full green CI run since the commit above

**Expected behavior**

**Logs**

**Additional context**

davidcassany:
In addition to some installer tests failures we also started having having failures due to not having enough disk space during the build.

Itxaka:
Installer tests can be splitted in different paths with the use of regex, but that would take precious macosX instances (we have a total of 5 in parallel)

For now, the installer tests are NOT required for a job to pass and be merged, so there is that :/

mudler:
not really related, but this workaround will make things worse: https://github.com/rancher-sandbox/cOS-toolkit/pull/920. Maybe we shall fallback to job cancelling via action? e.g.

```
jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - name: Cancel Previous Runs
        uses: styfle/cancel-workflow-action@0.9.1
        with:
          access_token: ${{ github.token }}
      #- name: Run Tests
      #  uses: actions/setup-node@v1
      #  run: node test.js
      # ... etc
```

mudler:
One failure is related to openSUSE upgrades in the base image:

```

 INFO   #9 41.58 Resolving package dependencies... INFO   
 INFO   #9 41.63 2 Problems: INFO   
 INFO   #9 41.63 Problem: the installed shadow-4.8.1-2.43.x86_64 requires 'libsemanage.so.1()(64bit)', but this requirement cannot be provided INFO   
 INFO   #9 41.63 Problem: the to be installed container-selinux-2.171.0-lp153.17.6.noarch requires 'selinux-policy >= 20211111-lp153.114.6', but this requirement cannot be provided INFO   
 INFO   #9 41.63  INFO   
 INFO   #9 41.63 Problem: the installed shadow-4.8.1-2.43.x86_64 requires 'libsemanage.so.1()(64bit)', but this requirement cannot be provided INFO   
 INFO   #9 41.63  Solution 1: Following actions will be done: INFO   
[...]

Issue #278, opened 2021-06-15
ci: add self-hosted runners


mudler:
We are at capacity with the GHA concurrent jobs limits, and this slows down development quite a lot.

Let's see if we can configure AWS spot instances  or w/e else can provide a bunch of runner to be run in this repository.

e.g. by following among the lines of https://github.com/philips-labs/terraform-aws-github-runner to run our tests on. We have also to figure out if we can run vbox (or qemu) to run our test suite on top.

kkaempf:
A self-hosted runner [is ready](https://github.com/rancher-sandbox/cOS-toolkit/settings/actions/runners). I just need to boot it.

mudler:
This issue is about having more than one. We run quite some jobs in parallel, and it would quickly become another bottleneck

mudler:
@Itxaka I'll take care of the single node only setup then :+1:

Itxaka:
I will have a look at some heat templates to easily add/remove ecp self-hosted runners

mudler:
@Itxaka added a sample k8s deployment of gh runner here: https://github.com/rancher-sandbox/cOS-toolkit/wiki/Github-runner-on-k8s

Itxaka:
This is a POC of a script to deploy github runners in ECP: https://github.com/rancher-sandbox/cOS-toolkit/pull/304

Tested and working, see readme for details

Works...ok-ish. Workflows would need some adaptations to fully work, user-data migth need also adaptations if we want to use this, but it makes no sense to develop it further.

 - Need to remove the build dir on each job as it can be left over from a previous job
 - No vagrant/qemu/packer installed by default, so not all jobs can run.
 - We can add those packages and run everything in the same base, no more osX required for qemu builds
 - We could have workers based on size and OS and then use the labels to run the jobs on them, still requires more config and work to the deployment scripts and workflows.
 - We could have a "master " node that receives one job, creates an on demand instance with this scripts and uses it for one job, deleting it afterwards.
   - Makes it behave more like github runners. 
   - Requires 1 "master" machine to sync everything
   - Should be easy to automate the creation, deletion, token adquisition, etc...
   - Requires investing in automation
   - Requires creating custom images with software preinstalled to avoid 5/10 minute machine boot and configuration per job
 - We could also have a big node that spawns several docker containers with the workers.
   - Allows us more flexibility
   - Supports wide range of base os
   - REALLY easy to grow or decrease
[...]

PR #934, opened 2021-12-06
ci: add cleanup images plugin

mudler:
Attempt to free up space during build (related to #917 )

### DIFF
--- a/.github/build.sh
+++ b/.github/build.sh
@@ -1,3 +1,6 @@
 #!/bin/bash
 
 sudo -E make PACKAGES="$1" build
+
+docker images --filter="reference=$REPO_CACHE" --format='{{.Repository}}:{{.Tag}}' | xargs -r docker rmi --force
+docker image prune --force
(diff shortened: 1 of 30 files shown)


mudler:
Rebased on top of #935

Itxaka (commented):
Yikes. Can't we simply docker system prune -a and call it a day? That should clean up everything not being used no?

Also, what do we expect to clean up from this? Aren't ci nodes spin up from scratch on each job?

mudler:
> Yikes. Can't we simply docker system prune -a and call it a day? That should clean up everything not being used no?
> 
Yes and no, I mean we definitely can delete all images after we finish building one package, the plugin is just a safeguard to avoid going to look at this again

> Also, what do we expect to clean up from this? Aren't ci nodes spin up from scratch on each job?
Nope, we run 1 job building all images, rationale is explained #917 . we can split build into multiple, but that has other problems as well (starvation)

Itxaka (approved):

mudler:
I think with #278 we can disable the plugin and the cleanup, and just run it after the whole build.. but keeping it for now until we have self-hosted runners

PR #953, opened 2021-12-09
Bump to leap based rpi4 firmware

### LINKED ISSUE #940: Firmware not working on rpi4 (cos not booting)
When booting on rpi4 u-boot and grub are properly booting, but the process gets stuck at:
```
EFI stub: Exiting boot services and installing virtual address map...
```

I assume this is due to a firmware bump on the repo.

dragonchaser:
Fixes #940

### DIFF
--- a/packages/firmware/build.yaml
+++ b/packages/firmware/build.yaml
@@ -20,7 +20,7 @@ package_dir: "/data/boot"
 
 {{if eq .Values.name "u-boot-rpi64"}}
 - mkdir -p /data/{boot,usr}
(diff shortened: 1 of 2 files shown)


mudler (approved):

mudler:
builds fine https://github.com/rancher-sandbox/cOS-toolkit/runs/4470557668?check_suite_focus=true, no reason to wait amd64 tests :)

kkaempf:
I can confirm that this is booting successfully on RPi4. But the _kernel_ fails on RPi3 after ~20 secs into boot at `pinctrl-bcm2835 3f200000.gpio: could not add GPIO chip`.

Also serial console is not enabled in the kernel :-(

PR #967, opened 2021-12-14
Increase the default timeout for early boot /oem mount

davidcassany:


davidcassany:
This is related to https://github.com/harvester/harvester/issues/1667#issuecomment-993130843 
Setting the default to 2min has the benefit to apply to a wider range of hardware and the only counter part is that deployments without using `/oem` mount could be penalized by 2min at boot time, which can be fixed by simply applying a custom timeout at kernel params:
https://rancher-sandbox.github.io/cos-toolkit-docs/docs/reference/packages/system-immutable-rootfs/#kernel-configuraton-paramters

mudler (approved):

Itxaka (approved):

PR #991, opened 2021-12-17
rpi: update example config file

mudler:
labels are explictly required since https://github.com/rancher-sandbox/cOS-toolkit/pull/894


davidcassany (approved):
Oh yes, thats true. This config file is explicitly required. I hope to get rid if this on elemental v2.0 😉

mudler:
Merging as doesn't affect CI

PR #994, opened 2021-12-20
Add manual workflows for force pushing/publishing the latest artifacts

Itxaka:
Adds 2 new workflows which are exclusively manually triggered, that can build and publish the latest artifacts, without running thorough the full pipeline, so they should be faster and allow us to retrigger the publish in case something happened during the final steps (quay failure, net issues, jobs cancelled, signatures failed, etc...)

It also allows to set the --force-push flag so artifacts can be republished and resigned.

Not sure about the current switches thougth, I think currently it should just push the packages under the packages dir? Im not clear on that part, should probably need a couple of manual runs to verify :D

### DIFF
--- a/.github/workflows/publish_arm64.yaml
+++ b/.github/workflows/publish_arm64.yaml
@@ -1,0 +1,148 @@
+name: publish
+on:
+  workflow_dispatch:
+    inputs:
+      flavor:
+        required: false
+        default: green
+        type: choice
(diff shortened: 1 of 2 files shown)


Itxaka:
canceling jobs as this do not affect any jobs so no reason to have the queue running them

mudler (approved):

PR #995, opened 2021-12-20
Bind mount initrd's tmpfs /tmp at initramfs stage

davidcassany:
This commit bind mounts /tmp from initrd environment to initramfs stage
environment (chroot on /sysroot). This enables the use of tools or
utilites in initramfs stage that relay on a writable /tmp.

Note this is an ephemeral filesystem that gets lost at switching root
stage, so right after initramfs stage.

### DIFF
--- a/packages/cos-setup/definition.yaml
+++ b/packages/cos-setup/definition.yaml
@@ -1,6 +1,6 @@
 name: cos-setup
 category: system
-version: "0.6"
+version: "0.6.1"
 requires:
   - name: "elemental-cli"
     category: "toolchain"
--- a/packages/cos-setup/cos-setup-initramfs.service
+++ b/packages/cos-setup/cos-setup-initramfs.service
@@ -7,7 +7,7 @@ Before=initrd.target
 
 [Service]
 RootDirectory=/sysroot
-BindPaths=/proc /sys /dev /run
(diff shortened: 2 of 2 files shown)


packages/cos-setup/definition.yaml
```diff
@@ -1,6 +1,6 @@
 name: cos-setup
 category: system
-version: "0.6"
+version: "0.6.1"
```
Itxaka, line 3: requires reverse bump?

Itxaka (approved):

packages/cos-setup/definition.yaml
davidcassany: mmm reverse bump... 😓 Probably because it's Monday morning, but I don't get what are you referring to... I just wanted to make a version bump, I though this was a patch level change but when I wanted to update the version I noticed this package not follow the common semver schema of MAJ.MIN.PTCH, so I added the patch level number.

Itxaka: I meant bumping the cos packages so they include this one, otherwise all the cos-system, cos-img, etc... packages are using the old cos-setup version unless we bump those no?

davidcassany: Oh, I see, I think you are right.

PR #996, opened 2021-12-20
bump luet-cosing + cosign

Itxaka:

### DIFF
--- a/packages/meta/collection.yaml
+++ b/packages/meta/collection.yaml
@@ -83,7 +83,7 @@ packages:
     requires:
       - category: toolchain
         name: cosign
-        version: "<1.4"
+        version: ">1.3.1" # signatures generated with >1.3.1 are not validated with versions <= 1.3.1
       - category: toolchain
         name: luet-cosign
         version: ">=0"
--- a/packages/cos/build.yaml
+++ b/packages/cos/build.yaml
@@ -52,7 +52,7 @@ requires:
   version: ">=0"
 - name: "cosign"
   category: "toolchain"
-  version: "<1.4"
+  version: ">1.3.1"
 - name: "luet-cosign"
   category: "toolchain"
   version: ">=0"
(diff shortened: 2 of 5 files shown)


packages/meta/collection.yaml
```diff
@@ -83,7 +83,7 @@ packages:
     requires:
       - category: toolchain
         name: cosign
-        version: "<1.4"
+        version: ">=1.4"
```
mudler, line 86:
```suggestion
        version: ">1.3.1"
```

mudler: Maybe we should also add a comment on why? (and link to the issues.. so we don't forget in the future?)

Itxaka: Trying to see first why locally validation passes but not in ci...

mudler:
@Itxaka [master](https://github.com/rancher-sandbox/cOS-toolkit/runs/4582855173?check_suite_focus=true) seems back to failing on publishing signatures, maybe the luet-plugin in the tree is incompatible with cosign 1.3.x?

Itxaka:
> @Itxaka master seems back to failing on publishing signatures, maybe the luet-plugin in the tree is incompatible with cosign 1.3.x?

weird, the fulcio-url is in 1.3.1 .... and the TUF_ROOT as well...

https://github.com/sigstore/cosign/commit/d401496d12bb592c5247d9414faea4985696e00c
https://github.com/sigstore/cosign/commit/bfd42e554d60b732a6403b7111ac840942e84572

Itxaka:
force merged https://github.com/rancher-sandbox/cOS-toolkit/pull/999 to get the logs

Itxaka:
That is still using 0.8 thougth, so it may have the same issue of parallel stuff accessing the same TUF_ROOT

Itxaka:
Makes no sense to wait for this anymore as it wont affect anything in this pipeline once it passes the build target.

PR #1002, opened 2021-12-21
Set concurrency to 1 for publish artifacts

Itxaka:
Temporary set concurrency to one while pushing artifacts to avoid
hitting a concurrency issue on luet-cosign


Itxaka:
should not affect PR workflows

mudler (approved):

PR #1004, opened 2021-12-21
Use latest luet from script in the publish job

Itxaka:
github workers are symlinking /bin to /usr/bin/ which results in luet
identifying the luet-cosign plugin twice and calling it twice which
results in failure to sign the artifacts due both plugins accessing the
same files at the same time.

This makes it so luet is installed from the get_luet.sh script directly,
which would install the latest version, which has a fix to have unique
plugin names in the go-pluggable lib[0]

This commit is safe to revert once the luet and luet-cosign packages
have been published as there contain fixes for this issue.

[0] https://github.com/mudler/go-pluggable/commit/9263b05c562e6207999fd74bebf2cf36fa013e02

### DIFF
--- a/.github/build.yaml.gomplate
+++ b/.github/build.yaml.gomplate
@@ -529,6 +529,7 @@
       COSIGN_EXPERIMENTAL: 1 # use keyless signing
       COSIGN_REPOSITORY: raccos/releases-{{{ $flavor }}}
       PUBLISH_ARGS: "--concurrency 1 --plugin luet-cosign"
+      LUET_INSTALL_FROM_COS_REPO: false
     steps:
       {{{ tmpl.Exec "prepare_build" }}}
       {{{ tmpl.Exec "prepare_worker" }}}
--- a/.github/workflows/build-master-blue-arm64.yaml
+++ b/.github/workflows/build-master-blue-arm64.yaml
@@ -118,6 +118,7 @@ jobs:
       COSIGN_EXPERIMENTAL: 1 # use keyless signing
       COSIGN_REPOSITORY: raccos/releases-blue
       PUBLISH_ARGS: "--concurrency 1 --plugin luet-cosign"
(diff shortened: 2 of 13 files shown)


Itxaka:
wont affect PR pipeline

davidcassany (approved):
I do not follow all the related details of this PR, but looks good to me.

.github/build.yaml.gomplate
```diff
@@ -529,6 +529,7 @@
       COSIGN_EXPERIMENTAL: 1 # use keyless signing
       COSIGN_REPOSITORY: raccos/releases-{{{ $flavor }}}
       PUBLISH_ARGS: "--concurrency 1 --plugin luet-cosign"
+      LUET_INSTALL_FROM_COS_REPO: false
```
davidcassany, line 532: This causes to install the latest version of luet? Am I getting it right?

Itxaka: So when we use `make deps` we use the scritp `get_luet.sh` to obtain the latest luet from the mudler/luet repo directly and use that + our repo to bootstrap the luet install, so we end up with luet installed from our releases-green repo and the releases-green repo configured in the luet.cfg file.

That proved difficult to use with a newly created arch/flavor as the repos on the first run will be empty, so  I modified the script to be able to just use the luet we download and copy that into the system with no repos configured[0]. So instead of downloading luet and installing luet from our cos repo, this other path will download luet and copy luet into the running system directly. This also bypasses whatever the version is on our repo and installs the latest (or you can override it with `LUET_VERSION` I think)


[0] https://github.com/rancher-sandbox/cOS-toolkit/commit/d4a2502af6d419f1b3ed9e25b252abe15dbf8fbd#diff-362156cbd5c315eb0e08dad196ecf6f0cf215fd71066c276ee46e0a624f5f647R17

PR #1011, opened 2021-12-22
Add signer to CI to resign artifacts

Itxaka:


Itxaka:
this **does** affect the PR workflow as it changes the build.go command to build so at least it should pass that part and compile it properly.

mudler (approved):
Nice! :100:

Itxaka:
Its tested until the signing, as I dont have the keys for the repo, nor the local OIDC support to test that part, but it should work as it uses the same commands as in luet-cosign.


Yes, we could use the cosign from the luet repo to do this to make sure we use the same version, but that would mean again relaying on whatever version is on our repos and we might want to play with that. Plus this should only be temporary to sign the missing version we currently have, after we should try to include this on the elemental-cli as it can help other people resign their own artifacts, even ignore the luet-cosign plugin altogether during build and do it in a  different step, or create signatures and store them in different repos, i.e. in case the repos are not available or if they want to sign to a private repository for offline access, etc...

.github/resign.go
```diff
@@ -0,0 +1,95 @@
...
+		referenceID = "repository.yaml"
+	}
+	d := installer.NewSystemRepository(types.LuetRepository{
+		Name:        "cOS",
+		Type:        "docker",
+		Cached:      true,
+		Urls:        []string{repo},
+		ReferenceID: referenceID,
+	})
+	ctx.Config.GetSystem().Rootfs = "/"
```
mudler, line 30: huh, actually if we bump the luet module now this is changed and much easier to work with

Itxaka: I dont even know what that is used for, I blindly copied it from the build.go and it didnt gave me any issues so working as intended LOL

Itxaka:
Also this could do with some goroutines to do the signing in parallel, otherwise it can be slow as hell

Itxaka:
It passed the critical part of building and using the build.go file, merging

PR #1016, opened 2021-12-22
resign: change parameters around

Itxaka:
Currently nothing is using the fulcio url but if someone uses it the order would be wrong (not according to cli help, but from testing)

Low prio


Itxaka:
superseeded by https://github.com/rancher-sandbox/cOS-toolkit/pull/1017/

PR #1017, opened 2021-12-22
resigner: trim space and ignore fulcioflag if not set

Itxaka:
Looks like the sign commadn is sensible to whitespace, so trim all the
space from the image name and only add the fulcioflag if its set to
avoid any whitespace in the command

### DIFF
--- a/.github/resign.go
+++ b/.github/resign.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"strings"
 )
 
 func getRepositoryPackages(repo string, ctx *types.Context) (searchResult client.SearchResult) {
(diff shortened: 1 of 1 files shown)


.github/resign.go
```diff
@@ -83,14 +86,17 @@ func checkAndSign(tag string, ctx *types.Context) {
...
-	args := []string{"verify", tag}
+	args = []string{"verify", tag}
 	ctx.Debug("Calling cosing with the following args:", args)
 	out, err := exec.Command("cosign", args...).CombinedOutput()
 	ctx.Debug("Verify output:", string(out))
 	if err != nil {
 		ctx.Warning("Artifact", tag, "has no signature, signing it")
-		args := []string{fulcioFlag, "sign", tag}
+		if fulcioFlag != "" {
+			args = []string{"sign", fulcioFlag, tag}
```
mudler, line 96: shall we maybe trim spaces also from `fulcioFlag`? :clown_face:

mudler (approved):

.github/resign.go
Itxaka: you joke, but I did just in case LOL

PR #1018, opened 2021-12-23
Add helm

mudler:
As I was having a look over cosign usage in os2, helm is used, and maybe it's better to have it here so we can consume it directly the signed version with along all the other packages.

Update:
Scope is to secure this part of the supply chain https://github.com/rancher/os2/blob/085bc09768be96ee90b07d23cbe653c485d08fc0/Dockerfile#L7

### DIFF
--- a/packages/utils/helm/build.yaml
+++ b/packages/utils/helm/build.yaml
@@ -1,0 +1,29 @@
+requires:
+  - name: "golang{{- if .Values.fips -}}-fips{{- end -}}"
+    category: "build"
+    version: ">=0"
+env:
+{{ template "golang_env" }}
+{{if .Values.fips}}
+- CGO_ENABLED=1
+{{ else }}
+- CGO_ENABLED=0
+- LDFLAGS="-s -w"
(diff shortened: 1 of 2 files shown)


davidcassany (approved):
Looks good to me, but, is there any reason not simply use the openSUSE package? This will become relevant if at some point if we pretend to register the system to some to some SLE.

mudler:
> Looks good to me, but, is there any reason not simply use the openSUSE package? This will become relevant if at some point if we pretend to register the system to some to some SLE.

Yes, I gave it a thought. I think we want to have fine grained control over the version (that could be fips, or just a specific pinned version) we want to select as we will retain all the versions in the cOS repos, in this way no moving pieces from repositories for such components. Besides, TW is stuck at 3.6.x.

davidcassany:
In fact helm is an utility that is clearly in our ecosystem, thus it should not be a terrible deal having to assume its maintenance under the limited scope of the reference images.

mudler:
No need for a full CI run

PR #1023, opened 2021-12-24
arm: create COS_PERSISTENT instead of COS_OEM

mudler:
Add a cloud-init also to automatically expand the COS_PERSISTENT
partition on first boot.

Testing it still


davidcassany (changes requested):
Cool, with this you are not loosing OEM functionality, you can still plug a USB having the OEM label with some custom cloud init data and it would be executed.

Note you are forgetting to update `examples/odroid-c2/files/system/oem/01_layout.yaml`

mudler:
> Cool, with this you are not loosing OEM functionality, you can still plug a USB having the OEM label with some custom cloud init data and it would be executed.
> 
> Note you are forgetting to update `examples/odroid-c2/files/system/oem/01_layout.yaml`

Good catch, I actually forgot to commit its deletion :sweat:

mudler:
Oh wait, I did commit it, I've deleted both layout files as now `cOS` defaults should be good

davidcassany (approved):

mudler:
This might need https://github.com/rancher-sandbox/cos-toolkit-docs/pull/67

mudler:
seems adding `rd.cos.oemlabel=COS_OEM` to bootargs doesn't help. Still dropping to emergency shell  from dracut

mudler:
ok, seems to boot when creating an COS_OEM (even from an .img file). So looks like entries listed in `VOLUMES` in layout.env, once specified are mandatory. I'll create a layout file to disable oem entirely

mudler:
@davidcassany needs a re-review please as things changed slightly. It looks like the OEM partition is required if specified in `VOLUMES`, although would make sense to offer a way to mount certain paths 'optionally' like `COS_OEM`, in that case we can still support the scenarios of plugging an USB stick and load cloud config files from there.

Is there a way to override the default ? by looking at the code at a quick glance I didn't noticed, but not sure if I might have missed something

mudler:
ok so after discussing with @davidcassany OEM gets mounted by default so there is no need to specify it in `VOLUMES`, mount will be attempted at boot even if not specified there. That means we can't rely on cOS defaults for ARM here, hence why the specific cloud-config file which overrides the layout

mudler:
ok now booting works. However automatic resizing of COS_PERSISTENT it doesn't. Fails silently here with an exit 2. I'd say to merge this and create a follow-up for the partition resizing as it is out of purpose (and can be workarounded easily after image creation too).

PR #1033, opened 2021-12-31
tests: bump cOS image version

mudler:
Needed to unbreak tests. The image which is used to run upgrades needs to have the GRUB menu changes introduced lately


mudler:
@davidcassany merging as it is a really minor change and master is broken otherwise.. :man_shrugging:

PR #1038, opened 2022-01-04
ci: restore docker-runner workflow

mudler:
Somehow it got removed..


mudler:
Merging, as it just restores what was added before.. and adds docker to the runner image, so can be used during actions

PR #1055, opened 2022-01-13
Add --silence-hooks to installer to set yip loglevel to error

### LINKED ISSUE #1053: Hide hook cos-setup messages from cos.sh output
There are hook yip logs during installation, it would be nice to have an option to hide them.

e.g.,

148959889-62a7f0a8-696f-4b23-8d24-611a5f4da3ef

An environment variable or option to hide hook messages.

Leave it alone. That's no harm at all.

**Additional context**

https://github.com/harvester/harvester-installer/pull/210#pullrequestreview-849244708

Itxaka:
This should remove all the info messages we get and instead report only
errors

Fixes #1053 

### DIFF
--- a/packages/installer/cos.sh
+++ b/packages/installer/cos.sh
@@ -302,6 +302,9 @@ installer_cleanup2()
     umount ${_STATEDIR}
     umount ${_RECOVERYDIR}
     [ -n "$_COS_INSTALL_ISO_URL" ] && umount ${_ISOMNT} || true
+    if [ "$_COS_SILENCE_HOOKS" = "true" ]; then
+        unset LOGLEVEL
+    fi
 }
 
 installer_cleanup()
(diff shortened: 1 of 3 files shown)


packages/installer/cos.sh
```diff
@@ -1403,6 +1406,8 @@ install() {
             --help)
                 usage
                 ;;
+            --silence-hooks)
```
mudler, line 1409: I'm totally ok with it, but from a naive perspective my first thought was to attach it to a general `--debug` flag instead of having its own option, what do you think?

Itxaka: like if debug is enabled, then show the normal logging from yip, or even set yip to debug, otherwise set it on error by default?

mudler: > like if debug is enabled, then show the normal logging from yip, or even set yip to debug, otherwise set it on error by default?

yup exactly, so in case if something goes off, someone can just add `--debug`, and we have more detailed logs to debug as well... but that's a thought! if you think it's better of keep them separated I'm fine with. Currently we don't use debug extensively here, but maybe it's something to keep in mind for elemental-cli

mudler (approved):

Itxaka:
already green and passed everything, only version numbers changing.

Issue #960, opened 2021-12-13
SElinux policies fails to build


mudler:
**cos-toolkit version:**
https://github.com/rancher-sandbox/cOS-toolkit/commit/bca5b092230d419ead0e83676a27769d73ab5e34

**CPU architecture, OS, and Version:**

**Describe the bug**
Seems both selinux/k3s and selinux/rancher fails to build locally with:

```
 INFO   #9 55.16 Resolving package dependencies...
 INFO   #9 55.24 2 Problems:
       #9 55.24 Problem: the installed shadow-4.8.1-2.43.x86_64 requires 'libsemanage.so.1()(64bit)', but this requirement cannot be provided
       #9 55.24 Problem: the to be installed container-selinux-2.171.0-lp153.17.7.noarch requires 'selinux-policy >= 20211111-lp153.114.7', but this requirement cannot be provided
       #9 55.24 
       #9 55.24 Problem: the installed shadow-4.8.1-2.43.x86_64 requires 'libsemanage.so.1()(64bit)', but this requirement cannot be provided
       #9 55.24  Solution 1: Following actions will be done:
       #9 55.24   deinstallation of shadow-4.8.1-2.43.x86_64
       #9 55.24   deinstallation of python-azure-agent-2.2.49.2-3.20.1.noarch
       #9 55.24  Solution 2: keep libsemanage1-3.0-1.27.x86_64
       #9 55.24  Solution 3: break shadow-4.8.1-2.43.x86_64 by ignoring some of its dependencies
       #9 55.24 
       #9 55.24 Choose from above solutions by number or skip, retry or cancel [1/2/3/s/r/c/d/?] (c): c

```

**To Reproduce**
sudo make PACKAGES="selinux/rancher selinux/k3s" build

**Expected behavior**
Packages to build

Itxaka:
This also happens with a plain leap 15.3 system and the same repo

Itxaka:
I talked with Johannes Segitz (maintainer of the security:SELinux packages on obs), seems like this is a known issue due to `libsemanage wasn't properly versioned in v1,that is now changed but unfortunately this is the fallout`

So we need to workaround on this until 15.4 is released (on June I think?)

Itxaka:
There is a workaround that I found that its...kind of crappy

[...]

PR #1056, opened 2022-01-13
Workaround to build selinux packages

Itxaka:
Use centos as base in order to avoid the current issue with building selinux
policies in leap 15.3

Workaround for #960 

### DIFF
--- a/packages/selinux/build.yaml
+++ b/packages/selinux/build.yaml
@@ -1,7 +1,4 @@
-requires:
-- name: "base"
-  category: "distro"
-  version: ">=0"
+image: centos:7
 env:
 - GITHUB_ORG={{ ( index .Values.labels "github.owner" ) }}
 - GITHUB_REPO={{ ( index .Values.labels "github.repo" ) }}
@@ -9,8 +6,7 @@ env:
 {{ if .Values.distribution }}
 {{if eq .Values.distribution "opensuse" }}
 prelude:
-- zypper ar https://download.opensuse.org/repositories/security:/SELinux/openSUSE_Leap_15.3/security:SELinux.repo
-- zypper --gpg-auto-import-keys in -y --allow-vendor-change --allow-downgrade container-selinux -libsemanage1
+- yum -y install container-selinux curl
 
 steps:
 - |
--- a/packages/selinux/collection.yaml
+++ b/packages/selinux/collection.yaml
@@ -1,7 +1,7 @@
 packages:
   - name: "k3s"
     category: "selinux"
-    version: "0.5.1"
+    version: "0.5.1-1"
     labels:
       github.repo: "k3s-selinux"
       github.owner: "k3s-io"
@@ -14,7 +14,7 @@ packages:
         curl -s -L "https://github.com/{{.Values.labels.github.owner}}/{{.Values.labels.github.repo}}/releases/download/{{.Values.labels.github.tag}}/sha256sum-centos7-noarch.txt" | grep "noarch" | cut -d" " -f 1
   - name: "rancher"
     category: "selinux"
-    version: 0.2.1-4
+    version: 0.2.1-5
     labels:
       github.repo: "rancher-selinux"
       github.owner: "rancher"


mudler (approved):
looks good here overall! just wondering about the checksum, shouldn't that be updated too?

Itxaka:
only changing our package version, not the remote one so checksum would be the same.

But somehow this is failing...only on CI....with a segmentation fault I think???

the wonders of tumbleweed :D

Itxaka:
Maybe I should just bump the revision to avoid any conflicts...I dont like having different version numbers :/

Itxaka:
This now passes the build stage :)

packages/selinux/build.yaml
```diff
@@ -1,16 +1,12 @@
-requires:
-- name: "base"
-  category: "distro"
-  version: ">=0"
+image: centos:7
 env:
 - GITHUB_ORG={{ ( index .Values.labels "github.owner" ) }}
 - GITHUB_REPO={{ ( index .Values.labels "github.repo" ) }}
 
 {{ if .Values.distribution }}
```
mudler, line 6: guess we can drop this if at all now?

mudler: ok ok sounds good, maybe we can just leave a small comment here on the reason why? is there a bugzilla entry open for the issue?

packages/selinux/build.yaml
```diff
@@ -1,16 +1,12 @@
-requires:
-- name: "base"
-  category: "distro"
-  version: ">=0"
+image: centos:7
```
davidcassany, line 1: oh, you are so brave 😄 This feels like building Frankenstein

packages/selinux/build.yaml
Itxaka: nope, only our own issue for tracking this.

davidcassany (commented):
No strong opinion but I am not sure this is a good idea. Is it actually functional? Are selinux policies for centos 7 functional in openSUSE Leap?

Anyway, if the intention is only to get the builds green probably it is simpler to disable the build of these packages until this is fixed.

mudler:
> No strong opinion but I am not sure this is a good idea. Is it actually functional? Are selinux policies for centos 7 functional in openSUSE Leap?
> 
> Anyway, if the intention is only to get the builds green probably it is simpler to disable the build of these packages until this is fixed.

thing is, in order to make it functioning was enough for me to just consume the files from the resulting RPM - so this I think will just work in the same way as before.

The problem now is more on derivative side - they don't see the blocker from zypper so they would just consume the rules regardless

Itxaka:
> No strong opinion but I am not sure this is a good idea. Is it actually functional? Are selinux policies for centos 7 functional in openSUSE Leap?

Oh wait..you didnt know? We actually using package build for centos7 in here, even when using leap 15.3 :) thats the only type of release they have.

I guess we could just use the repo directly, but Im not sure what they are actually doing to build the policies that we end up with on the rpm....

End result is the same no matter the base 'cause we are only saving the compiled policies, no other data so it theory it should work as it was before.

I would even say more, it may work better than before as installing the centos7 rpm files on leap resulted in some errors which were a bit hidden, while installing under centos there is no errors from the install:


From https://github.com/rancher-sandbox/cOS-toolkit/runs/4280068702?check_suite_focus=true :
```
       #8 0.419 k3s-selinux-0.5-1.el7.noarch.rpm: OK
       #8 0.427 warning: k3s-selinux-0.5-1.el7.noarch.rpm: Header V4 RSA/SHA1 Signature, key ID e257814a: NOKEY
       #8 0.476 Preparing...                          ########################################
       #8 0.484 Updating / installing...
       #8 0.484 k3s-selinux-0.5-1.el7                 ########################################
 INFO   #8 6.148 Problems processing filecon rules
       #8 6.148 Failed post db handling
       #8 6.148 Post process failed
 INFO   #8 6.525 /usr/sbin/semodule:  Failed!
 INFO   #8 DONE 6.6s
```

mudler:
> Anyhow we might not need selinux packages at all in the future as might be just built on OBS, and then we don't need anymore to provide it for derivatives

Actually, there seems to be something moving: https://build.opensuse.org/package/show/devel:kubic/k3s-selinux

Itxaka:
> Actually, there seems to be something moving: build.opensuse.org/package/show/devel:kubic/k3s-selinux

That will have the same issue, wont it? It will require container-selinux, which will require libsemanage which will require an updated shadow package....

Itxaka:
This may not be necessary, they may add a workaround to the shadow package on the selinux repo to make it build/publish so we may have the package available today.

Itxaka:
Looks like the shadow package is now published, this should not be needed anymore!

PR #1062, opened 2022-01-14
Use the new elemental run stage in substitution of cos-setup

Itxaka:


Itxaka:
probably requires 0.5.0 to be published already: https://github.com/rancher-sandbox/cOS-toolkit/pull/1063

Itxaka:
Umm, the generated iso is stuck on "A start job is running for /dev/root" indefinitely. @davidcassany do you think I broke the initramfs with the changes? As far as I can see, only the naming has changed, and I renamed all the calls from the cos-generator.sh script to the new service files, so it should be all the same, same services running, unless Im missing something that I cant see :/

Itxaka:
I both cut a new elemental release, used the new version and grepped cos-setup everywhere :/

packages/cos-setup/cos-setup-initramfs.service
```diff
@@ -10,7 +10,7 @@ RootDirectory=/sysroot
 BindPaths=/proc /sys /dev /run /tmp
 Type=oneshot
 RemainAfterExit=yes
-ExecStart=/usr/bin/elemental run-stage initramfs
+ExecStart=-/usr/bin/elemental run-stage initramfs
 
 [Install]
 RequiredBy=initrd.target
```
davidcassany, line 16: This could be an issue too. Honestly I do not remember why I added this as a hard requirement. `WantedBy` looks llike a better approach to me now

Itxaka:
ok, adding the "-" to the services makes the iso boot so indeed, elemental is returning errors, so the service never completes. Adding `-` in front of the command means that systemctl will ignore the exit status.

So now we have to make sure to agree on what do we do. Do we ignore exit status on the service files so the system can boot even with wrong stages? Do we add a flag to not return errors from run stage commands? Do we modify yip upstream to not return errors on non-errors (i.e. if there is no metadata, yip returns an error, it should not probably)?

image


IMHO, not failing to boot is important because it allows you to access the machine and check what is the error (did the config get mangled? is the config incorrect? Is there a bug on the elemental/yip part instead? 

So I would go with a flag to enable no errors on run-stage, then use it in the service files, and in the installer if the strict flag is enabled

cc @dragonchaser @davidcassany @mudler

davidcassany:
@Itxaka as we discussed in the planning meeting I'd say that by default we should ignore errors in run stage command and eventually add some sort of `--strict` flag to the command to allow the user to be strict if needed. IMHO it is also worth considering to make all stages `WantedBy` the specific target rather than `RequiredBy`. With that if some stage fails systemd will report the error but the mentioned target will not fail and continue with the scheduled units. Or at least this is what I'd expect.

Itxaka:
okayyyy, pinning to an older version does not work. Does that mean that I need to copy the old package over the backports??? yikes

davidcassany (approved):
Nice! Looks good to me, I see only the raw test is failing, have you tried it locally? it should be relatively easy to reproduce this locally.

Itxaka:
yep, can reproduce locally. seems like during cos-setup-rootfs elemental should execute the userdata and save it to disk, then the same elemental would pick it up and partition the disk.

But that is not happening somehow. rootfs is executed correctly and the provider stuff seems to work. Gonna use some debugging around to see, because it looks like something is still wrong on the run-stage

Itxaka:
ok, got it. I should not skip directories I guess becuase in this case I skip non existing dirs, but the /oem dir is created afterwards....

PR #1074, opened 2022-01-18
Use 'elemental install' instead of the former bash script 

### LINKED ISSUE #1072: Consume elemental-cli install
Consume elemental-cli in place of cos-installer in cos-toolkit:

- [x] modify system/cos so that uses elemental-cli for installation #1074
- [x] make sure it passes tests and maybe validate some of the scenarios that are not covered manually too https://github.com/rancher-sandbox/cOS-toolkit/issues/581
- [x] check harvester [impact](https://github.com/harvester/harvester-installer/blob/3f1dbfccacc7c2aa57ec84449d27719d2586b59f/package/harvester-os/files/usr/sbin/harv-install#L306)

davidcassany:
Fixes #1072 


tests/installer/installer_test.go
```diff
@@ -174,7 +174,7 @@ var _ = Describe("cOS Installer tests", func() {
 				By("Running the cos-installer with a config file")
 				Expect(err).To(BeNil())
 				By("Running the installer")
-				out, err := s.Command("cos-installer --config /tmp/config.yaml /dev/sda")
+				out, err := s.Command("cos-installer --cloud-init /tmp/config.yaml /dev/sda")
```
davidcassany, line 177: @Itxaka in elemental the installer flag `--config` is translated to `--cloud-init`, was done it on purpose? I did not realize of this mismatch while checking flags as the functionality is there and I can't remember if there was a reason to name it differently. Indeed I prefer `--cloud-init` as we might want to reserve `--config` for installer specific configurations in the future.

Itxaka: umm, not on purpose, but probably due to the `config-dir` flag already in elemental and to clarify what type of config is, as on the new elemental we are supposed to have a config file.

davidcassany:
Marked as ready to review since CI tests are expected to be green and I believe there won't be further significant changes from now on. However before merging I'd like to verify some manual tests and double check some little behavior changes (environmental variables, defaults, etc.)

packages/installer/definition.yaml
```diff
@@ -1,6 +1,6 @@
 name: "installer"
 category: "utils"
-version: "0.27.0"
+version: "0.28.0"
```
Itxaka, line 3: current master is 0.28.0!

packages/cos/collection.yaml
```diff
@@ -2,20 +2,18 @@ packages:
   - &cos
     name: "cos"
     category: "system"
-    version: 0.7.14-2
+    version: 0.7.23
```
Itxaka, line 5: master version is 0.7.15

packages/installer/definition.yaml
Itxaka: Also the requires have changed (elemental), be sure to rebase to master before!

packages/cos/collection.yaml
Itxaka: current requires of cos has the installer at version 0.27.0, you would need to bump it if you bump the installer version also https://github.com/rancher-sandbox/cOS-toolkit/blob/master/packages/cos/build.yaml#L11

Itxaka (changes requested):
3 small changes

Itxaka (approved):
Lgtm

davidcassany:
There are a couple of behavior changes to discuss:

* `--config` flag has been renames to `--cloud-init`. I do believe this actually makes sense, but note this is a breaking change. Shall we add some `--config` deprecated flag that behaves as `--cloud-init`? I don't think so, as I'd like to start using `elemental install --config <install-yaml>` at some point not that far away. For that to happen I believe it is better to actually force a flag rename now.
* `cosign` plugin is non functional with the current implementation as it used to be. Cosign plugin requires few environmental variables to run. Former installer code used to silently inject them. I don't want to do such thing in elemental install code. This is way too far from obvious. In addition, `cosign` plugin in luet is likely to be enabled by default. Would it actually make sense to invert the logic of `--no-cosign` flag and use a `--cosign` flat to explicitly enable it? This way we, if we run sanity checks on `--cosign` presence to verify env vars, I believe it would be way easier for the user to understand some extra configs or steps might be required. Also, in parallel, I believe having a proper channel of communication between elemental and luet plugins is something to consider, however I am not aware of the challenges of such a thing.

Itxaka:
RE: config vs cloud-init

cloud-init all way. Like we have to maintain the functionality the same as before, but that doesnt mean we need to keep the flags and everything the same, for example, we are changing the config files location and format.

IMHO having an alias would be shooting in our own feet. Lets transition to nicer things now that we can :D

davidcassany:
> Like I get that on luet we needed to do it as a plugin, but on elemental we dont need to, we can have it directly and take advantage of us controlling all the flags passed and so on, so its much easier to avoid the env vars. Plus with our config stuff it should still be possible to use env vars (with ELEMENTAL on front) so win win. I would not support luet plugins on elemental as they have to be tied to whatever luet does, while on elemental we can do whatever we want.

I like this idea, then I just need to understand when and how to call the required plugins. Do you have any simple hints regarding that? Mostly about when to call it, I guess right after the luet call, but I am not 100% sure. I guess I will simply discover how to call the plugins.
Gonna have a look to this idea, if this is straight forward I'll try to implement it in the scope of this PR, otherwise I'd suggest to leave it for a follow up PR with a dedicated card.

Itxaka:
yes, leaving for a different PR seems like a good idea.

The plugins stuff should be simple (HA!) as we only need to listen for the events and react to them, so it should be the cosign object or whatever, subscribing to the bus.Manager events and reacting to them.

@mudler probably knows if this is correct or more complicated than that?

davidcassany:
> The plugins stuff should be simple (HA!) as we only need to listen for the events and react to them, so it should be the cosign object or whatever, subscribing to the bus.Manager events and reacting to them.

mmm not so sure, I guess this requires to run luet and listen to events at the same time... plus having some coordination between both. Unless I am not understanding it to me it feels like this approach of relaying on luet events on elemental code requires setting a plugin to communicate luet and elemental (a mutex to to wait and raise trigger on events) and then run an event listener in elemental in a separate thread... Isn't it too complicated? Most events I see in Luet looks like they are before or after the main blocks (upgrade, unpack, install, etc), so I'd say simply running the plugin before or after is enough and do not mess with luet's events bus.

Itxaka:
sure we can, but AFAIK is just a simple use of the manager from go-plugabble to subscribe and receive events no? Maybe not, I dont know, I just expected it to work like that, probably Im wrong :D

Itxaka:
bah, go pluggable only works with external plugins, not with internal objects :/

Itxaka:
In any case, as we control the workflow, its even simpler to just call the cosign/mtree objects on the created artifacts at build time, or after download(even before download with cosign!) the install/upgrade artifacts :D

davidcassany:
> In any case, as we control the workflow, its even simpler to just call the cosign/mtree objects on the created artifacts at build time, or after download(even before download with cosign!) the install/upgrade artifacts :D

Yes this is what I thought it would be simpler. Just create an mtree and cosign interfaces in elemental and call what we need when we need it.

PR #1077, opened 2022-01-19
Add framework images

mudler:
Adds multiplatform framework images that can be directly consumed downstream in Dockerfiles with `COPY --from=<image> / /` to apply the framework without any `luet` setup/interaction.

Context: os2 create its own framework images, but that was already along the lines for the #108 features to bring static binaries for airgap. 

By shipping our own tagged framework images we relief machinery while building Dockerfiles, and offer also a way to get signed images with all the tools (cosign,etc included)

### DIFF
--- a/images/framework/cos.yaml
+++ b/images/framework/cos.yaml
@@ -1,0 +1,21 @@
+repositories:
+- &cos
+  name: "cos"
+  description: "cOS official"
+  # Repository type
+  type: "docker"
+  # Repository cache
+  cached: true
+  # Repository priority
+  priority: 1
+  # Repository architecture
+  arch: "amd64"
+  # Enable/disable Docker notary checks
+  # 
+  verify: false
+  urls:
+  - "quay.io/costoolkit/releases-green"
+- <<: *cos
+  arch: "arm64"
+  urls:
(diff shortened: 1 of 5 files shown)


images/framework/cos.yaml
```diff
@@ -0,0 +1,21 @@
...
+  priority: 1
+  # Repository architecture
+  arch: "amd64"
+  # Enable/disable Docker notary checks
+  # 
+  verify: false
+  urls:
+  - "quay.io/costoolkit/releases-green"
+- <<: *cos
+  arch: "arm64"
```
Itxaka, line 19: umm, AFAIK luet is (was?) arch agnostic? How would this work? Inst this setting boths repos for both arches?

mudler: so luet _is_ arch agnostic, but allows to have clients define arch for repositories, in which case gets automatically enabled if the relevant arch matches in runtime.

https://rancher-sandbox.github.io/cos-toolkit-docs/docs/examples/creating_bootable_images/#configuration-file

Itxaka (approved):

mudler:
Merging as can see the outcome of it only from running on master :slightly_smiling_face:

PR #1082, opened 2022-01-21
Move elemental to git based tracking for versions

Itxaka:
Instead of having to bump the version on elemental each time we need a
change, we can move to using a git_sha and keep our versions separated
from the elemental ones for now as we may need pathes and commits from
elemental and we dont want to keep generating releases.

Once elemental is stable we can change back to using the package version
as the elemental version again, thus for now we should only bump the
revision and keep the version number at 0.0.12 until then.

### DIFF
--- a/packages/toolchain/elemental-cli/collection.yaml
+++ b/packages/toolchain/elemental-cli/collection.yaml
@@ -3,12 +3,14 @@ packages:
     name: "elemental-cli"
     category: "toolchain"
     bin_name: "elemental"
-    version: 0.0.11
+    version: 0.0.13-0
     fips: false
     labels:
       github.repo: "elemental-cli"
       github.owner: "rancher-sandbox"
       autobump.revdeps: "true"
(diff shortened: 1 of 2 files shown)


packages/toolchain/elemental-cli/collection.yaml
```diff
@@ -3,12 +3,14 @@ packages:
...
     category: "toolchain"
     bin_name: "elemental"
-    version: 0.0.5-1
+    version: 0.0.12-0
     fips: false
     labels:
       github.repo: "elemental-cli"
       github.owner: "rancher-sandbox"
       autobump.revdeps: "true"
+      git.hash: "d1d11006f556aacaddf82a403894a4445b7b696c"
```
davidcassany, line 12: this is populated automatically by the autobump right?

davidcassany (approved):
Looks good to me if the autobump is also capable to update the hash and version

PR #1091, opened 2022-01-24
scripts: detect system arch

mudler:


scripts/get_luet.sh
```diff
@@ -15,10 +15,20 @@ LUET_CONFIG_PROTECT=${LUET_CONFIG_PROTECT:-1}
...
+if [ -z "$LUET_ARCH" ]; then
+    LUET_ARCH=$(uname -m)
 fi
 
+case $LUET_ARCH in
+    amd64|x86_64)
+        LUET_ARCH=amd64
+        ;;
+    arm64|aarch64|arm*)
+        LUET_ARCH=armv7
```
Itxaka, line 28: is this correct? armv7 is not aarch64...

mudler: oh well, we need to as it is used in the repo generation :sweat_smile:

Itxaka: well, as long ar the LUET_ARCH is not used for constructing repos for example, it should be ok. but the only repos we have are constructed with arm64 naming on there so...

Itxaka (approved):

Itxaka:
doesnt seem to work for raw images :D

mudler:
> doesnt seem to work for raw images :D

Indeed :roll_eyes:

PR #1098, opened 2022-01-25
Use pull-image from elemental instead of luet util unpack

Itxaka:


davidcassany (approved):
Looks good

PR #1117, opened 2022-02-07
ci: Fixup docker socket permission in runner image

mudler:
In case /var/run is mounted, it prevents issues with the runner trying to run docker

Also adds an example k8s deployment


.github/docker/ghrunner/entrypoint.sh
```diff
@@ -7,15 +7,16 @@ REPO=${REPO:-dockerhub-autobuild}
...
 
 if [ -z "${ORG}" ]; then
-    echo "missing ORG, bailing out!"
+    echo "WARN: missing ORG"
+    exit 1
 fi
 
 if [ -z "${REPO}" ]; then
-    echo "missing REPO, bailing out!"
+    echo "WARN: missing ORG"
```
Itxaka, line 15: Missing REPO

mudler: argh, fixing, thx!

Itxaka (approved):

mudler:
I've already tried the image as it's already pushed :smiley:  but doesn't affect CI at all.. so I'd just go ahead and merge it (and seems to run fine now: https://github.com/rancher-sandbox/os2/runs/5090979831?check_suite_focus=true)

PR #1118, opened 2022-02-07
ci: github-runner add affinity rules

mudler:
So replicas are spread into different nodes


mudler:
Just redeployed workers with this in place

PR #1119, opened 2022-02-07
 Use elemental reset command instead of cos.sh script 

davidcassany:


davidcassany:
Note there will be a couple of installer tests failures. Those are known and unrelated to the reset command. Those are known and partially, if not fully, fixed in #1122. The fix for those is related to adapting the log messages that installer test expects, I did not include a fix for those to not conflict and overlap with #1122. So these installer test failures can be ignored.

mudler (approved):
Looks good here!

PR #1122, opened 2022-02-10
Use elemental upgrade instead of cos.sh script

Itxaka:


packages/installer/upgrade.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 set -e
 
-cos upgrade $@
+elemental --debug --logfile /tmp/elemental.log upgrade $@
```
davidcassany, line 4: Good point I forgot about the logfile in the installer and was about to for the reset command

davidcassany (approved):
Looks good

packages/installer/upgrade.sh
Itxaka: The logfile is also picked up by the test runner if there is a failure and uploaded to github!

tests/installer/installer_test.go
```diff
@@ -203,29 +202,19 @@ var _ = Describe("cOS Installer tests", func() {
...
+			It("fails to install with a wrong config file and strict flag", func() {
 				err := s.SendFile("../assets/broken-config.yaml", "/tmp/config.yaml", "0770")
 				By("Running the cos-installer with a broken config file")
 				Expect(err).To(BeNil())
 				By("Running the installer")
-				out, err := s.Command("cos-installer --cloud-init /tmp/config.yaml /dev/sda")
-				Expect(err).To(BeNil())
+				out, err := s.Command("cos-installer --strict --debug --cloud-init /tmp/config.yaml /dev/sda")
+				Expect(err).To(HaveOccurred())
 				Expect(out).To(ContainSubstring("Copying Active image..."))
```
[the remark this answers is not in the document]
Itxaka: yeah, was trying to somehow make it work but...like we are already testing this part on elemental and its easy to test in there...should probably kill it and forget about it.

davidcassany:
@Itxaka I think we can ignore the installer tests for this PR and fix them in follow up PRs (they are unrelated to upgrade), if the tests for the intalled system (squashfs and nonsquashfs) are green I'd merge it.

Itxaka:
> @Itxaka I think we can ignore the installer tests for this PR and fix them in follow up PRs (they are unrelated to upgrade), if the tests for the intalled system (squashfs and nonsquashfs) are green I'd merge it.

Should be an easy fix, its only that the installer has changed the log output to use the LABEL instead of the active/passive. I can add those changes to this PR

PR #1127, opened 2022-02-14
Disable some efi installer tests

davidcassany:


Itxaka (approved):

Itxaka:
I would say no need to wait, lets force push this directly

davidcassany:
> I would say no need to wait, lets force push this directly

Right, lets give it a try, if there is something wrong with this setup it only affects installer efi tests and it should be simple to fix.

Issue #842, opened 2021-11-04
Autobumper should not add version if there is no version key


Itxaka:
**cos-toolkit version:**
master

**CPU architecture, OS, and Version:**
N/A

**Describe the bug**
Currently we use anchors on the cos collection.yaml file to avoid repeating some of the stuff, including version numbers for all the different cos-whatever artifacts.
This keeps the versions in sync, especially between the cos system and cos recovery, as they should always match,

Autobumper is adding version numbers to those entries and may desync them, so you end up with a recovery with a version and a system with another.

The contents are the same, so its not a big problem but it could lead to confusion, after a cos-reset of one of your nodes in a cluster it suddenly reports a different version number for example.

**To Reproduce**
have a recovery with a different version number than your cos system, run cos-reset, check version number

**Expected behavior**
Recovery and system to have the same version.

**Logs**
Tests cached this https://github.com/rancher-sandbox/cOS-toolkit/runs/4097481553?check_suite_focus=true

**Additional context**

Itxaka:
This can be closed. If its not implemented already its not gonna be and its only a bit annoying not a real bug

Issue #654, opened 2021-09-16
Consider rewriting the autobumper in golang


mudler:
**Is your feature request related to a problem? Please describe.**
The autobumper which is responsible of @cOS-cibot automated PRs is currently implemented in bash and using `yq` (https://github.com/Luet-lab/extensions/blob/master/extensions/autobump-github/luet-autobump-github). Although it works, it reached it's implementation limits and the lack of tests makes hard to implement new features

**Describe the solution you'd like**

A reimplementation of the auto-bumper, for e.g. in golang, complete of test-suite.

During https://github.com/rancher-sandbox/cOS-toolkit/issues/602#issuecomment-915104257 for example we could have introduced the ability to handle different package versions in the same tree, as the bash version now just does handle bumps by editing the specfiles in place. 

A WIP comprensive of unit tests is here: https://github.com/mudler/luet-autobumper, it is not feature-parity yet with the bash equivalent but it already auto-bumps inplace or either add new spec files to the tree. It also handles reverse dependencies correctly.

**Describe alternatives you've considered**
Currently with #653 we address to #602 by allowing all the released-ever metadata to be installed. We download the available metadata in the repository before publishing the new tree version.

**Additional context**
See #602 and #653

PR #1149, opened 2022-02-25
bump elemental to v0.0.13

davidcassany:

### DIFF
--- a/packages/cos/collection.yaml
+++ b/packages/cos/collection.yaml
@@ -2,7 +2,7 @@ packages:
   - &cos
     name: "cos"
     category: "system"
-    version: 0.8.2-6
+    version: 0.8.2-8
     description: "cOS base image, used to build cOS live ISOs"
     brand_name: "cOS"
     labels:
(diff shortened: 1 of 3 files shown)


mudler (approved):

packages/cos/collection.yaml
```diff
@@ -2,20 +2,20 @@ packages:
...
+    version: 0.8.2-8
     description: "cOS base image, used to build cOS live ISOs"
     brand_name: "cOS"
     labels:
       autobump.revdeps: "true"
   - !!merge <<: *cos
     name: "cos-container"
     description: "cOS container image, used to build cOS derivatives from scratch"
-    version: 0.8.2-7
+    version: 0.8.2-8
```
davidcassany, line 13: @mudler do you have an idea why these version literals keep appearing here? Is it caused by the autobump? I removed them already few times, it is quite annoying having to manually trace all this literals when we could enforce them to be equal and centralized.

mudler: Correct :( it is the autobumper which doesn't understand yaml syntax properly and re-applies them anytime.. sadly that would require some work on it which is tracked in #842 and #654

PR #1158, opened 2022-03-02
Split cloud config files

mudler:
This PR creates single packages for each cloud config file so they can be installed singularly. It also moves the current `system/cloud-config` as a meta package, that pulls down all the cloud config files

### DIFF
--- a/packages/cloud-config/build.yaml
+++ b/packages/cloud-config/build.yaml
@@ -7,23 +7,23 @@ requires:
   version: ">=0"
 
 steps:
-- mkdir -p /system/oem
-- cp -rfv oem/{{.Values.oem_file}} /system/oem
-- chmod -R 600 /system/oem
+- mkdir -p {{.Values.oem_dir}}
+- cp -rfv oem/{{.Values.oem_file}} {{.Values.oem_dir}}
+- chmod -R 600 {{.Values.oem_dir}}
 {{ if .Values.templated }}
-- cos-tmpl-render /system/oem/{{.Values.oem_file}}
+- cos-tmpl-render {{.Values.oem_dir}}/{{.Values.oem_file}}
 {{ end }}
 
 {{ if eq .Values.oem_file "02_upgrades.yaml" }}
 {{ if .Values.codename }}
-- sed -i 's/:FLAVOR:/{{.Values.codename}}/g' /system/oem/02_upgrades.yaml
+- sed -i 's/:FLAVOR:/{{.Values.codename}}/g' {{.Values.oem_dir}}/02_upgrades.yaml
 {{end}}
 {{ if .Values.arch }}
   {{if eq .Values.arch "x86_64"}}
-- sed -i 's/:ARCH://g' /system/oem/02_upgrades.yaml
+- sed -i 's/:ARCH://g' {{.Values.oem_dir}}/02_upgrades.yaml
   {{end}}
   {{if eq .Values.arch "aarch64"}}
-- sed -i 's/:ARCH:/-arm64/g' /system/oem/02_upgrades.yaml
+- sed -i 's/:ARCH:/-arm64/g' {{.Values.oem_dir}}/02_upgrades.yaml
   {{end}}
 {{end}}
 {{ end }}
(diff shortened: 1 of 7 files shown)


packages/cloud-config/collection.yaml
```diff
@@ -0,0 +1,39 @@
+packages:
+- &base
+  name: defaults
+  category: cloud-config
+  version: "0.1"
+  oem_file: "01_defaults.yaml"
```
Itxaka, line 6: cant we add the destination in here as well? so in case of change (shouldn't happen) everything is in the same place?

packages/cloud-config/collection.yaml
```diff
@@ -0,0 +1,39 @@
...
+  templated: false
+  requires:
+  - name: "cos-setup"
+    category: "system"
+    version: ">=0"
+- <<: *base
+  name: "rootfs"
+  oem_file: "00_rootfs.yaml.tmpl"
```
Itxaka, line 14: SUPER NIT: why is the 00_X file the second one done. This is killing my OCD. can it be put in order?

packages/cloud-config/build.yaml
```diff
@@ -7,11 +7,14 @@ requires:
...
 - chmod -R 600 /system/oem
-- cos-tmpl-render /system/oem/00_rootfs.yaml.tmpl
-- cos-tmpl-render /system/oem/06_recovery.yaml.tmpl
+{{ if .Values.templated }}
+- cos-tmpl-render /system/oem/{{.Values.oem_file}}
+{{ end }}
+
+{{ if eq .Values.oem_file "02_upgrades.yaml" }}
 {{ if .Values.codename }}
 - sed -i 's/:FLAVOR:/{{.Values.codename}}/g' /system/oem/02_upgrades.yaml
```
Itxaka, line 19: arent we able to use the templating to substitute this by default?

packages/cloud-config/build.yaml
```diff
@@ -7,11 +7,14 @@ requires:
...
 
 steps:
-- mkdir -p /system /oem
-- cp -rfv oem /system
+- mkdir -p /system/oem
+- cp -rfv oem/{{.Values.oem_file}} /system/oem
 - chmod -R 600 /system/oem
-- cos-tmpl-render /system/oem/00_rootfs.yaml.tmpl
-- cos-tmpl-render /system/oem/06_recovery.yaml.tmpl
+{{ if .Values.templated }}
```
Itxaka, line 13: I wonder if we should use a helper here to make it simpler...

Itxaka (approved):

packages/cloud-config/build.yaml
mudler: thought about it, but the usage is strictly scoped to this package at the end

davidcassany: yes, me too. Probably something like that is enough:

```bash
for tmpl in {{.Values.oem_dir}}/*.tmpl; do
 cos-tmpl-reder "${tmpl}"
done
```
Or, alternatively, as in the above comment it could be applied on per package basis.

packages/cloud-config/build.yaml
```diff
@@ -7,19 +7,23 @@ requires:
   version: ">=0"
 
 steps:
-- mkdir -p /system /oem
-- cp -rfv oem /system
-- chmod -R 600 /system/oem
-- cos-tmpl-render /system/oem/00_rootfs.yaml.tmpl
-- cos-tmpl-render /system/oem/06_recovery.yaml.tmpl
+- mkdir -p {{.Values.oem_dir}}
+- cp -rfv oem/{{.Values.oem_file}} {{.Values.oem_dir}}
```
davidcassany, line 11: Why don't we copy files based on the package name? This way we don't need the `oem_file` template, we just pick the require file (or multiple files) in each package.

IMHO this way is more scalable and easier to read.

davidcassany (approved):
Looks good to me.

IMHO the `oem_file` and `templated` parameters of the `coloud-config/collention.yaml` could be avoided. It is just a matter a of style, feel free to ignore it.

packages/cloud-config/build.yaml
mudler: well, they have numbers in front and I avoided to do any smart code here. I thought it would be quite unreadable to have numbers in front of the package name, and wildcard.. well can have unexpected consequences.

PR #1168, opened 2022-03-08
Drop cos.sh script

### LINKED ISSUE #1159: Drop cos.sh script
This card is about dropping `packages/installer/cos.sh` script.

While working on #1144 I realized we do not have a proper plan/steps for that and we should. I believe we are really close to do it in terms of feature set. I believe we mostly miss a replacement for `cos-deploy` (only for few corner cases if we actually aim to support them).

I believe we should keep `cos-installer` and equivalent wrappers around elemental just for backward compatibility but get rid of `cos.sh` in `installer` package, we still have the backport if we need the old `cos.sh`.

AFAIK this is only going to affect Harvester, so I guess this should be coordinated with them, even we already had some discussions about it and the impact surface is really limited.

davidcassany:
This PR removes the cos.sh script from the installer package and it only keeps the wrappers around elemental for backward compatibility. In addition it also removes the `cos-rebrand` command.

As a result of this change `cos-deploy` is based on `elemental reset` command which is only supported when running from the recovery partition. In practice it means that running `cos-deploy` from active or passive systems is no longer supported, it should not be a problem as this was never meant to be primary use case.

Fixes #1159


mudler (approved):
Looks good here!

davidcassany:
I added the do not merge label since I want to sync with harvester before merging this PR.

@bk201 in this PR we are finally removing the old `cos.sh` script. This impacts to Harvester as this script was called within the harvester installer script. As we discussed earlier there are two alternatives from Harvester PoV to adapt to the new situation:

* Start using elemental install as discussed and exposed in https://github.com/rancher-sandbox/cOS-toolkit/issues/1072#issuecomment-1024411482
* Pin `system/installer` used by Harvester to version `v0.28.2`, this is the version we are still publishing in backports. No plans to delete it yet.

Does it looks good to you? Can we proceed with the merge? We don't want to break anything on Harvester side.

bk201:
@davidcassany Thanks for pinging. We [pinned luet repository to an index](https://github.com/harvester/os2/blob/1a736bdc35a0d97a13f64319410a46eb515ff49f/files/etc/luet/luet.yaml#L17), and the installed utils/installer-0.30.0 looks good to us. And I think we're going to migrate to elemental install in the next iteration, since it should be the main supported tool.  cc @johnliu55tw

Itxaka:
is there any reason to keep the installer package now? Like cant the other packages derivatives just use the elemental-cli package and we can drop this already? Im not sure what its the advantage of keeping a wrapper package anymore, especially if the backport is already in there for derivatives to use it.

packages/installer/deploy.sh
```diff
@@ -3,4 +3,4 @@ set -e
 
 export INTERACTIVE=true
 
-cos install --no-format $@
\ No newline at end of file
+elemental --debug --logfile /tmp/elemental.log reset $@
```
Itxaka, line 6: Im not sure if this is good to have on a production package. I mean, its good for testing, but any derivative using this is gonna get a lot of logs which makes no sense to have.

mudler:
> is there any reason to keep the installer package now? Like cant the other packages derivatives just use the elemental-cli package and we can drop this already? Im not sure what its the advantage of keeping a wrapper package anymore, especially if the backport is already in there for derivatives to use it.

this might impact too many CIs at the moment. The backward compatibility just makes sure a smooth transitioning... I would keep it around for a while until we are all settled with the new CLI

packages/installer/deploy.sh
davidcassany: Yes, to remove that we need to also modify our tests. However we are documenting everywhere to use `elemental` not its backward compatibility wrappers `cos-*`, so only those using these wrappers will face this logs.

We can remove `--logfile` options from wrappers but then we need to start testing in our CI using `elemental` or to embed this config with some env var. I'd suggest to do it in follow up PR so we prior backward compatibility and reduced impact on CI tests first.

davidcassany:
> @davidcassany Thanks for pinging. We [pinned luet repository to an index](https://github.com/harvester/os2/blob/1a736bdc35a0d97a13f64319410a46eb515ff49f/files/etc/luet/luet.yaml#L17), and the installed utils/installer-0.30.0 looks good to us. And I think we're going to migrate to elemental install in the next iteration, since it should be the main supported tool. cc @johnliu55tw

Then my suggestion is to include installer v0.30.0 in backports. That means the backport package includes `cos.sh` script and `elemental` binary and uses it for `cos-installer`, `cos-reset` and `cos-upgrade` wrappers. Those are not used by Harvester, so no issue there. @mudler I think adding v0.30.0 into the backport makes sense as the backported package then provides all the tools, old `cos.sh` and also a reasonable `elemental`. Also, this way, we completely align with Harvester and provide a drop-in backport for them in case this is needed.

davidcassany:
@mudler Just validated manually that an disk created with `make raw_disk` form this branch creates an image that can be imported to AWS as a snapshot and create an instance out of this. Using the `images/sample_user-data.yaml` just sets active and passive from the recovery image.

@bk201 @johnliu55tw so all in all there is nothing preventing us to no merge this PR. Please consider that any `utils/installer` cOS-toolkit package  over v0.30.0 won't have the `cos.sh` script any longer.

PR #1191, opened 2022-03-21
Add service that allows us to control the cdrom eject

Itxaka:
This creates a service file that we can control by setting
/run/cos/ejectcd and that will also check to see if we booted with the
cdboot stanza in the cmdline in order to trigger.

By default is set to run after shutdown and reboot if those two
conditions met.

This allows us and derivatives to control ejecting the cd at the end of
an install for example

This is part of https://github.com/rancher-sandbox/os2/issues/31 and I consider that it works much better if we support it trougth the base cOS, instead of patching it in the middle, which we can do, but its nice to have also for our testing.

### DIFF
--- a/packages/cos-setup/build.yaml
+++ b/packages/cos-setup/build.yaml
@@ -14,3 +14,4 @@ steps:
 - systemctl enable cos-setup-fs.service
 - systemctl enable cos-setup-boot.service
 - systemctl enable cos-setup-network.service
+- systemctl enable cos-setup-ejectcd.service
--- a/packages/cos-setup/cos-setup-ejectcd.service
+++ b/packages/cos-setup/cos-setup-ejectcd.service
@@ -1,0 +1,20 @@
+[Unit]
+Description=Eject the DVD
+After=shutdown.target reboot.target
+DefaultDependencies=no
+# check the ejectcd file was created
+ConditionPathExists=/run/cos/ejectcd
+# also check if we booted from CD
+# probably both work, but if we change the cdlabel then the first fails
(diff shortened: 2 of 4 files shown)


mudler (approved):
Looks good here, this needs docs too

davidcassany (approved):
Looks good to me

Itxaka:
umm, doesnt seem to work. Im guessing that becuase the system is inmutable and we run the eject at the end, when the disk has been remounted RO, the changes are lost. Its also run after /tmp has been unmounted so no idea where I could put that...

mudler:
> umm, doesnt seem to work. Im guessing that becuase the system is inmutable and we run the eject at the end, when the disk has been remounted RO, the changes are lost. Its also run after /tmp has been unmounted so no idea where I could put that...

It doesn't trigger the service at all? do you have some unit logs? maybe you can push the output logs to `/oem` and make sure that is mounted first if you can't catch those up

maybe also you can try by specifying a `/usr/lib/systemd/system-shutdown/name.shutdown` script instead? (assumption is that persists there as there is no need of service activation) If neither work we might have to go back at the drawing board, as this could require changes to our dracut modules to support this

Itxaka:
> maybe also you can try by specifying a /usr/lib/systemd/system-shutdown/name.shutdown script instead? (assumption is that persists there as there is no need of service activation) If neither work we might have to go back at the drawing board, as this could require changes to our dracut modules to support this

Yep, that was my next try and that seems to work. Adding a simple script in there with the eject, trigger the eject at the very end so there is no issues in there at all. Plus that makes it work with elemental only and not touching the base os? Which may be kind of good...or not, I dont know anymore :/

PR #1204, opened 2022-03-25
Unflaky ci tests

### LINKED ISSUE #1202: Unflaky CI tests
There are several points of our pipeline which makes our tests flaky and as such we have red builds since a while. This card is to fixups the known issues and/or disable the tests suites we don't care about.

Example:
- https://github.com/rancher-sandbox/cOS-toolkit/runs/5662577414?check_suite_focus=true 
  ( here the script can be enhanced, as API rate limit of GH kicked in )
- https://github.com/rancher-sandbox/cOS-toolkit/runs/5662459315?check_suite_focus=true runs the same test suites on BIOS, but efi is much slower and time outs the suite. Let's just keep a EFI test suite for sanity check (booting it up would be enough) and keep the BIOS one. Aside now with the elemental-cli we have better battery tests for that part as well.

mudler:
- Disable part of the EFI installer tests which makes it flaky. Keeps one portion for sanity check (that it boots and can install, without performing a reboot)
- Split up installer tests
- Tighten up the bootstrap script used in ci to install `luet`. Adds checksum verification and also installation via docker from our repositories which is then forced to be used in our pipelines.

Fixes #1202

### DIFF
--- a/scripts/get_luet.sh
+++ b/scripts/get_luet.sh
@@ -7,11 +7,6 @@ fi
 set -ex
 export LUET_NOLOCK=true
 
-LUET_VERSION=$(curl -s https://api.github.com/repos/mudler/luet/releases/latest | grep tag_name | awk '{ print $2 }' | sed -e 's/\"//g' -e 's/,//g' || echo "0.9.24" )
-if [[ -z "${LUET_VERSION}" ]]; then
-    LUET_VERSION="0.22.7"
-fi
-
 LUET_ROOTFS=${LUET_ROOTFS:-/}
 LUET_DATABASE_PATH=${LUET_DATABASE_PATH:-/var/luet/db}
 LUET_DATABASE_ENGINE=${LUET_DATABASE_ENGINE:-boltdb}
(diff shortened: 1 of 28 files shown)


davidcassany (approved):

scripts/get_luet.sh
```diff
@@ -33,19 +30,36 @@ case $LUET_ARCH in
...
+  REPO_URL="quay.io/costoolkit/releases-green-$LUET_ARCH"
+else
+  REPO_URL="quay.io/costoolkit/releases-green"
+fi
+
+if [[ "$DOCKER_INSTALL" == "true" ]]; then
+  _DOCKER_IMAGE="$REPO_URL:luet-toolchain-$LUET_VERSION"
+  echo "Using luet bootstrap version from docker image: ${_DOCKER_IMAGE}"
+  docker run --entrypoint /usr/bin/luet --name luet ${_DOCKER_IMAGE} --version
+  docker cp luet:/usr/bin/luet ./
```
Itxaka, line 43: should we clean the container afterwards?

Itxaka (approved):

PR #1211, opened 2022-03-30
simplify AMI publish name

Itxaka:
- No need for github sha, is part of the tags
 - No need for arch its already in there as part of ec2
 - No need for character substitution, we no longer use + signs in
versions


Itxaka:
no need to expend cpu cycles on tests for this

PR #1220, opened 2022-04-06
Disable floppy when grub searchs for files/labels

Itxaka:
It was reported that searching includes floppy search by default and
that could lead to grub hanging while checking non-existant floppy
drives.

This patch fully disables floppy search

Partial: https://github.com/harvester/harvester/issues/2115

### DIFF
--- a/.github/build.yaml.gomplate
+++ b/.github/build.yaml.gomplate
@@ -56,7 +56,7 @@
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-            go-version: '^1.16'
+            go-version: '1.17'
   {{{- end }}}
   {{{- if $config.local_runner }}}
       - name: Install make
@@ -386,7 +386,7 @@
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-            go-version: '^1.16'
+            go-version: '1.17'
       - uses: actions/checkout@v2
       - name: Download vagrant box
         uses: actions/download-artifact@v2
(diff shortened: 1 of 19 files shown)


mudler (approved):
Looks good

johnliu55tw:
I was testing the `--no-floppy` option, and seems like this option is not available for subcommands such as `search.fs_label`?
image

Maybe it's because I'm testing on GRUB 2.04 instead of the latest 2.06?

Itxaka:
Umm maybe the parameter order is wrong, search.fs_label is an alias to search. Let me have a look

davidcassany (approved):

Itxaka:
>@Itxaka just wants to confirm if this fix only applies to new-installed systems. If a node is upgraded from a previous version, the --no-floppy options are not updated to /run/initramfs/cos-state/grub2/grub.cfg.

@bk201 For updated nodes, as soon as its updated to a derivative that has this package fixed, it should replace the existing grub.cfg, so it should work for upgrades as well. After all, the grub2 package for that upgraded version will contain this fix.

mudler:
> > @Itxaka just wants to confirm if this fix only applies to new-installed systems. If a node is upgraded from a previous version, the --no-floppy options are not updated to /run/initramfs/cos-state/grub2/grub.cfg.
> 
> @bk201 For updated nodes, as soon as its updated to a derivative that has this package fixed, it should replace the existing grub.cfg, so it should work for upgrades as well. After all, the grub2 package for that upgraded version will contain this fix.

It is indeed correct @bk201 . this won't be reflected to nodes until they do a full reset from recovery ( at least, the bash implementation was up to, I've just checked and indeed https://github.com/rancher-sandbox/elemental/blob/89d323a36cd0234e462bc189010069b29b323129/pkg/action/reset.go#L208 is still the case)

Issue #1225, opened 2022-04-11
xorriso failure when trying to build my 1st derivative  iso standard image from examples/standard 


j-landru:
**cos-toolkit version:**

Latest, downloaded using  `git clone` command

**CPU architecture, OS, and Version:**

`Linux hlandru-ptux04 5.10.61-gentoo #1 SMP Mon Aug 30 11:05:59 CEST 2021 x86_64 Intel(R) Core(TM) i7-7600U CPU @ 2.80GHz GenuineIntel GNU/Linux`

**Describe the bug**

Trying to learn cOS-toolkit by building my 1st derivative iso live image from unmodified standard example, I'm unable to generate iso either with  syslinux as well as grub2. Squashfs is built but xorriso step failed displaying.

- from syslinux iso.yaml file

`FAILURE : Given path does not exist on disk: -boot_image system_area='/tmp/luet-geniso2259780003/tempISO/boot/syslinux/isohdpfx.bin'`

- from  grub2 iso.yaml file

`FAILURE : Given path does not exist on disk: -boot_image system_area='/tmp/luet-geniso3847620838/tempISO/boot/x86_64/loader/boot_hybrid.img'`

Reading the RTFM, and googling don't help me, so sorry if  it's not the good space to post such beginner issue. What am I missing within the doc or doing wrong ?

Thanks

**To Reproduce**

Build standard image from `/examples/standard/Dockerfile` en push it on personal dockerhub repository
1. `docker build the image`
2. `docker push` the image to some registry
3. set iso.yaml file copying/pasting from doc examples

Build the iso file following "Build iso"
` docker run -v $PWD:/cOS -v /var/run:/var/run --entrypoint /usr/bin/luet-makeiso -ti --rm quay.io/costoolkit/toolchain ./iso.yaml --image jlandru/example`

**Expected behavior**

Obtaining an iso image file to VM boot it in live mode.

**Logs**

- Log using syslinux iso.yaml : 
`.
.
.
INFO[0443] 🍹 Generate ISO cOS-0.20220411.iso            
xorriso 1.4.6 : RockRidge filesystem manipulator, libburnia project.

Drive current: -outdev 'stdio:cOS-0.20220411.iso'
Media current: stdio file, overwriteable
Media status : is blank
Media summary: 0 sessions, 0 data blocks, 0 data, 15.5g free
Added to ISO image: directory '/'='/tmp/luet-geniso2259780003/tempISO'
[...]

PR #1233, opened 2022-04-13
Add experimental package with openrc services

mudler:
Those are taken from
https://github.com/c3os-io/c3os/tree/94e6d6fa67a6ff25be8f146f892ebeaaf865ebc1/overlay/files-alpine and seems to be working, providing minimum functionality for alpine(and openrc based) support. I'd like to upstream them here so to avoid re-implementation spreading across.

### DIFF
--- a/packages/openrc/cos-setup/build.yaml
+++ b/packages/openrc/cos-setup/build.yaml
@@ -1,0 +1,9 @@
+requires:
+- name: "base"
+  category: "distro"
+  version: ">=0"
+output_dir: "/package"
+steps:
+- mkdir /package
+- cp -rfv files/* /package
+- chmod +x /package/usr/bin/cos-setup-reconcile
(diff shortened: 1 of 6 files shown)


Itxaka (approved):
not familiar with openrc but looks good

j-landru:
@mudler Thanks

As I said in https://github.com/rancher-sandbox/cOS-toolkit/issues/1225, I'm a k3os orphan user and I plan to use cos as an alternative to build an immutable iso file to deploy an IPv6 only k3s cluster. But I would like to stay on Alpine distro as for k3os. As a more than 20 years Gentoo user,  I appreciate openrc init alternative even if I'm not against dominant systemd.
I won't miss to follow this PR and send feedback in case of issue

P.S. : In case of Alpine adoption as one of the available cos distro, which color to name Alpine cos flavour. ?

PR #1249, opened 2022-04-22
Cleanups rdeps

mudler:
Fixes: https://github.com/rancher-sandbox/cOS-toolkit/issues/1232

PR #1250, opened 2022-04-22
Use final images

mudler:
This is to show how would look like with https://github.com/rancher-sandbox/cOS-toolkit/pull/1249#issuecomment-1106190253 and mainly to check base if it makes sense cc @Itxaka @davidcassany @mjura .

Based on top of https://github.com/rancher-sandbox/cOS-toolkit/pull/1249

Note the difference here is we don't pull anymore the build artifacts while publishing, as all should be already in place for the indexing to kick off.

### DIFF
--- a/.github/build.go
+++ b/.github/build.go
@@ -133,6 +133,8 @@ func main() {
 
 	missingPackages := []client.Package{}
 
+	fatalErrors := os.Getenv("DOWNLOAD_FATAL_MISSING_PACKAGES") == "true"
+
 	for _, p := range packs.Packages {
 		if !client.Packages(currentPackages.Packages).Exist(p) {
 			missingPackages = append(missingPackages, p)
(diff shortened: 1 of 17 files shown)


Itxaka:
how about the staging env that we talked about? After all upload/download is the fastest path rather than rebuild the package, would that also work in this case? Like we publish to a staging env, and then on publish we download and push to final repo, generate index?

Im also worried about the repo size growing stupidly high (more than what we current have) if all CI packages keep getting pushed there. i.e. autobumper will push new packages every night, wouldn't it?

I like the speed up thought.

davidcassany:
Would it really speed that much? isn't #1249 the real boost on CI? Looks like we will only gain the time to upload and download binaries within github jobs, am I right? If so I'd say this is not that relevant, making the build faster is the real key. Having a simple update on a low package in the current cos dependency tree takes around an hour to build, if we can minimize that without having to rebuild the whole tree and just the tow affected packages (the updated one and cos) this is the real win.

mudler:
> how about the staging env that we talked about? After all upload/download is the fastest path rather than rebuild the package, would that also work in this case? Like we publish to a staging env, and then on publish we download and push to final repo, generate index?

Indeed the staging env looks a good idea, it would really help in so many scenarios. definetly something we should look at
> 
> Im also worried about the repo size growing stupidly high (more than what we current have) if all CI packages keep getting pushed there. i.e. autobumper will push new packages every night, wouldn't it?

Not really, this applies only to master builds and the publishing pipelines. PRs, branch builds, etc are excluded

> 
> I like the speed up thought.

Note there is just a tiny speed up. the real gain of this is that also the build host publishes the artifacts - so we have guarantee there is nothing in the middle happening during uploads

mudler:
> Would it really speed that much? isn't #1249 the real boost on CI? Looks like we will only gain the time to upload and download binaries within github jobs, am I right? If so I'd say this is not that relevant, making the build faster is the real key. Having a simple update on a low package in the current cos dependency tree takes around an hour to build, if we can minimize that without having to rebuild the whole tree and just the tow affected packages (the updated one and cos) this is the real win.

https://github.com/rancher-sandbox/cOS-toolkit/pull/1249 is the real boost indeed. if you look at all the cOS packages built, now we are around 8mins. that encompassess all the images flavors (container, recovery img,  etc), so the difference is already noticeable (it was around ~25/30min)

PR #1251, opened 2022-04-22
Add NetworkManager package to cOS and derivatives

### LINKED ISSUE #1228: Explore cOS and derivatives with NetworkManager
This card is about exploring needed configuration to make cOS and derivatives functional with NetworkManager.

This spans to:
- initrd modules
- runtime default configuration
- changes that might be needed to https://github.com/rancher-sandbox/cOS-toolkit/blob/master/packages/cloud-config/oem/05_network.yaml (our default network cloud-init config)

mjura:
Add NetworkManager package to provide automatic networks detection and configuration for systems.

Fixes #1228

Signed-off-by: Michal Jura <mjura@suse.com>

### DIFF
--- a/packages/cos/collection.yaml
+++ b/packages/cos/collection.yaml
@@ -2,7 +2,7 @@ packages:
   - &cos
     name: "cos"
     category: "system"
-    version: 0.8.9-18
+    version: 0.8.9-20
     description: "cOS base image, used to build cOS live ISOs"
     brand_name: "cOS"
     labels:
--- a/examples/odroid-c2/Dockerfile
+++ b/examples/odroid-c2/Dockerfile
@@ -40,6 +40,7 @@ RUN zypper in -y \
     mdadm \
     multipath-tools \
     nano \
+    NetworkManager \
     nfs-utils \
     odroidc2-firmware \
     open-iscsi \
@@ -61,6 +62,11 @@ RUN zypper in -y \
     wpa_supplicant
 
 RUN zypper cc
+
+# Configure NetworkManager as default network management service
+RUN zypper remove -y wicked
+RUN systemctl disable wicked \
+    && systemctl enable NetworkManager
 
 # Copy the luet config file pointing to the upgrade repository
 COPY conf/luet.yaml /etc/luet/luet.yaml
--- a/examples/rpi/Dockerfile
+++ b/examples/rpi/Dockerfile
@@ -38,6 +38,7 @@ RUN zypper in -y \
     mdadm \
     multipath-tools \
     nano \
+    NetworkManager \
     nfs-utils \
     open-iscsi \
     open-vm-tools \
@@ -59,6 +60,11 @@ RUN zypper in -y \
     wpa_supplicant
 
 RUN zypper cc
+
+# Configure NetworkManager as default network management service
+RUN zypper remove -y wicked
+RUN systemctl disable wicked \
+    && systemctl enable NetworkManager
 
 # Copy the luet config file pointing to the upgrade repository
 COPY conf/luet.yaml /etc/luet/luet.yaml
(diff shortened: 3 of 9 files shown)


davidcassany (approved):
Looks good to me, even I can't understand why tests were he triggered up with this PR 🤔 

Also I am not sure we need to bump `base` package version (in `packages/base/definition.yaml`) and `cos` package version (in `packages/cos/collection.yaml`) to ensure this is eventually released.

I am fine switching to NetworkManager, it looks like this is what we will end up using in short/mid term.

mudler (changes requested):
This change requires a bump of the cOS packages to propagate to vanilla images.

The cOS image version can be set here: https://github.com/rancher-sandbox/cOS-toolkit/blob/master/packages/cos/collection.yaml#L5 (note, we increment the number after the `dash` when there are no notable changes except when revbumps are necessary, otherwise we stick with standard semver)

mudler:
Some questions pops on top of my head: 

- are there any changes required for derivatives to use NetworkManager by default? if yes, what are those? would just enabling the service be enough?
- are there any steps required to disable wicked entirely? or we just have support for both in that way? My guess is that zypper will figure which one to enable, but we probably might want to have a dedicated section in our docs about this
- do derivatives need to do any change in order to consume NetworkManager, if yes, what are those?

these questions sounds a good candidate for our docs to give indication as well for who is writing derivatives

mudler:
Also, does https://github.com/rancher-sandbox/cOS-toolkit/blob/master/packages/cloud-config/oem/05_network.yaml require any adaptation? do we need anything from cloud-init side? If that's the case I'd probably rename `cloud-config/network` to `cloud-config/wicked-network` and add a `provide` section for `cloud-config/network`

mjura:
I have did some additional tests and investigation, it seems that I did get functionality of `NetworkManager` at the beginning.

Here are answers to questions from previous comments:
* besides installing `NetworkManager` needs to be only installed and service needs to be enabled in `systemd` - both changes I have added to this PR
* to disable `wicked` we have to stop service and disabled it with `systemd`, after it package can be removed - this change also includes this step
* besides installing and enabling systemd service we don't have to do anything, everything rest will work out-of-the-box

mjura:
> Also, does https://github.com/rancher-sandbox/cOS-toolkit/blob/master/packages/cloud-config/oem/05_network.yaml require any adaptation?

Nope, part which you have pointed me it is for initramfs stage, for my understanding this has to stay as it is.

mjura:
> Also I am not sure we need to bump `base` package version (in `packages/base/definition.yaml`) and `cos` package version (in `packages/cos/collection.yaml`) to ensure this is eventually released.

I have updated version for `packages/base/definition.yaml` and `packages/cos/collection.yaml`

mudler (approved):
Looks good!

mudler:
> I have did some additional tests and investigation, it seems that I did get functionality of `NetworkManager` at the beginning.
> 
> Here are answers to questions from previous comments:
> 
>     * besides installing `NetworkManager` needs to be only installed and service needs to be enabled in `systemd` - both changes I have added to this PR
> 
>     * to disable `wicked` we have to stop service and disabled it with `systemd`, after it package can be removed - this change also includes this step
> 
>     * besides installing and enabling systemd service we don't have to do anything, everything rest will work out-of-the-box

That's perfect, then the examples should be good enough to show up how to do this  :+1:

mjura:
PR was rebased

mudler (approved):

packages/cos/collection.yaml
```diff
@@ -2,7 +2,7 @@ packages:
   - &cos
     name: "cos"
     category: "system"
-    version: 0.8.9-18
+    version: 0.8.9-19
```
mudler, line 5: ouch, looks like the rebase merged the bump too, so now that's not being picked up by CI

mjura: I was thinking that we want to increment this version? how should I fix it then ?

Issue #1238, opened 2022-04-14
Create flavor for SLE Micro for Rancher


mudler:
This card is about to switch our current supported flavor (green) from leap to SLE Micro for Rancher ( http://registry.suse.com/suse/sle-micro-rancher/5.2 ).

I'd propose then to have a new flavor (e.g. `teal` which is the color result of mixing green and blue  ) and switch our current CI tests over that and use it as primary base image. this might be challenging for several aspects (ARM, etc) so this card is a spike to see how much we can get from there at start, by adding an additional flavor based on it at least.

With this in place then we could move the `green` variant to track latest leap versions, or either even tumbleweed if we have feature parity with SLE Micro for Rancher/leap.

## Action items

- [x] Add  new flavor for teal https://github.com/rancher-sandbox/cOS-toolkit/tree/master/values
- [x] Adapt CI to push to new flavor (skipping tests suites for now, treating like `orange` and `blue`)  https://github.com/rancher-sandbox/cOS-toolkit/tree/master/.github/config
- [x] Do any required changes to fixup build (if needed)

mudler:
ok, seems I'm partly hitting https://github.com/rancher-sandbox/cOS-toolkit/issues/1240 as arm is completely stuck, going to try force using luet from upstream

PR #1259, opened 2022-04-26
Add teal flavor

mudler:
Add teal flavor to our pipelines.

Part of https://github.com/rancher-sandbox/cOS-toolkit/issues/1238


mudler:
I'd go ahead with this and check out manually the artifacts to see if there are any issues and will double check with the ones we currently have for green, I bet there will be some small adjustments to do

packages/kernel/build.yaml
```diff
@@ -3,6 +3,16 @@ requires:
   category: "distro"
   version: ">=0"
 
+excludes:
+- ^/boot/grub2
```
mudler, line 7: I think those two previously just slipped in, no real need to add them as part of the kernel package

packages/utils/gomplate/build.yaml
```diff
@@ -7,16 +7,9 @@ package_dir: /{{.Values.name}}
 
 env:
 - PATH=$PATH:/usr/local/go/bin
+- CGO_ENABLED=0
```
mudler, line 10: somehow those slipped in :man_shrugging:

packages/kernel/build.yaml
```diff
@@ -3,6 +3,16 @@ requires:
...
+- ^/boot/initrd
+{{ if .Values.codename }}
+{{if eq .Values.codename "teal" }}
+unpack: true
+includes:
+- ^/boot|^/boot/.*
```
mudler, line 13: The real difference in teal is that we extract all the kernels from the image, so no zypper in or build happening

packages/templates/distro.yaml
```diff
@@ -1,4 +1,6 @@
 {{ define "distro_install" }}
+## All packages for teal are in the teal image already.
+{{ if ne .Values.codename "teal" }}
```
mudler, line 3: This is gating accidental calls to `zypper in..` if we are in `teal`. supposedly, we don't add any repo to it

mjura (approved):
LGTM

PR #1262, opened 2022-04-27
Add a deps target for CI 

Itxaka:
This target will always try to upgrade the packages in the system with
latest versions from our repo

### DIFF
--- a/.github/build.yaml.gomplate
+++ b/.github/build.yaml.gomplate
@@ -72,7 +72,7 @@
   {{{ $target := . }}}
       - name: Run make {{{ $target }}}
         run: |
-          {{{ if eq $target "deps" }}}export DOCKER_INSTALL=true{{{end}}}
+          {{{ if eq $target "deps_ci" }}}export DOCKER_INSTALL=true{{{end}}}
           sudo -E make {{{ $target }}}
   {{{ if eq $target "deps" }}}
   {{{- if ne $config.luet_override "" }}}
@@ -162,7 +162,7 @@
       - name: Login to Quay Registry
         run: echo ${{ secrets.QUAY_PASSWORD }} | docker login -u ${{ secrets.QUAY_USERNAME }} --password-stdin quay.io
   {{{- end }}}
-      {{{ tmpl.Exec "make" "deps" }}}
+      {{{ tmpl.Exec "make" "deps_ci" }}}
       {{{ tmpl.Exec "make" "validate" }}}
       - name: Build packages 🔧
         run: |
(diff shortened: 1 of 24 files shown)


make/Makefile.iso
```diff
@@ -82,7 +82,7 @@ else
 endif
 
 .PHONY: iso
-iso: $(LUET) $(YQ) $(MAKEISO) $(MKSQUASHFS)
+iso: $(LUET) yq $(MAKEISO) $(MKSQUASHFS)
```
davidcassany, line 85: I guess `$(MAKEISO)` here should be substituted to `makeiso` right?

davidcassany (approved):
LGTM, just a little leftover I think

make/Makefile.iso
Itxaka: Good catch!

kkaempf:
Not sure if this is the right approach. The luet installs might also override system binaries (coming e.g. from RPM).
While this PR solves one problem, I fear it might trigger many others.

Itxaka:
> Not sure if this is the right approach. The luet installs might also override system binaries (coming e.g. from RPM).
While this PR solves one problem, I fear it might trigger many others.

I guess this affects building on OBS? How does obs tackles dependency? Not calling luet install, Im sure.

Well, we can do something else, keep the make deps as usual, and have a make deps_ci which does what this patch does. Would that work for you @kkaempf ?

kkaempf:
> I guess this affects building on OBS?
No, not at all.
This affects a local system where binaries are installed from RPM or DEP. Just 'blindly' calling "luet install ..." now likely clobbers these installs.

> 
> Well, we can do something else, keep the make deps as usual, and have a make deps_ci which does what this patch does. 

That should work, thanks.

davidcassany:
> Not sure if this is the right approach. 

On second thought I think this a good point. We should limit this to the toolchain we provide in  `packages/toolchain` so I would exclude `jq` from this PR. 

> The luet installs might also override system binaries (coming e.g. from RPM).

This is a good point I think it should be considered a bug in our toolchain packages, the luet packages should not interfere with whatever the system provides. However this is hard to solve in a sane way, a common solution would be to set luet packages to install under `/usr/local/` buit this collides with our immutable layout approach. So I am not sure there is any other way to make sure we minimize the use of luet packages and ensure those are not provided neither conflict with the system.

Itxaka:
ok, this should work, this only runs on our CI and tries to upgrade all packages in the deps. Should be safe enough to run only on our ci while leaving the make deps target untouched so it can be used locally or on OBS.

davidcassany (approved):
Looks good

PR #1265, opened 2022-04-28
debug

Itxaka:


mudler:
@Itxaka there is an unrelated issue with the QEMU tests which may impact you in building phase. since the `luet` version in the `arm64` tree is really behind, after #1249  was merged building cos images doesn't work as requires `0.31.2` at least. I'm trying to workaround it somehow because It's getting in the way with #1259

PR #1266, opened 2022-04-28
ci: add option to force install from upstream repo

mudler:
Bump to latest luet for bootstrap

### DIFF
--- a/.github/build.yaml.gomplate
+++ b/.github/build.yaml.gomplate
@@ -72,14 +72,24 @@
   {{{ $target := . }}}
       - name: Run make {{{ $target }}}
         run: |
-          {{{ if eq $target "deps_ci" }}}export DOCKER_INSTALL=true{{{end}}}
+          {{{ if eq $target "deps_ci" }}}
+          export DOCKER_INSTALL=true
+          {{{end}}}
           sudo -E make {{{ $target }}}
-  {{{ if eq $target "deps" }}}
+  {{{ if eq $target "deps_ci" }}}
   {{{- if ne $config.luet_override "" }}}
       - name: Override luet version at toolchain/luet@{{{ $config.luet_override }}} 🔧
         run: |
           sudo luet replace --yes toolchain/luet --for toolchain/luet@{{{ $config.luet_override }}}
   {{{  end }}}
+  {{{ if ne $config.luet_script_override "" }}}
+      - name: Override luet version at {{{ $config.luet_script_override }}} 🔧
+        run: |
+          export DOCKER_INSTALL=false
+          export LUET_INSTALL_FROM_COS_REPO=false
+          export LUET_VERSION={{{$config.luet_script_override}}}
+          sudo -E scripts/get_luet.sh
+  {{{ end }}}
   {{{  end }}}
 {{{end}}}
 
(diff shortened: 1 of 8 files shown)


mudler:
@Itxaka I could also reuse this branch to directly publish to arm64 bypassing tests if we make the build work, wdyt?

.github/build.yaml.gomplate
```diff
@@ -72,8 +72,19 @@
...
       - name: Run make {{{ $target }}}
         run: |
-          {{{ if eq $target "deps_ci" }}}export DOCKER_INSTALL=true{{{end}}}
+          {{{ if eq $target "deps_ci" }}}
+          export DOCKER_INSTALL=true
+          {{{end}}}
+          {{{ if ne $config.luet_script_override "" }}}
+          export DOCKER_INSTALL=false
+          export LUET_INSTALL_FROM_COS_REPO=false
+          export LUET_INSTALL_FROM_COS_REPO=false
```
Itxaka, line 81: duplicated entry

.github/config/nightly.yaml
```diff
@@ -24,6 +24,7 @@ flavors:
         build_image_qemu: true
         build_image_vbox: true
         build_raw_image: true
+        luet_script_override: ""
```
Itxaka, line 27: so the value is false or empty?

mudler: I just changed it slightly, it is a string that you can use to specify a luet version to use directly :+1:

.github/workflows/build-master-blue-arm64.yaml
```diff
@@ -62,10 +62,18 @@ jobs:
       - name: Run make deps_ci
         run: |
           export DOCKER_INSTALL=true
+          export DOCKER_INSTALL=false
```
Itxaka, line 65: duplicated entry here too int he final output

mudler: argh there it's a bit not nice from the template itself, there the first one is coming from another option, not sure if want to make it too tight to the override

.github/workflows/build-master-blue-arm64.yaml
```diff
@@ -62,10 +62,18 @@ jobs:
...
+          export LUET_INSTALL_FROM_COS_REPO=false
+          export LUET_VERSION=0.32.0
           sudo -E make deps_ci
```
Itxaka, line 68: shouldnt deps_ci call the get_luet script? Like it does it already no?

mudler: yup it does, but then when we call make deps we overwrite it with the version coming from the repos.. it's not nice, we re-install it twice here

mudler: reworked it slightly, at least now should be a bit more obvious

mudler: and I broke it somehow :D

mudler: oh I see, ci is calling `deps_ci` now. adapted it

mudler:
@Itxaka arm64 repo were published, adapted to drop the overrides but keep the handles in the pipeline config (if we might need that in future)

PR #1269, opened 2022-04-28
Pin packer version

Itxaka:
Seems that newer versions dont work properly with qemu user networking,
making the vm not get an ip

Pinning to an older version seems to make it work until the problems are
fixed

### DIFF
--- a/.github/build.yaml.gomplate
+++ b/.github/build.yaml.gomplate
@@ -276,7 +276,7 @@
       PKR_VAR_arch: {{{ $config.arch }}}
       PKR_VAR_flavor: {{{ $flavor }}}
       PKR_VAR_feature: vagrant
-      PKR_VAR_accelerator: none
+      PKR_VAR_accelerator: tcg
       PKR_VAR_cpus: 2
       PKR_VAR_memory: 4096
       {{{if eq $config.arch "arm64" }}}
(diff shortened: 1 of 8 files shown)


packer/images.json.pkr.hcl
```diff
@@ -114,6 +114,7 @@ source "qemu" "cos-arm64" {
   disk_size              = "${var.disk_size}"
   format                 = "qcow2"
   headless               = true
+  net_bridge             = "virbr0"
```
mudler, line 117: now... it starts to need this?! seriously?! :rofl:

mudler: good catch btw!

mudler (approved):

packer/images.json.pkr.hcl
Itxaka: this could be because package by default uses user networking which on qemu is...complicated. Lots of variance in configs in the different distros and based on the configured system, so for all we know a simple hardware update could cause this or a newer qemu version, or a different network config for the system, etc...


Its kind of a pita :/

mudler (approved):

mudler:
@Itxaka now the QEMU part is passing! I'd say to merge it anyway and see the specific issue which fails now separately, it looks like there is a typo in the repository which is missing a dash

```
          INFO[2022-04-29T17:37:51Z] Copying COS_ACTIVE image...                  
           INFO    Downloading  quay.io/costoolkit/releases-greenarm64:repository.yaml
```

but nevertheless is unrelated to this, looks like it might have slipped before as arm64 was failing since a while already..

PR #1276, opened 2022-05-02
Automatic Boot assessment

mudler:
This PR adds automatic boot assessment and a test case for it.

Besides it also fix a pipeline issue to allow ISO to build again


packages/cloud-config/oem/08_boot_assessment.yaml
```diff
@@ -0,0 +1,111 @@
...
+          [ ! -f "/tmp/mnt/STATE/grubcustom" ] || ( cat /tmp/mnt/STATE/grubcustom | grep -v "grub_boot_assessment" )
+      commands:
+      - |
+        cat << EOF >> /tmp/mnt/STATE/grubcustom
+        set bootfile="/grub_boot_assessment"
+        search --no-floppy --file --set=bootfile_loc "${bootfile}"
+        if [ "${bootfile_loc}" ]; then
+           source "(${bootfile_loc})${bootfile}"
+        EOF
+    # Overrides the active cmdline by adding "rd.emergency=reboot", "rd.shell=0" and "panic=5"
```
mudler, line 58: @davidcassany in my tests, I managed to get the initramfs stuck by removing "usr', "sbin", "lib" from the image. It stayed to initramfs stuck after I could see it mounted the images (the debug messages where displaying ext4 mount info text blobs)

I then tried to reduce the dracut timeout to something smaller without success (with `rd.timeout`), so it seems in certain cases the initramfs stays still in the foreground trying to do something (I couldn't check what exactly). However, didn't seemed related to any dracutinitqueue timeout afaict, but any pointer might be appreciated.

The thing is, If we have an option to let dracut start in a time range would be good as well. so in any case we can restart into fallback even after a long time passed.

Anyway, this mechanism seems to be flexible enough, as a reboot would automatically select fallback in that case as well. it is a matter on how dracut is configured at that point

tests/fallback/fallback_test.go
```diff
@@ -22,6 +25,75 @@ var _ = Describe("cOS booting fallback tests", func() {
...
+			// Break the upgrade
+			out, _ = s.Command("sudo mount -o rw,remount /run/initramfs/cos-state")
+			fmt.Println(out)
+
+			out, _ = s.Command("sudo mkdir -p /tmp/mnt/STATE")
+			fmt.Println(out)
+
+			s.Command("sudo mount /run/initramfs/cos-state/cOS/active.img /tmp/mnt/STATE")
+
+			for _, d := range []string{"usr/lib/systemd"} {
```
mudler, line 73: here we trigger boot assessment by deleting systemd from the image after the upgrade was done

packages/cloud-config/oem/08_boot_assessment.yaml
```diff
@@ -0,0 +1,112 @@
...
+      - |
+          sudo mount -o rw,remount /run/initramfs/cos-state
+          grub2-editenv /run/initramfs/cos-state/boot_assessment set osupgrade=
+          grub2-editenv /run/initramfs/cos-state/boot_assessment set tentative=
+          sudo mount -o ro,remount /run/initramfs/cos-state
+    - name: "Create upgrade failure sentinel if necessary"
+      if: |
+          cat /proc/cmdline | grep -q "upgrade_failure"
+      files:
+      - path: /run/cos/upgrade_failure
```
mudler, line 27: This adds a new sentinel file, `/run/cos/upgrade_failure` when we are booting from fallback after a failed upgrade. Needs some rooms in our docs

mudler:
blocked by #1287

packages/cloud-config/oem/08_boot_assessment.yaml
```diff
@@ -0,0 +1,112 @@
...
+    # Remove any grub sentinel and add
+    # System sentinel to /run to indicate if we are booting after a failed upgrade.
+    # This needs to run only on the active partition.
+    - name: "Remove upgrade sentinels"
+      if: |
+          cat /proc/cmdline | grep -q "active.img"
+      commands:
+      - |
+          sudo mount -o rw,remount /run/initramfs/cos-state
+          grub2-editenv /run/initramfs/cos-state/boot_assessment set osupgrade=
```
mudler, line 20: I'm wondering if renaming `osupgrade` to something more generic, like `boot_assessment`. In this way we don't tie implementation to the use case, as we could also trigger this in other situations as well

packages/cloud-config/oem/08_boot_assessment.yaml
davidcassany: mm we should analyze this specific case. IIRC `rd.timeout` is only honored by the `initqueue` so if your problem was before/beyond the initqueue then it makes sense rd.timeout was simply ignored.

Note the boot is systemd based, so it can happen some early boot services are in starting state forever, like mountpoints without a timeout or with an insane timeout. So it is likely we need to work per use case basis... this is hard, since there things that are not directly configured by us. 

We could try to analyze the boot procedure and try to find services that could stuck forever, note this has potential to be distro specific. 

Dracut bootup procedure is documented here https://man7.org/linux/man-pages/man7/dracut.bootup.7.html

tests/fallback/fallback_test.go
mudler: NOTE: triggers a kernel panic

packages/cloud-config/oem/08_boot_assessment.yaml
```diff
@@ -0,0 +1,114 @@
...
+    boot.before:
+    # Remove any grub sentinel and add
+    # System sentinel to /run to indicate if we are booting after a failed upgrade.
+    # This needs to run only on the active partition.
+    - name: "Remove upgrade sentinels"
+      if: |
+          cat /proc/cmdline | grep -q "active.img"
+      commands:
+      - |
+          sudo mount -o rw,remount /run/initramfs/cos-state
```
davidcassany, line 24: sudo? isn't it executed by root already?

packages/cloud-config/oem/08_boot_assessment.yaml
mudler: doing it in a new commit

davidcassany (approved):
Looks good to me

Even I am wondering if shoudn't we reuse the current grub environment files so no need to rw remount which feels quite hackish. 

Also we could think on making the emergency service to reboot by default unless we are in debug mode. 

Anyway I like the feature and I think this is nice. I am wondering if shouldn't we make it more integrated in our source code (as a dracut module) and make it opt in/out.

packages/cloud-config/oem/08_boot_assessment.yaml
mudler: I've been able to trace back my attempt and the specific case is triggered by removing the following folders from the image and trying it to boot: "bin", "usr", "sbin", "lib".

That triggers a boot stuck which doesn't seem to honor `rd.timeout`. Note I can see the mounting messages of the images, so we get past to the point where images are loaded. Interestingly, doesn't seems to trigger a kernel panic.

packages/cloud-config/oem/08_boot_assessment.yaml
mudler: argh right, that slipped in by some copy paste

mudler:
> Looks good to me
> 
> Even I am wondering if shoudn't we reuse the current grub environment files so no need to rw remount which feels quite hackish.

mmm you mean to use `/grubenv`? wouldn't be still needed to rw remount it? I didn't used that one as left for derivatives/developers
> 
> Also we could think on making the emergency service to reboot by default unless we are in debug mode.
> 
> Anyway I like the feature and I think this is nice. I am wondering if shouldn't we make it more integrated in our source code (as a dracut module) and make it opt in/out.

very good point indeed, at some point this def. have room in the initrd and would make it much solid. It would still need some touchbase from the cloud-init mechanism to assess that boot goes fine ...so on the other hand having it "compressed" in one file makes it portable. For the moment it is a basic playground to see how much we can assess from this, Indeed some pieces would have preferably room in the initrd dracut modules, but as it is now it's really easy to opt-in/out for that.

tests/fallback/fallback_test.go
mudler: this is weird. on c3os it triggered a kernel panic while on cOS it just makes systemd go into freeze mode. Anyhow, I've added `systemd.crash_reboot` so it does restart when PID 1 goes into freeze infinite loops.

mudler:
Ok, seems I got all the pieces - hopefully tests should pass now. There were just few minor issues with the hook installation.

Weirdly, when I was trying this test case on c3os it was triggering a Kernel panic, but here it doesn't , we get systemd freezing because of missing files. I did add `systemd.crash_reboot` among the other initrd options (`panic=5, rd.emergency=reboot`). Ideally we want to reboot asap in case of boot failures in the active partition.

Itxaka:
This broke arm64 fallback tests...somehow?

mudler:
> This broke arm64 fallback tests...somehow?

yeah :/ looks like on arm... grub fallsback instead of booting into initramfs :shrug: might be the different GRUB version?

Itxaka:
Yeah, could be. I added as skip for those tests on arm64 in my PR for now https://github.com/rancher-sandbox/cOS-toolkit/pull/1294

Itxaka:
weird, seems like in arm64 boxes, the boot assesment file was not installed, so no 08_assesment.yaml file in /system/oem/ .... tested with the artifacts from master arm 64 https://github.com/rancher-sandbox/cOS-toolkit/runs/6310310329?check_suite_focus=true

mudler:
> 

that would be quite weird. In the earlier phases of tests it is checked that boot assessment is installed, and it seems to just fail down afterwards.... so the assessment logic is installed into grub

Itxaka:
blarh, the reason was the libvirt-vagrant provider again, as it doesnt really remove the box. So much time without using vagrant directly that I forgot!

PR #1281, opened 2022-05-03
Export LUET_ARCH in iso jobs

Itxaka:


Itxaka:
Seems that after the changes in the luet install, x86 repos are being set on arm64 runs during the iso creation, due to luet forcing the deps install. As before we didnt force the install/upgrade, and the binaries were already in there, we were not hitting this issue.

Itxaka:
wrong job!

PR #1284, opened 2022-05-03
Install deps before adding local repo

Itxaka:
As local repo can be of a different arch, we want to install the deps
before ading it, in case the local repo is different fromt eh runner
arch


Itxaka:
umm, so this works but we also lose the ability to use the latest packages from the build. Not good enough :/

I still think we should merge this until we find a proper solution so we can keep building arm64 for the moment.

mudler (approved):

mjura (approved):
Looks ok

PR #1289, opened 2022-05-03
Do not add final repo for local-iso build target

### LINKED ISSUE #1282: elemental build-iso: wrong repo priority order
Seems that during the PR builds while building the iso we are picking up the wrong repository priority order, and as a result PR and CI are testing remote packages than the local changes.

See e.g. : https://github.com/rancher-sandbox/cOS-toolkit/pull/1276

davidcassany:
Fixes #1282

### DIFF
--- a/.github/build.yaml.gomplate
+++ b/.github/build.yaml.gomplate
@@ -248,7 +248,7 @@
       - name: Build ISO from local build 🔧
         if: github.event_name != 'schedule' && !startsWith(github.ref, 'refs/tags/')
         run: |
-          sudo -E MAKEISO_ARGS="--name cOS-Seed-{{{ $flavor }}}-${{ env.COS_VERSION }}-{{{ $config.arch }}} make local-iso
+          sudo -E MAKEISO_ARGS="--name cOS-Seed-{{{ $flavor }}}-${{ env.COS_VERSION }}-{{{ $config.arch }}}" make local-iso
       - name: Build ISO from remote repositories 🔧
         if: github.event_name == 'schedule' || startsWith(github.ref, 'refs/tags/')
         run: |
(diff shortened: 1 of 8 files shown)


mudler:
mm wait, this is removing the final-repo at all while we build the local-iso, right? but how it is getting the missing packages that are not built locally?

.github/build.yaml.gomplate
```diff
@@ -248,7 +248,7 @@
       - name: Build ISO from local build 🔧
         if: github.event_name != 'schedule' && !startsWith(github.ref, 'refs/tags/')
         run: |
-          sudo -E MAKEISO_ARGS="--name cOS-Seed-{{{ $flavor }}}-${{ env.COS_VERSION }}-{{{ $config.arch }}} --repo ${{ env.FINAL_REPO }}" make local-iso
```
mudler, line 251: double quote was accidentally removed I think

davidcassany:
> mm wait, this is removing the final-repo at all while we build the local-iso, right? but how it is getting the missing packages that are not built locally?

yes good point, see https://github.com/rancher-sandbox/elemental/blob/62b0cfd19fcdf4e059e908ce9b7f861541daef58/cmd/build-iso.go#L112-L123

`elemetal build-iso` creates a repo list from manifest or from defaults (if no manifest). Then the `--repo` flag appends to this repository list with priority set to `1` (while default priority in manifest or default repo is `90`).

`$FINAL_REPO` is already the default repository. I guess I forgot to remove it from here when I added all defaults in elemental.

mudler (approved):
looks good!

.github/build.yaml.gomplate
```diff
@@ -248,7 +248,7 @@
...
+          sudo -E MAKEISO_ARGS="--name cOS-Seed-{{{ $flavor }}}-${{ env.COS_VERSION }}-{{{ $config.arch }}} make local-iso
```
mudler, line 251:
```suggestion
          sudo -E MAKEISO_ARGS="--name cOS-Seed-{{{ $flavor }}}-${{ env.COS_VERSION }}-{{{ $config.arch }}}" make local-iso
```

PR #1293, opened 2022-05-03
Fix PACKAGES selection from makefile

Itxaka:
When we changed the makefile we didnt take into account the PACKAGES
selection

also uses the new --arch flag to build the arm Isos

### DIFF
--- a/make/Makefile.build
+++ b/make/Makefile.build
@@ -56,9 +56,9 @@ _VALIDATE_OPTIONS?=-s
 #
 YQ_VERSION=$(shell yq -V | cut -d " " -f 3 | cut -d "." -f 1)
 ifeq ("$(YQ_VERSION)", "3")
-	PACKAGES?=$(shell yq r -j $(MANIFEST) | jq -r '.packages[],.build | .[]' | sort -u)
+	PACKAGES?=$(shell yq r -j $(MANIFEST) | jq -r '.build,.iso[] | .[]' | sort -u)
 else
-	PACKAGES?=$(shell yq e -j $(MANIFEST) | jq -r '.packages[],.build | .[]' | sort -u)
+	PACKAGES?=$(shell yq e -j $(MANIFEST) | jq -r '.build,.iso[] | .[]' | sort -u)
 endif
 
 ifeq ("$(PACKAGES)","")
--- a/make/Makefile.iso
+++ b/make/Makefile.iso
@@ -78,7 +78,7 @@ endif
 ifneq ("$(ISO)","")
 	@echo "'$(ISO) exists, run 'make clean_iso' folled by 'make $@' to recreate"
 else
-	$(ELEMENTAL) --debug --config-dir $(ROOT_DIR) build-iso $(MAKEISO_ARGS) --date --repo $(DESTINATION)
+	$(ELEMENTAL) --debug --config-dir $(ROOT_DIR) build-iso $(MAKEISO_ARGS) --arch $(ARCH) --date --repo $(DESTINATION)
 endif
 
 .PHONY: iso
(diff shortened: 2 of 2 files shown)


.github/build.yaml.gomplate
```diff
@@ -249,11 +248,11 @@
       - name: Build ISO from local build 🔧
         if: github.event_name != 'schedule' && !startsWith(github.ref, 'refs/tags/')
         run: |
-          sudo -E MAKEISO_ARGS="--name cOS-Seed-{{{ $flavor }}}-${{ env.COS_VERSION }}-{{{ $config.arch }}}" make local-iso
+          sudo -E MAKEISO_ARGS="--name cOS-Seed-{{{ $flavor }}}-${{ env.COS_VERSION }}-{{{ $config.arch }}}{{{- if eq $config.arch "arm64"}}} --repo ${{ env.FINAL_REPO }}{{{end}}}" make local-iso
```
davidcassany, line 251: This will collide with the fix on #1289 It might result on ignoring the local builds and use only the FINAL_REPO.

IMHO we need to make elemental aware of the build arch, otherwise having default repos will always be tricky as it will use x86 repos for non x86 archs. We need something like what is discussed in https://github.com/rancher-sandbox/cos-toolkit-docs/pull/82#discussion_r860632614 in think that for arm builds the default repository should be the arm repository. We could add the arch field in `v1.Repository` and filter by arch in Luet interface implementation and make the default include all archs. But then we need a way to tell elemental which is the architecture we are building for.

.github/build.yaml.gomplate
```diff
@@ -249,11 +248,11 @@
...
       - name: Build ISO from remote repositories 🔧
         if: github.event_name == 'schedule' || startsWith(github.ref, 'refs/tags/')
         run: |
-          sudo -E MAKEISO_ARGS="--name cOS-Seed-{{{ $flavor }}}-${{ env.COS_VERSION }}-{{{ $config.arch }}}" make iso
+          sudo -E MAKEISO_ARGS="--name cOS-Seed-{{{ $flavor }}}-${{ env.COS_VERSION }}-{{{ $config.arch }}}{{{- if eq $config.arch "arm64"}}} --repo ${{ env.FINAL_REPO }}{{{end}}}" make iso
```
mudler, line 255: if we add back the final repo here aren't we back at https://github.com/rancher-sandbox/cOS-toolkit/pull/1289 ?

davidcassany (changes requested):
I believe with this setup we will reproduce #1282 issue on arm. The FINAL_REPO for local builds should be the default, not added by the  `--repo` flag, otherwise there is a repo priority issue. Added repositories always have high and equal priority. `make local-iso` already adds the local repository via `--repo` flag.

.github/build.yaml.gomplate
davidcassany: An alternative without touching elemental code could be to set the arm repository with the manifest. Using the manifest is the only way to overwrite default repository and with full priority control, so we could add the arm repository in manifest and this would overwrite the default and set the arm repo with a low priority (default prio would be good, currently set to `90`) to give preference on the local repository that `make local-iso` target adds.

Itxaka: Yes, this is only a workaround for the current build stuff on arm. I need to be able to build the proper isos for now to see if the other( wrong repo in the cloud config yamls) arm issues are now fixed. This is not meant to be merged as is :)

Itxaka:
> I believe with this setup we will reproduce #1282 issue on arm. The FINAL_REPO for local builds should be the default, not added by the  `--repo` flag, otherwise there is a repo priority issue. Added repositories always have high and equal priority. `make local-iso` already adds the local repository via `--repo` flag.

Could you recheck now? Patch has changed to use the --arch from elemental and that now makes it work without any workarounds.

mudler (approved):
Looking good now!

Issue #1003, opened 2021-12-21
Include git sha in repository snapshot names


mudler:
**Is your feature request related to a problem? Please describe.**
Currently repository snapshots are prefixed only by date, which makes difficult to track back to the actual commit/tag. `luet create-repo` takes a `--snapshot-id` argument, so we can use that while pushing with the CI. While promoting snapshots, we can stop also indexing all packages (which would alleviate a bit of work from the CI), so this card is also about adapting our CI.

**Describe the solution you'd like**

- [x] snapshots to include the git sha/tag
- [x] releases to include also a file with the repository references of the release
- [x] stop indexing all packages. as we can rely on snapshots we can just publish the current repository state. https://github.com/rancher-sandbox/cOS-toolkit/blob/c981c45e3fff1f0337d549ad5d8acdf269f69f9b/.github/build.yaml.gomplate#L567

**Describe alternatives you've considered**

**Additional context**
Depends on #985

Itxaka:
https://github.com/rancher-sandbox/cOS-toolkit/pull/1312 for the snapshot-id of both commits and tags

Itxaka:
https://github.com/rancher-sandbox/cOS-toolkit/pull/1314 for  `releases to include also a file with the repository references of the release`

Itxaka:
umm @mudler Im not sure hwo to approach `stop indexing all packages. as we can rely on snapshots we can just publish the current repository state. [master/.github/build.yaml.gomplate#L557](https://github.com/rancher-sandbox/cOS-toolkit/blob/master/.github/build.yaml.gomplate#L557)`

Is that done on the build.go code? Like it should get a repositoryID reference in order to check the artifacts for that or....

Im a bit lost on that part, as I have no idea how we index everything. I guess the DOWNLOAD_ALL stuff will also had something to do with that?

mudler:
> umm @mudler Im not sure hwo to approach `stop indexing all packages. as we can rely on snapshots we can just publish the current repository state. [master/.github/build.yaml.gomplate#L557](https://github.com/rancher-sandbox/cOS-toolkit/blob/master/.github/build.yaml.gomplate#L557)`
> 
> Is that done on the build.go code? Like it should get a repositoryID reference in order to check the artifacts for that or....
> 
> Im a bit lost on that part, as I have no idea how we index everything. I guess the DOWNLOAD_ALL stuff will also had something to do with that?

Sorry the late reply, I've lost this comment. 

[...]

PR #1294, opened 2022-05-05
Set packer qemu plugin to use

Itxaka:
Bump it to use 1.0.3 at least, which has some fixes for efi firmware

also bumps packer to the latest version. Problem seems to be a combination of packer 1.8.0 + qemu plugin 1.0.2 (bundled)

### DIFF
--- a/.github/workflows/build-pr-green-arm64.yaml
+++ b/.github/workflows/build-pr-green-arm64.yaml
@@ -185,7 +185,7 @@ jobs:
       - name: Install Packer
         uses: hashicorp-contrib/setup-packer@v1
         with:
-            packer-version: 1.7.3
+            packer-version: 1.8.0
       - name: Build QEMU Image 🔧
         run: |
           source .github/helpers.sh
@@ -361,7 +361,7 @@ jobs:
       - name: Install Packer
         uses: hashicorp-contrib/setup-packer@v1
         with:
-            packer-version: 1.7.3
+            packer-version: 1.8.0
       - name: Build QEMU Image 🔧
         run: |
           source .github/helpers.sh
--- a/.github/workflows/build-pr-green-arm64.yaml
+++ b/.github/workflows/build-pr-green-arm64.yaml
@@ -208,7 +208,7 @@ jobs:
     env:
       ARCH: arm64
       VAGRANT_CPU: 2
-      VAGRANT_MEMORY: 5120
+      VAGRANT_MEMORY: 4096
       VAGRANT_FIRMWARE: /usr/share/AAVMF/AAVMF_CODE.fd
       COS_HOST: "192.168.122.50:22"
       COS_TIMEOUT: 1800
@@ -385,7 +385,7 @@ jobs:
     env:
       ARCH: arm64
       VAGRANT_CPU: 2
-      VAGRANT_MEMORY: 5120
+      VAGRANT_MEMORY: 4096
       VAGRANT_FIRMWARE: /usr/share/AAVMF/AAVMF_CODE.fd
       COS_HOST: "192.168.122.50:22"
       COS_TIMEOUT: 1800
--- a/.github/workflows/build-pr-green-arm64.yaml
+++ b/.github/workflows/build-pr-green-arm64.yaml
@@ -168,6 +168,7 @@ jobs:
       PKR_VAR_accelerator: tcg
       PKR_VAR_cpus: 2
       PKR_VAR_memory: 4096
+      PACKER_GITHUB_API_TOKEN: ${{ github.token }}
       PKR_VAR_qemu_binary: qemu-system-aarch64
       PKR_VAR_firmware: /usr/share/qemu-efi-aarch64/QEMU_EFI.fd
       PACKER_TARGET: qemu.cos-arm64
@@ -215,7 +216,7 @@ jobs:
     needs: qemu-squashfs-green
     strategy:
       matrix:
-        test: [test-fallback, test-smoke, test-upgrades-images-signed, test-upgrades-images-unsigned, test-upgrades-local]
+        test: [test-smoke, test-upgrades-images-signed, test-upgrades-images-unsigned, test-upgrades-local]
     steps:
       - name: Install Go
         uses: actions/setup-go@v2
@@ -344,6 +345,7 @@ jobs:
       PKR_VAR_accelerator: tcg
       PKR_VAR_cpus: 2
       PKR_VAR_memory: 4096
+      PACKER_GITHUB_API_TOKEN: ${{ github.token }}
       PKR_VAR_qemu_binary: qemu-system-aarch64
       PKR_VAR_firmware: /usr/share/qemu-efi-aarch64/QEMU_EFI.fd
       PACKER_TARGET: qemu.cos-arm64
@@ -391,7 +393,7 @@ jobs:
     needs: qemu-nonsquashfs-green
     strategy:
       matrix:
-        test: [test-fallback, test-smoke, test-upgrades-images-signed, test-upgrades-images-unsigned, test-upgrades-local]
+        test: [test-smoke, test-upgrades-images-signed, test-upgrades-images-unsigned, test-upgrades-local]
     steps:
       - name: Install Go
         uses: actions/setup-go@v2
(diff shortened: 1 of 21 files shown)


tests/fallback/fallback_test.go
```diff
@@ -39,6 +39,11 @@ var _ = Describe("cOS booting fallback tests", func() {
 	}
 
 	Context("image is corrupted", func() {
+		BeforeEach(func() {
+			if s.GetArch() == "arm64" {
+				Skip("Tests not supported on arm64 architecture")
```
mudler, line 44: :+1:

mudler (approved):

mudler:
fun, now packer seems to call github...:
```
Failed getting the "github.com/hashicorp/qemu" plugin:
1 error occurred:
	* Plugin host rate limited the plugin getter. Try again in 14m32.154283s.
HINT: Set the PACKER_GITHUB_API_TOKEN env var with a token to get more requests.
```

tests/fallback/fallback_test.go
mudler: I was about to switch to just run smoke/upgrades on arm64.. fallback at this point is very focused, but not sure. smoke at least give a good assessment and upgrades tests that upgrades are working at least

either way looks good to me

Itxaka:
> Failed getting the "github.com/hashicorp/qemu" plugin:
1 error occurred:
	* Plugin host rate limited the plugin getter. Try again in 14m32.154283s.
HINT: Set the PACKER_GITHUB_API_TOKEN env var with a token to get more requests.


Well, it makes sense becuase it installs the plugins from github. But I did not expect it to use the API for that...that is not nice.

tests/fallback/fallback_test.go
Itxaka: if you want I can drop that in this PR directly

Itxaka: I directly dropped the fallback tests for arm64

.github/workflows/build-pr-green-arm64.yaml
```diff
@@ -207,15 +208,15 @@ jobs:
     env:
       ARCH: arm64
       VAGRANT_CPU: 2
-      VAGRANT_MEMORY: 5120
+      VAGRANT_MEMORY: 6500
```
mudler, line 211: actually... this is a good catch. elemental shouldn't use /tmp by default during upgrades, or it would rely on enough ram to be availlable.. we used to default to COS_PERSISTENT here

mudler: MEMORY here should even be lower to be realistic. RPI and such don't have so much ram available

Itxaka: > VAGRANT_MEMORY here should even be lower to be realistic. RPI and such don't have so much ram available

Not really, this is CI specific. The defaults on the vagrant file should be lowered instead. I put this to 10Gb and 8 cpu on my local host to speed up, nothing to do with the default values. Plus this is for the test machine. Do we want to test like if we were running the tests against an RPI? Because I can already feel the pain just thinking about that lmao

mudler: no, but having an arm host pretending to have 8gb/10gb of ram is really unreal scenarios... imho we want a test that gives us some guarantees, if we start to pump it to make it work then there is no sense in running it then.

The fact that already made us aware of the elemental-cli used /tmp here is really helpful.. I wouldn't raise the ram here to fix the test

Itxaka: no no, the ram is not being raised to fix the test, it is to enable to run and get the full logs on where the hell is elemental thinking of storing the stuff, as locally it uses the proper state dir....so I need to see the full log of this properly to understand the issue as I cant reproduce it. Hopefully now with the full debug log added it will show, plus it can be triggered again as I lowered the mem to 4096

mudler: ah alright alright gotcha :+1: thought we were bumping ram to make it pass :sweat_smile: looks good then :+1:

Itxaka: its a bit weird cause it only happens on arm somehow....

Itxaka: interesting that it seems to be luet using the /tmp dir....
```
   INFO    Downloading image  quay.io/costoolkit/releases-green-arm64:cos-system-0.8.9-20
           INFO    Image: quay.io/costoolkit/releases-green-arm64:cos-system-0.8.9-20. Pulled: sha256:083d78ba6786108b31368703f64d8fdd7385febf584674f51f2258e967cfe1cd. Size: 840.3MiB
            ERROR     Failed compressing package quay.io/costoolkit/releases-green-arm64:cos-system-0.8.9-20: write /tmp/tmpluet/artifact596912627.zst: no space left on device
           INFO    Checking for file conflicts..
          panic: runtime error: invalid memory address or nil pointer dereference
          [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x96f4c8]
          
          goroutine 72 [running]:
          github.com/mudler/luet/pkg/api/core/types/artifact.(*Checksums).generateSum(0x400012dcc0, 0x0?, {{0x1120fa8?, 0x401070bc80?}, {0xecc3de?, 0x4000000000?}})
          	/luetbuild/go/pkg/mod/github.com/mudler/luet@v0.0.0-20220127164253-6f77fa2b3a72/pkg/api/core/types/artifact/checksum.go:78 +0x48
          github.com/mudler/luet/pkg/api/core/types/artifact.(*Checksums).generateSHA256(0x0?, 0x0?)
          	/luetbuild/go/pkg/mod/github.com/mudler/luet@v0.0.0-20220127164253-6f77fa2b3a72/pkg/api/core/types/artifact/checksum.go:73 +0x5c
          github.com/mudler/luet/pkg/api/core/types/artifact.(*Checksums).Generate(...)
          	/luetbuild/go/pkg/mod/github.com/mudler/luet@v0.0.0-20220127164253-6f77fa2b3a72/pkg/api/core/types/artifact/checksum.go:60
          github.com/mudler/luet/pkg/api/core/types/artifact.(*PackageArtifact).Verify(0x0)
          	/luetbuild/go/pkg/mod/github.com/mudler/luet@v0.0.0-20220127164253-6f77fa2b3a72/pkg/api/core/types/artifact/artifact.go:112 +0x70
          github.com/mudler/luet/pkg/installer.(*LuetInstaller).getPackage(0x400005c688?, {0x4028b6d7c0?, 0x40086b8bd0?, {0x111d130?, 0x401f740480?}}, {0x112a070?, 0x4028ae8c40?})
          	/luetbuild/go/pkg/mod/github.com/mudler/luet@v0.0.0-20220127164253-6f77fa2b3a72/pkg/installer/installer.go:947 +0x68
          github.com/mudler/luet/pkg/installer.(*LuetInstaller).downloadWorker(0x40000cf280, 0x96d000?, 0x40015ff2d8?, 0x0, 0x16?, {0x112a070, 0x4028ae8c40})
          	/luetbuild/go/pkg/mod/github.com/mudler/luet@v0.0.0-20220127164253-6f77fa2b3a72/pkg/installer/installer.go:981 +0xb4
          created by github.com/mudler/luet/pkg/installer.(*LuetInstaller).download
          	/luetbuild/go/pkg/mod/github.com/mudler/luet@v0.0.0-20220127164253-6f77fa2b3a72/pkg/installer/installer.go:638 +0x3dc
          : Process exited with status 2
```

Itxaka: I guess we should set the `LuetConfig.System.TmpDirBase` somewhere but not sure where.... luet default to `os.tmpdir()` any ideas @mudler ?

Itxaka:
Several issues here:
 - Seems installing the plugin to a fixed version makes packer hit the github api. Bad thing, its can hit the max requests and break the ci job.
 - Seems like arm64 consumes more memory than the x86 version, thus exhausting the VM mem, which in turn makes the /tmp folder run out of space (its tmpfs 25% of the physical memory)
 - Repo meta file is 300Mb by itself, which consumes a lot of memory - https://github.com/rancher-sandbox/cOS-toolkit/issues/1003
 - Seems like we dont ship the luet config with  a `tmpdir_base: "/var/tmp/luet"` config like with do with our get_luet.sh script, which means that it gets the default /tmp value, thus hitting this. Not sure if we want to ship a fix for this or not. Our upgrade command on elemental works around this by trying to get several different locations (like COS_PERSISTENT) for the tmpdir.

Knowing this, this PR is not necessary, and the fixes are too extensive for a non-tracked issue, so I will close this for the moment and open some cards for this.

PR #1297, opened 2022-05-06
Import docs from external repo

mudler:
This is part of https://github.com/rancher-sandbox/cOS-toolkit/issues/1286. I'd like to first merge it in and see the preview on github and then update links to point to new URLs.

There are just few minor adaptations (e.g. now packages docs are generated from the tree, no checkout needed)


.github/workflows/docs-publish.yml
```diff
@@ -0,0 +1,33 @@
...
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@master
+        with:
+          submodules: recursive
+      - name: Install deps
+        run: |
+              npm install postcss-cli
+              npm install autoprefixer
+              curl https://raw.githubusercontent.com/rancher-sandbox/cOS-toolkit/master/scripts/get_luet.sh | sudo bash -
```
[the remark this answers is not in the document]
mudler: fixed!

docs/Makefile
```diff
@@ -0,0 +1,21 @@
+export HUGO_VERSION?=0.85.0
```
Itxaka, line 1: cant we move this makefile into the /make folder so we have everything together?

docs/scripts/build.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+set -e
```
Itxaka, line 2: something for later, but there is probably a github action for this instead of having to do it via scripts. I dont think the scripts should be availabel tbh, there should eb no need for someone to publish the docs manually, ever.

Itxaka (approved):

docs/scripts/build.sh
Itxaka: errrh, this was meant for the publish.sh script, not the build ¬_¬

docs/Makefile
mudler: I'd really would go for optimizations on follow-ups as all of this can be handled much better (we can even have hugo inside the repo.. so versions are tight and we treat it like all the other CI deps)

mudler:
Not waiting for CI as it is unrelated. Will clean up makefiles in follow-ups, as I have already to update all links once its published

PR #1299, opened 2022-05-06
Small docs enhancements

mudler:
This PR adds a makefile.docs and links the docs building to the top level, it also changes CI to call that instead.

It also adds hugo in the toolchain. Once it's published I'd just use that in the CI so we are good with hugo versions used in the tree (and can cleanup more the docs Makefile)


make/Makefile.docs
```diff
@@ -0,0 +1,11 @@
+#
+# cOS-toolkit Makefile.docs
+#
+#
+
+
+docs-serve:
```
Itxaka, line 7: what is the problem for moving the whole makefile here instead of having this fake makefile that calls another makefile?

packages/toolchain/hugo/definition.yaml
```diff
@@ -0,0 +1,11 @@
+name: "hugo"
+category: "toolchain"
```
Itxaka, line 2: nice!

Itxaka (approved):

make/Makefile.docs
mudler: I like that it is having its own Makefile in its own folder until everything is not still "inglobated" (hugo, deps, etc), once we get hugo in the repos most of that is going to disappear and be just 2 scripts

PR #1303, opened 2022-05-09
Drop fallback tests for arm64

Itxaka:
Also enables more tests for master/release on arm64

### DIFF
--- a/.github/config/master.yaml
+++ b/.github/config/master.yaml
@@ -53,7 +53,7 @@ flavors:
         cache_repository: "build"
         organization: "quay.io/costoolkit"
         skip_tests: false
-        run_tests: ["test-fallback"]
+        run_tests: ["test-smoke", "test-upgrades-images-unsigned" ]
         flavor: "green"
         skip_tests_flavor: [ "blue","orange", "teal"]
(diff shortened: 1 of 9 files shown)


mudler (approved):

.github/config/master.yaml
```diff
@@ -53,7 +53,7 @@ flavors:
         cache_repository: "build"
         organization: "quay.io/costoolkit"
         skip_tests: false
-        run_tests: ["test-fallback"]
+        run_tests: ["test-smoke", "test-upgrades-images-signed", "test-upgrades-images-unsigned", "test-upgrades-local" ]
```
mudler, line 56: maybe we can just keep the unsigned tests for now? I guess we want to test if the upgrade scenario works, It should be enough x86_64 covering the signed scenarios, no?

mudler: yes definetly. at the times we didn't had elemental tests at all, so that was our only way of coverage. if elemental-cli have unsigned/signed tests we can cut here just with unsigned to have an integration test

Itxaka:
This already passed all tests, including arm64. Last commit is a rebase of everything, merging directly.

PR #1304, opened 2022-05-09
Add meta/cos-light

mudler:
Drop dependency of elemental-cli on cos-setup which brings the binary
also in unwanted situations. We have metapackages to cover desired
use-scenarios.

Also adds a new meta "cos-light" which ships just static config and a
selection of default cloud-config files that is meant to be overlayed over `/`. (Context, would be used in os2 to just consume framework configs when switching to SLE Micro for Rancher as base)

See: https://github.com/rancher-sandbox/rancher-node-image/pull/2
Part of: https://github.com/rancher-sandbox/os2/issues/94

### DIFF
--- a/docs/content/en/docs/_index.md
+++ b/docs/content/en/docs/_index.md
@@ -14,13 +14,9 @@ cOS is a toolkit which allows container images to be bootable in VMs, baremetals
 
 cOS allows to create meta-Linux derivatives which are configured throughout cloud-init configuration files and are immutable by default.
 
-cOS and derivatives shares a common feature set, can be upgraded in a **OTA-alike*** style, and upgrades are delivered with standard container registries. 
+cOS and derivatives shares a common feature set, can be upgraded with a A/B mechanism, and upgrades are delivered with standard container registries. 
 
 cOS comes also with vanilla images that can be used to boot directly container images built with the toolkit.
-
-{{% alert title="Note" %}}
-Note here we refer to "OTA-alike" merely as a method of distributing upgrades. It does not involve any specific setup (either wireless or cabled works) except an network connection to the registry where images are stored. Updates are delivered from one central location (the container registry) which hosts all images available for the clients to pull from.
-{{% /alert %}}
 
 ## Why cOS? 
 
@@ -38,7 +34,7 @@ To build a bootable image is as simple as running `docker build`.
 - Cloud-init driven
 - Based on systemd
 - Built and upgraded from containers - It is a [single image OS](https://quay.io/repository/costoolkit/releases-green)!
-- OTA updates
+- A/B updates
 - Easy to customize
 - Cryptographically verified
 - instant switch from different versions
--- a/docs/content/en/docs/Creating derivatives/creating_bootable_images.md
+++ b/docs/content/en/docs/Creating derivatives/creating_bootable_images.md
@@ -42,10 +42,56 @@ With the config file:
 In the example above, the cos-toolkit parts that are **required** are pulled in by `RUN luet install -y meta/cos-minimal`.
 Afterwards we install k9s and nerdctl packages to create our derivative with those packages on it.
 
+### Meta packages
+
+The toolkit is split into several meta-packages to pull only the individual needed part or a subset of them which are tied to a specific purpose, the meta packages are [documented in their own section](../../reference/packages/meta-packages).
+
 {{<package package="meta/cos-minimal" >}} is a meta-package that will pull {{<package package="toolchain/luet" >}}, {{<package package="toolchain/yip" >}}, {{<package package="utils/installer" >}}, {{<package package="system/cos-setup" >}}, {{<package package="system/immutable-rootfs" >}}, {{<package package="system/base-dracut-modules" >}}, {{<package package="system/grub2-config" >}}, {{<package package="system/cloud-config" >}}. 
 
+{{< tabpane >}}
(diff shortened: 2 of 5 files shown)


packages/cos-setup/definition.yaml
```diff
@@ -1,7 +1,3 @@
 name: cos-setup
 category: system
-version: 0.6.2-3
-requires:
-  - name: "elemental-cli"
-    category: "toolchain"
-    version: ">0.0.10"
+version: 0.6.2-3
```
Itxaka, line 3: wait, but this...this doesnt work right? cos-setup should require elemental otherwise its useless? you would get services failing? Is there a way of marking a soft dependency, i.e. luet advises that package X is needed for package Y to work? 

This seems like a recipe for disaster and bug reports :/

mudler: why someone should pull cos-setup individually at all? our front end user facing parts are the metas, and if they pick individual packages they have to get elemental-cli. technically speaking, cos-setup provides systemd units and glue, by itself it doesn't require elemental-cli at all

Itxaka: I dont know why would anyone pull it individually, but I know that the systemd files that are enabled by default with this package require elemental :/

Alright, alright, I guess we will see in the future :D

Itxaka (approved):

packages/cos-setup/definition.yaml
mudler: argh I know what you mean. to be completely fair I could do some way arounds it, but I wanted to keep it simple. if something more complex is needed we shall see

Itxaka: then no soft-requires in luet I guess? Or recommends?

Itxaka: nah, lets not complicate things now as its not needed, if issues arise in the future then we can probably work our way around that.

Itxaka: >  that would require a bit of work on the solver.

Would it really? It would just use the same stuff the solved does for requires but instead of adding them to the list to install, you can just add them to a list to report, report back to the logger and ven install them if there is a --install-recomends flag.

Of course Im talking out of my ass here as I havent looked at the solved in quite some time lol, always easy to talk about things when you dont remember :P

mudler: > > that would require a bit of work on the solver.
> 
> Would it really? It would just use the same stuff the solved does for requires but instead of adding them to the list to install, you can just add them to a list to report, report back to the logger and even install them if there is a --install-recomends flag.
> 
> Of course Im talking out of my ass here as I havent looked at the solved in quite some time lol, always easy to talk about things when you dont remember :P

oh that's a different thing then, I thought we were talking about reccomends in a way that you could install or not optionally from the CLI. that beside a list would need to take into account of all the potential conflicts, runtime deps, etc optionally when the solver computes the solution. (and then we should allow partial solutions... and then that's where all the big work would go)

from a notification point of view that is correct, should be also easy to exploit labels for that

mudler:
ISO looks good, getting issues using the ova image somehow.. can't seems to import it locally. trying now to debug the image with vbox to see if it fails here..

mudler:
mh. getting weird issue locally while trying to import the box image:

```
=> default: Importing base box 'cOS'...
Progress: 90%There was an error while executing `VBoxManage`, a CLI used by Vagrant
for controlling VirtualBox. The command and stderr is shown below.

Command: ["import", "/home/ettore/.vagrant.d/boxes/cOS/0/virtualbox/box.ovf", "--vsys", "0", "--vmname", "cOS_1652260257555_6699", "--vsys", "0", "--unit", "10", "--disk", "/home/ettore/VirtualBox VMs/cOS_1652260257555_6699/cOS-disk001.vmdk"]

Stderr: 0%...10%...20%...30%...40%...50%...60%...70%...80%...90%...100%
Interpreting /home/ettore/.vagrant.d/boxes/cOS/0/virtualbox/box.ovf...
OK.
0%...10%...20%...30%...
Progress state: NS_ERROR_INVALID_ARG
VBoxManage: error: Appliance import failed
VBoxManage: error: Code NS_ERROR_INVALID_ARG (0x80070057) - Invalid argument value (extended info not available)
VBoxManage: error: Context: "RTEXITCODE handleImportAppliance(HandlerArg*)" at line 1119 of file VBoxManageAppliance.cpp
```

docs/content/en/docs/_index.md
```diff
@@ -19,7 +19,7 @@ cOS and derivatives shares a common feature set, can be upgraded in a **OTA-alik
 cOS comes also with vanilla images that can be used to boot directly container images built with the toolkit.
 
 {{% alert title="Note" %}}
-Note here we refer to "OTA-alike" merely as a method of distributing upgrades. It does not involve any specific setup (either wireless or cabled works) except an network connection to the registry where images are stored. Updates are delivered from one central location (the container registry) which hosts all images available for the clients to pull from.
+Note here we refer to "OTA-alike" merely as a method of distributing upgrades. It does not involve any specific setup (either wireless or cabled works) except an network connection to the registry where images are stored. Updates are delivered from one central location (the container registry) which hosts all images available for the clients to pull from in A/B fashion.
```
Itxaka, line 22: should we just drop the OTA-alike directly and just use A/B system?

mudler: good point, done :+1:

PR #1312, opened 2022-05-11
Tag create-repo and publish-repo repositories 

Itxaka:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,9 +106,9 @@ ISO?=$(shell ls $(ROOT_DIR)/*.iso 2> /dev/null)
 
 # Used by publish-repo and create-repo for the snapshot-id
 # This is the commit of the latest tag
-LAST_TAGGED_COMMIT=$(shell git rev-list --tags --max-count=1)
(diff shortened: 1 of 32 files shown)


Makefile
```diff
@@ -104,6 +104,8 @@ PACKER_TARGET?=virtualbox-iso.cos
 #
 ISO?=$(shell ls $(ROOT_DIR)/*.iso 2> /dev/null)
 
+# Used by publish-repo and create-repo for the snapshot-id 
+GIT_COMMIT ?= $(shell git rev-parse HEAD)
```
mudler, line 108: could we also cover tags here? so we automatically get tagged snapshots

Itxaka: yes, Im wrapping my head on how to make this work for both normal commits and tagged snapshots...but make it work on the makefile directly instead of relaying on the CI magic...Im not sure how feasible this is :/

mudler: maybe something around the lines of https://github.com/c3os-io/c3os/blob/master/versions.sh#L3 ?

Itxaka: oh I got something in there to change between commit and tag :D

mudler (approved):

PR #1314, opened 2022-05-12
Publish a luet repo on release

Itxaka:
This generates a valid luet repo file with the proper reference for the
tag used on release, so you can download it directly into your
/etc/luet/repos.conf.d and start using it

Part of https://github.com/rancher-sandbox/cOS-toolkit/issues/1003

### DIFF
--- a/.github/build.yaml.gomplate
+++ b/.github/build.yaml.gomplate
@@ -621,6 +621,27 @@
 
 {{{ end }}}
 
+{{{define "generate_repo_files"}}}
+  {{{ $config := (datasource "config") }}}
+  {{{ $flavor := . }}}
+  generate-repo-files-{{{ $flavor }}}:
+    if: startsWith(github.ref, 'refs/tags/')
+    runs-on: ubuntu-latest
+    needs:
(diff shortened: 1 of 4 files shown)


Itxaka:
@mudler were you thinking something like this? Or just a plain text file with the repo reference (which IMHO is a bit extra, as the repository is gonna be tag-repository.yaml always, so its easy to pin yourself)

mudler:
> This generates a valid luet repo file with the proper reference for the tag used on release, so you can download it directly into your /etc/luet/repos.conf.d and start using it

luet repo add takes also an url, so should work too :slightly_smiling_face:

mudler (approved):
Looking great!

Itxaka:
> luet repo add takes also an url, so should work too slightly_smiling_face

ok, thats cool, didnt knew that :o

Itxaka:
I tested this as much as I could and it seems to work but wont be fully confirmed until we have a release , unfortunately. Its simple enough that it should not cause any issues..............................................................................famous last words as usual

PR #1315, opened 2022-05-12
Auto tag on cos/system bump

Itxaka:
This action will check after every master merge, check if the diff
contains changes to the cos package files, extract the current tag for
cos/system, get the latest published tag and compare both. If the
current cos/system version is different from the last tag, then it will
create a new tag with that version, pointing to the current commit.


Itxaka:
TESTING ONLY

.github/workflows/autorelease.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Check if same tag as previous
+        id: check
+        run: |
+          if [[ "${{steps.current.outputs.tag}}" == "${{steps.previous.outputs.tag}}" ]]; then
+            echo "::set-output name=bumped::true"
+          else
+            echo "::set-output name=bumped::false"
+          fi
+      - uses: rickstaa/action-create-tag@v1
+        if: ${{steps.check.outputs.nope}} == "no,impossible,dont tag yet"
```
mudler, line 41: this is awesome, I thought it would have needed some specific PAT token or w/e in order to accomplish this which always put me off doing this kinda of things. Is this working just with github secrets?

Itxaka: yup! and just tested a modified version of this on my own fork and it actually works :partying_face:

.github/workflows/autorelease.yaml
```diff
@@ -0,0 +1,44 @@
+name: autorelease
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+concurrency:
+  group: ci-autorelease-${{ github.repository }}
```
mudler, line 8: one thing I am a bit unsure is that if this runs aside with master both pipelines will try to publish, which could cause conflicts.. is there a way to trigger it after master actions are done?

Itxaka: mmmh, I dont understand this. This job only runs once something is merged to master, it cannot run in parallel with any other job, only with itself and that would cancel if a new one was triggered. Can you explain further what you mean?

Also, if this tries to crate a tag that already exists, it just fails, nothing else.

mudler: I mean, if this runs on master, and then it decides to tag, it will trigger then the release pipeline, which means master and the release pipeline could run at that point in parallel

Itxaka: is that a problem now with the tagged repositories? Or is this about pushing the artifacts tot he repo, that they could collide?

Itxaka: we could also run this every night and forget about it, that way it may even be better, max of one release per day, in case we merge several things that bump the cos version.

Itxaka: changed it to scheduled, also has a workflow dispatch so we  manually trigger if necessary

mudler: looks good!

mudler (approved):

mudler:
That's A-W-E-S-O-M-E!

.github/workflows/autorelease.yaml
```diff
@@ -0,0 +1,43 @@
+name: autorelease
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 2 * * *'
```
mudler, line 5: double-thinking, maybe it makes sense to move it to weekly. Afterall we have already nightlies, autobumps included which will run and probably choke entirely the CI. instead if we tag on weekends we get the start of week with a fresh new tag of latest

Itxaka: good to me

Itxaka: moved it to run at 2am each saturday

PR #1318, opened 2022-05-13
Stop indexing all packages on publish

Itxaka:
Rely into the snapshot repo in order to not publish the whole repo


mudler:
> So this is good because we will publish only the snapshot, but wont this mean that the plain repository.yaml will now get outdated?
No no, this will also publish the `repository.yaml`, so latest will always get on pair. What we do here is just avoid to repopulate the new index with all the metadata coming from the old one.

You would use references only to pinpoint, otherwise it goes straight to latest

mudler (approved):

mudler:
No need to wait CI imho. we will notice this immediately as after a publish everything should be speeded up as index would be so much less in size

Issue #1319, opened 2022-05-13
Autoreleaser should ignore build versions


Itxaka:
**cos-toolkit version:**

**CPU architecture, OS, and Version:**

**Describe the bug**

Currently the bumber bumps cos by appending dash and a number, but our releases are all semver versions.

**To Reproduce**

**Expected behavior**

Not sure. I mean we should probably bump only on semver fully matched, so ignoring the dash seems the best thing to do.

Another issue is that I expected the tag to trigger the release pipeline, but it doesn't sue to the tag being created by the GitHub actions. Should probably move that to use the bot credentials instead.

**Logs**

**Additional context**

PR #1320, opened 2022-05-13
Fix tag and tag pushing for autorelease

Itxaka:
Ignore any build number (i.e. -NUMBER) that the bot auto pushes when
comparing version as we only want the semver to be pushed.

Also push the tag ourselves with the rest api so we can use the CiBot
PAT, which means that release jobs will be triggered on tag release.
Otherwise, because the tag cames from the github actions itself, the
jobs wont trigger. This is a deliberate decision from github in order to
avoid circular jobs.

Fixes: #1319

### DIFF
--- a/.github/workflows/autorelease.yaml
+++ b/.github/workflows/autorelease.yaml
@@ -17,7 +17,7 @@ jobs:
       - name: Get current tag
         id: current
         run: |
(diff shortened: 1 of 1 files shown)


mudler (approved):

Itxaka:
No need to wait, I guess

PR #1307, opened 2022-05-09
Testing new elemental

davidcassany:
Signed-off-by: David Cassany <dcassany@suse.com>

PR #1321, opened 2022-05-13
Change green for teal as the default base

Itxaka:
Can you imagine if this worked in the first run?

Changes:
 - Green is now teal, teal changed to green so all tests and pipelines and such run on teal instead
 - On local-build it adds to the manifest.yaml the final repo with a lower priority than the build dir but a higher priority than the default elemental repo (green) so any needed packages are picked up from the proper flavor
 - change `GreenRepo` to `ArtifactsRepo` in the tests so its clear what it is.
 - change default test repo to teal flavor

Not any other changes than that. Seems as there is too many but its mostly autogenerated workflows


Itxaka:
meh, isos still get the needed packages from green. Not sure if we should add the repo here or what. If we add the repo, then we end up with the old situation of getting the packages from the remote instead of local :/

And changing the default repo in elemental?? Maybe we should be able to add repo+priority directly with the cmd

Itxaka:
mtree fails to validate the oem dir... not sure why. 

`"oem": unexpected path`

Itxaka:
uh, somehow the `oem/` dir is being passed as `oem` instead, so it fails to match the excludes...

```
[2022-05-18 08:50:01] Checking /usr/local/elemental-upgrade against validation file /tmp/luet-mtree-metadata-mtree3168654540/cos-system-0.8.9-18.mtree
[2022-05-18 08:50:01] Using the following exclude list: [var/cache/luet oem/ usr/local/ lost+found tmp/ mnt/]
[2022-05-18 08:50:03] Checking lost+found against exclude list
[2022-05-18 08:50:03] Checking item var/cache/luet against value lost+found
[2022-05-18 08:50:03] Checking item oem/ against value lost+found
[2022-05-18 08:50:03] Checking item usr/local/ against value lost+found
[2022-05-18 08:50:03] Checking item lost+found against value lost+found
[2022-05-18 08:50:03] Path lost+found found against exclude values [var/cache/luet oem/ usr/local/ lost+found tmp/ mnt/], skipping entry.
[2022-05-18 08:50:03] Checking oem against exclude list
[2022-05-18 08:50:03] Checking item var/cache/luet against value oem
[2022-05-18 08:50:03] Checking item oem/ against value oem
[2022-05-18 08:50:03] Checking item usr/local/ against value oem
[2022-05-18 08:50:03] Checking item lost+found against value oem
[2022-05-18 08:50:03] Checking item tmp/ against value oem
[2022-05-18 08:50:03] Checking item mnt/ against value oem
[2022-05-18 08:50:03] Dindnt match oem
[2022-05-18 08:50:03] Item oem has diff type extra
[2022-05-18 08:50:03] Validation failed :(
[2022-05-18 08:50:03] validation failed
```

Itxaka:
ok now, my head is exploding, if I pull the artifact directly it has no oem dir:

```
cos:~ # elemental pull-image quay.io/costoolkit/releases-teal:cos-system-0.8.9-18 /tmp/teal
INFO[2022-05-18T09:10:31Z] Starting elemental version v0.0.14           
INFO[2022-05-18T09:10:31Z] Unpacking docker image: quay.io/costoolkit/releases-teal:cos-system-0.8.9-18 
INFO[2022-05-18T09:10:42Z] Pulled: sha256:6b14b257a57efb1d723780ca1692e28c8d514271a5cef85e5ab6622f910d583d quay.io/costoolkit/releases-teal:cos-system-0.8.9-18 
INFO[2022-05-18T09:10:42Z] Size: 739.4MiB                               
cos:~ # ls /tmp/teal/
bin  boot  dev	etc  home  lib	lib64  mnt  opt  proc  root  run  sbin	srv  sys  system  tmp  usr  var
```

but if I check under the extracted upgrade path, there is an oem dir:

```
cos:~ # ls /usr/local/elemental-upgrade/
bin   dev  home  lib64	     mnt  opt	root  sbin  sys     tmp  var
boot  etc  lib	 lost+found  oem  proc	run   srv   system  usr
```

so its the upgrade creating that dir somehow, thus throwing the mtree into chaos...

Itxaka:
https://github.com/rancher-sandbox/elemental/blob/main/pkg/utils/common.go#L112 is what creates the empty oem dir. Not sure why the hell this doesn't fail on green?

Itxaka:
iso building from green repos was fixed on latest elemental
mtree failure to skip oem dir was fixed on latest mtree

This should hopefully now pass the tests.

Itxaka:
Test passing, fixed the example docker for standard, ready to review @mjura @mudler @davidcassany

davidcassany:
@Itxaka sorry just saw merging #1307 caused little conflicts here...

packages/cos-config/cos-config
```diff
@@ -27,7 +27,7 @@
 
 # This is the repository that hosts the signature files used by cosign and luet-cosign plugin during upgrade/deploy to
 # check the artifact signatures
-#COSIGN_REPOSITORY="raccos/releases-green"
+#COSIGN_REPOSITORY="raccos/releases-teal"
```
davidcassany, line 30: Note after merging #1307 this is no longer used. We should consider changing elemental defaults to use teal.

davidcassany (approved):
Looks good! :+1:

mudler (approved):

mudler:
Looking good here!

packages/cos-config/cos-config
Itxaka: wait, we pick the repo now from green directly? liek a default? argh, I migth have not see that in your PR.

In any case, quay.io now supports cosing signatures, so maybe it would be good to just drop this completely from everywhere....

Itxaka:
So I fixed arm64 tests the other day....

south-park-its-gone


Looking at those for now...blergh

Itxaka:
on arm64, after packer creates the box it results in not finding the /`boot/vmlinuz` file, so the box cant boot. Weird when the iso boots just fine???

davidcassany:
> So I fixed arm64 tests the other day....

Were they gone before I merged #1307 today morning... ? 😕

davidcassany:
> on arm64, after packer creates the box it results in not finding the /`boot/vmlinuz` file, so the box cant boot. Weird when the iso boots just fine???

mmm this ISO load the kernel and initrd from the ISO filesystem itself, so to create the iso kernel and initrd are extracted from OS image and copied to the ISO filesystem, so if the ISO boots there must be a Kernel somewhere. I bet the problem is with a missing symlink. IIRC there was something related to that on OS2.

Itxaka:
> I probably we are missing something like that.

Good point, although I kind of remember already having that in the grub code when we first created the arm64 flavors.... :eyes:

Itxaka:
wird, there is no vmlinuz link in the iso either...how does it boot then? The grub params are very clear, it sets the kernel to `/boot/vmlinuz`

davidcassany:
> wird, there is no vmlinuz link in the iso either...how does it boot then? The grub params are very clear, it sets the kernel to `/boot/vmlinuz`

ISOs work different, they use a different fixed kernel and initrd names. This is happening here https://github.com/rancher-sandbox/elemental/blob/de33226faf1a051f3e46b9a9d70401c61fd5c144/pkg/action/build-iso.go#L131-L142

Build-iso finds kernel and initrd names based on heuristic check from here https://github.com/rancher-sandbox/elemental/blob/de33226faf1a051f3e46b9a9d70401c61fd5c144/pkg/elemental/elemental.go#L524-L538

Itxaka:
ok, seems the kernel package is wrong for teal a it does not link the vmlinuz due to an if-else, should be an easy fix if I was able to understand all the ifs in there LOL

Itxaka:
Not gonna wait for amr64 as other issues have arisen due to the new elemental config, so as soon as teal passes the tests, will merge this

Itxaka:
Umm, grub assesment is failing on teal during the recovery upgrade, probably we are not mounting the STATE dir during recovery upgrade?

also pretty terrible that this doesnt fail? I mena, I get it that its not a hard failure but...not sure if this should fail, currently we only have some config in there but down the line we could have much more

```
ERRO[2022-05-19T16:12:38Z] grub2-editenv: error: cannot open `/tmp/mnt/STATE/boot_assessment': Read-only file system.
: failed to run grub2-editenv /tmp/mnt/STATE/boot_assessment set enable_boot_assessment=yes
: exit status 1 
ERRO[2022-05-19T16:12:38Z] 1 error occurred:
	* failed to run grub2-editenv /tmp/mnt/STATE/boot_assessment set enable_boot_assessment=yes
: exit status 1
 
ERRO[2022-05-19T16:12:38Z] open /tmp/mnt/STATE/grub_boot_assessment: read-only file system 
ERRO[2022-05-19T16:12:38Z] 1 error occurred:
	* open /tmp/mnt/STATE/grub_boot_assessment: read-only file system
 
```

Itxaka:
This last issue seems to come from the boot assesment package. Even running the cloud-init manually results in a failure:

```
INFO[2022-05-19T16:24:49Z] Command output: mount: /tmp/mnt/STATE: WARNING: source write-protected, mounted read-only. 
```

will sync tomorrow with @mudler about this, as long as COS_STATEis mounted on the system in RO, you cannot mount it anywhere RW, you need to remount it RW

davidcassany:
> probably we are not mounting the STATE dir during recovery upgrade?

Exactly think this is the case

> also pretty terrible that this doesnt fail? I mena, I get it that its not a hard failure but...not sure if this should fail, currently we only have some config in there but down the line we could have much more

The real question is where this boot assessment is happening. In fact I was surprised the boot assessment actually worked, probable it simply didn't 😓 I even ahd had a patch for that, but I saw green tests on the CI and I did not want to touch it any further. I think the problem is here:

https://github.com/rancher-sandbox/cOS-toolkit/blob/4c7e555321e3412221cc8df0ad09a494b49bbf0e/packages/cloud-config/oem/08_boot_assessment.yaml#L43-L47

Under some circumstances state is already mounted as read only, so these commands fail.

mudler:
> This last issue seems to come from the boot assesment package. Even running the cloud-init manually results in a failure:
> 
> ```
> INFO[2022-05-19T16:24:49Z] Command output: mount: /tmp/mnt/STATE: WARNING: source write-protected, mounted read-only. 
> ```
> 
> will sync tomorrow with @mudler about this, as long as COS_STATEis mounted on the system in RO, you cannot mount it anywhere RW, you need to remount it RW
> 
> This may work on booting from COS_ACTIVE/PASSIVE as we remount the state partition RW but when upgrading recovery, we mount COS_RECOVERY instead as thats the one we are modifying

boot assessment does make sense only on COS_ACTIVE/COS_PASSIVE indeed, the COS_RECOVERY is not meant to flip the switches during upgrades of it - however - since there is no "recovery upgrade" hook, but there is just "upgrade" hook it was run anyway (without having any effect, before).

I'm not sure if this is somehow tied to the elemental-cli bump which changed the hooks order, that might have had some impact, but nevertheless shouldn't make the test fail, as if no sentinel is set, boot assessment won't just trigger

Issue #1327, opened 2022-05-17
Try ARM self-hosted runner


mudler:
Try to start up an ARM GH runner on the new host with our runner docker image https://github.com/rancher-sandbox/cOS-toolkit/tree/master/.github/docker/ghrunner 

ping @pgonin

mudler:
Job running on the new runner: https://github.com/rancher-sandbox/cOS-toolkit/runs/6472529374?check_suite_focus=true . Will open up a PR for docs on how to start it up

PR #1331, opened 2022-05-17
Use hugo from toolchain to generate docs

mudler:


mudler:
(using the ARM label to test #1327 )

PR #1338, opened 2022-05-19
Disable mtree verify by default

mjura:
We would like to disable mtree plugin by default,
which seems to not be used for now.

Fixes #183

Signed-off-by: Michal Jura <mjura@suse.com>

This PR depends on https://github.com/rancher-sandbox/elemental/pull/208

### DIFF
--- a/tests/fallback/fallback_test.go
+++ b/tests/fallback/fallback_test.go
@@ -46,7 +46,7 @@ var _ = Describe("cOS booting fallback tests", func() {
 			// Auto assessment was installed
 			bootAssessmentInstalled()
 
-			out, err := s.Command(fmt.Sprintf("elemental upgrade --no-verify --system.uri docker:%s:cos-system-%s", s.ArtifactsRepo, s.TestVersion))
+			out, err := s.Command(fmt.Sprintf("elemental upgrade --system.uri docker:%s:cos-system-%s", s.ArtifactsRepo, s.TestVersion))
 			Expect(err).ToNot(HaveOccurred(), out)
 			Expect(out).Should(ContainSubstring("Upgrade completed"))
 
--- a/tests/upgrades-images-unsigned/upgrade_test.go
+++ b/tests/upgrades-images-unsigned/upgrade_test.go
@@ -23,7 +23,7 @@ var _ = Describe("cOS Upgrade tests - Images unsigned", func() {
 	})
 	Context("After install", func() {
 		When("images are not signed", func() {
-			It("upgrades with --no-verify", func() {
+			It("upgrades", func() {
 
 				if s.GetArch() == "aarch64" {
 					By("Upgrading aarch64 system")
@@ -38,7 +38,7 @@ var _ = Describe("cOS Upgrade tests - Images unsigned", func() {
 				Expect(out).ToNot(Equal(""))
 
 				version := out
-				out, err = s.Command(fmt.Sprintf("elemental upgrade --no-verify --system.uri docker:%s:cos-system-%s", s.ArtifactsRepo, s.TestVersion))
+				out, err = s.Command(fmt.Sprintf("elemental upgrade --system.uri docker:%s:cos-system-%s", s.ArtifactsRepo, s.TestVersion))
 				Expect(err).ToNot(HaveOccurred(), out)
 				Expect(out).Should(ContainSubstring("Upgrade completed"))
 				By("rebooting")
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ isowork/
 *.iso.sha256
 iso-meta.json
 iso-meta.yaml
+.idea/
 .qemu/
 .vagrant/
 ^Vagrantfile
--- a/docs/content/en/docs/Creating derivatives/creating_bootable_images.md
+++ b/docs/content/en/docs/Creating derivatives/creating_bootable_images.md
@@ -133,7 +133,7 @@ The workflow would be then:
 
 1) `docker build` the image
 2) `docker push` the image to some registry
-3) `elemental upgrade --no-verify --docker-image $IMAGE` from a cOS machine or (`elemental reset` if bootstrapping a cloud image)
+3) `elemental upgrade --docker-image $IMAGE` from a cOS machine or (`elemental reset` if bootstrapping a cloud image)
 
 The following can be incorporated in any standard gitops workflow.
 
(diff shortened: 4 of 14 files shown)


packages/backports/installer/cos.sh
```diff
@@ -266,10 +266,10 @@ usage()
     echo ""
     echo "Example: $PROG-install /dev/vda"
     echo "  install:"
-    echo "  [--partition-layout /path/to/config/file.yaml ] [--force-efi] [--force-gpt] [--docker-image IMAGE] [--no-verify] [--no-cosign] [--iso https://.../OS.iso] [--debug] [--tty TTY] [--poweroff] [--no-format] [--config https://.../config.yaml] DEVICE"
+    echo "  [--partition-layout /path/to/config/file.yaml ] [--force-efi] [--force-gpt] [--docker-image IMAGE] [--verify] [--no-cosign] [--iso https://.../OS.iso] [--debug] [--tty TTY] [--poweroff] [--no-format] [--config https://.../config.yaml] DEVICE"
     echo ""
     echo "  upgrade:"
-    echo "  [--strict] [--recovery] [--no-verify] [--no-cosign] [--directory] [--docker-image] (IMAGE/DIRECTORY)"
```
mudler, line 272: this is the previous installer backport, I'd leave it as it is, as it serves just for legacy support

mudler:
In order to trigger tests against the elemental-cli version you have in the branch, you need to bump the elemental-cli package version here https://github.com/rancher-sandbox/cOS-toolkit/blob/master/packages/toolchain/elemental-cli/collection.yaml#L14 with it's associated "git.hash" pointing to your commit/branch, also a bump of the cOS collection is needed too, otherwise tests would not go over the new version

packages/backports/installer/cos.sh
mjura: ok, let's leave it

mjura:
Ok, I have use git.hash from elemental, let's see if this will work
https://github.com/rancher-sandbox/cOS-toolkit/pull/1338/files#diff-2c30c62eb9ddaad10b7772d0502c65c0f514bb028fea72a626a99a664e23c826R14

tests/upgrades-images-unsigned/upgrade_test.go
```diff
@@ -38,7 +38,7 @@ var _ = Describe("cOS Upgrade tests - Images unsigned", func() {
 				Expect(out).ToNot(Equal(""))
 
 				version := out
-				out, err = s.Command(fmt.Sprintf("elemental upgrade --no-verify --system.uri docker:%s:cos-system-%s", s.ArtifactsRepo, s.TestVersion))
+				out, err = s.Command(fmt.Sprintf("elemental upgrade --verify --system.uri docker:%s:cos-system-%s", s.ArtifactsRepo, s.TestVersion))
```
davidcassany, line 41: Isn't this a behavior change? Prior the tests were testing without mtree verification, with this change they will start using verification, so if images are not signed it will fail.

tests/fallback/fallback_test.go
```diff
@@ -46,7 +46,7 @@ var _ = Describe("cOS booting fallback tests", func() {
 			// Auto assessment was installed
 			bootAssessmentInstalled()
 
-			out, err := s.Command(fmt.Sprintf("elemental upgrade --no-verify --system.uri docker:%s:cos-system-%s", s.ArtifactsRepo, s.TestVersion))
+			out, err := s.Command(fmt.Sprintf("elemental upgrade --verify --system.uri docker:%s:cos-system-%s", s.ArtifactsRepo, s.TestVersion))
```
davidcassany, line 49: Isn't this a behavior change?

davidcassany (changes requested):
In addition to what Ettore already said regarding the backports (they should remain untouched) I think the tests are having a behavior change. If this change is intended then the titles and descriptions of the tests should be updated.

tests/upgrades-images-unsigned/upgrade_test.go
mjura: argh, you are right, thank you

tests/fallback/fallback_test.go
mjura: you are right, thank you

mjura:
@mudler @davidcassany I fixed your comments this change is ready to go

mudler:
To properly run tests against, a bump of the cOS collection is needed (otherwise you will run tests against the old image, which contains the old elemental-cli)

mudler:
I'm a bit puzzled now, don't we have anymore a test which leverages `--verify` now anymore? I think previous tests which were using it now should enable `--verify` explicitly, no?

mudler:
It needs also cosign enabled too, which is unrelated to this change. I think we lost it as well somehow when switching it off by default.. ?

mjura:
I have updated `tests/upgrades-images-signed/upgrade_test.go` and `packages/cos/collection.yaml` let's see what CI is going to say?

mudler (approved):
Looking good!

packages/toolchain/elemental-cli/collection.yaml
```diff
@@ -11,7 +11,7 @@ packages:
       autobump.revdeps: "true"
       autobump.strategy: "git_hash"
       autobump.git.branch: "main"
-      git.hash: "4ca13d5eff80257dbe9b31e91ac24352b9118b5f"
+      git.hash: "2a84274944eb74c01bd6a9a86a3c4712aaedc1a2"
```
mudler, line 14:
```suggestion
      git.hash: "4f8bf31dff1e5b539a6654810317c851faf70732"
```

mudler:
It seems to fail tests as the version of the cli wasn't bumped, maybe can just cherry-pick https://github.com/rancher-sandbox/cOS-toolkit/pull/1335 ?

packages/toolchain/elemental-cli/collection.yaml
mjura: Thank you @mudler I was wondering why it is failing, rebased

mjura:
Hopefully now, tests will be passing properly

davidcassany (approved):
Looks good :+1:

Itxaka (approved):

Itxaka:
Some CI jobs failed due to deps not being installed, all github network issues. Should not matter much as long as one of the tests suite passes...

Issue #1339, opened 2022-05-19
Release pipeline fails during repository-file upload


mudler:
**Describe the bug**
Seems the Release pipeline is failing uploading the repository file

**Additional context**
https://github.com/rancher-sandbox/cOS-toolkit/runs/6500514840?check_suite_focus=true

Itxaka:
I know exactly why this is

PR #1340, opened 2022-05-19
Fix yq stanza for generating repo files

Itxaka:
Fixes https://github.com/rancher-sandbox/cOS-toolkit/issues/1339


mudler (approved):
good catch!

Itxaka:
not wiating for tests, it makes no sense, release pipeline

PR #1345, opened 2022-05-20
Fix raw_disk building

Itxaka:
Set the proper values on the manifest so it picks up the repos.
Also set the FINAL_REPO for the ARCH on raw_image build with a priority
of 30 so its below the local build repository


Itxaka:
This wont make it work by itself, we require a new elemental with https://github.com/rancher-sandbox/elemental/pull/214#pullrequestreview-979751144

Itxaka:
This fixes the raw image for all flavors. They are now build with the proper packages from the flavor

mudler (approved):

PR #1348, opened 2022-05-20
Move from leap to tumbleweed

Itxaka:
Fixes (When it works): https://github.com/rancher-sandbox/cOS-toolkit/issues/1257

### DIFF
--- a/packages/firmware/collection.yaml
+++ b/packages/firmware/collection.yaml
@@ -10,12 +10,12 @@ packages:
       autobump.string_replace: '{ "prefix": "" }'
       autobump.prefix: "prefix"
       autobump.hook: |
-        curl -s -L https://download.opensuse.org/ports/aarch64/tumbleweed/repo/oss/$(curl -s -L https://download.opensuse.org/ports/aarch64/tumbleweed/repo/oss/repodata/repomd.xml | dasel select -p xml 'repomd.data.[0].location.-href') | gunzip | dasel -r xml -p json | jq '.metadata.package[] | select(.name=="u-boot-rpiarm64") | select(.arch!="src").version | map(.)[2] + "-" +  map(.)[1]' -r | tail -n1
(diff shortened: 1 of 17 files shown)


packages/firmware/collection.yaml
```diff
@@ -10,12 +10,12 @@ packages:
       autobump.string_replace: '{ "prefix": "" }'
       autobump.prefix: "prefix"
       autobump.hook: |
-        curl -s -L https://download.opensuse.org/distribution/leap/15.3/repo/oss/$(curl -s -L https://download.opensuse.org/distribution/leap/15.3/repo/oss/repodata/repomd.xml | dasel select -p xml 'repomd.data.[0].location.-href') | gunzip | dasel -r xml -p json | jq '.metadata.package[] | select(.name=="u-boot-rpiarm64") | select(.arch!="src").version | map(.)[2] + "-" +  map(.)[1]' -r | tail -n1
```
mudler, line 13: nope, not for firmwares. we have been already there, and seems Leap ones get more attention, TW got broken for RPIs several times

Itxaka: really? How weird, I would not expect that! Reverting this!

values/teal-x86_64.yaml
```diff
@@ -4,7 +4,7 @@ codename: "teal"
 arch: "x86_64"
 golang_arch: "amd64"
 
-tool_image: opensuse/leap:15.3
```
mudler, line 7: I'd be tempted to stay stable on teal, and avoid tw or latest tags. We can use the other flavors to assess issues with newer images

mudler:
Needs a bump to the cos collection for a good test :clown_face:

mudler (approved):
Looks good!

Itxaka:
well, it fails incredibly...

Itxaka:
```
#8 [4/5] RUN zypper in -y --no-recommends grub2-i386-pc grub2-x86_64-efi grub2 curl tar xz git
       #8 sha256:bbbb63be7aaebdea5c8499111de8918f42d25cb2a35afb7be7e9725b7162da39
 INFO   #8 0.409 terminate called after throwing an instance of 'std::system_error'
       #8 0.409   what():  Operation not permitted
```

works locally :thinking:

packages/firmware/collection.yaml
Itxaka: done!

values/teal-x86_64.yaml
Itxaka: done!

Issue #1323, opened 2022-05-14
fs integrity check on boot


mudler:
**Is your feature request related to a problem? Please describe.**
In a deployment of mine it happened that COS_PERSISTENT was corrupted and prevented the system to boot. As I could not access to the machine except ssh from rescue mode (from host provider) and control panel it makes it really hard to debug

**Describe the solution you'd like**
A default cloud-config setting that checks fs in early stages and attempt to fix errors. ideally should output logs to state partition so they can be viewed afterwards, so it can just enable ideally systemd-fsck (although might not fit our partitioning logics https://github.com/systemd/systemd/blob/main/src/fsck/fsck.c#L298) .  We need a mechanism to propagate back this to derivatives from a common place (like, we do have already meta/cos-light for an opinionated default)

**Describe alternatives you've considered**
Let derivatives handle some sane default setting by themselves, and just adding appropriate docs for it

**Additional context**

mudler:
Testing locally by trying to re-use systemd-fsck

mudler:
I've triggered an ext corruption of COS_PERSISTENT with: 

```
dd if=/dev/zero count=1 bs=4096 seek=0 of=/dev/XXXX
```

The system stalls during initramfs booting: 
![Screenshot from 2022-05-18 11-59-12](https://user-images.githubusercontent.com/2420543/169013398-2b09d818-5736-4bd8-9102-2369674dfb8d.png)

mudler:
I couldn't work on this card today, so I'm leaving my notes from yesterday:

- I'm trying to change `cos-loop-img.sh` and `cos-mount-layout.sh` by calling systemd-fsck appropriately, but currently my generated initramfs fail booting even LiveCD.

PR #1352, opened 2022-05-20
Add fsck to dracut modules

mudler:
Also sets up fsck.repair=yes and fsck.mode=force by default.
Derivatives can opt-out by providing their own bootargs.cfg

Fixes: #1323 


packages/immutable-rootfs/30cos-immutable-rootfs/module-setup.sh
```diff
@@ -29,7 +29,7 @@ install() {
     # Include utilities required for cos-setup services,
     # probably a devoted cos-setup dracut module makes sense
     inst_multiple -o \
-        partprobe sync udevadm lsblk sgdisk parted mkfs.ext2 mkfs.ext3 mkfs.ext4 mkfs.vfat mkfs.fat mkfs.xfs blkid e2fsck resize2fs mount xfs_growfs umount
+        /usr/lib/systemd/systemd-fsck partprobe sync udevadm lsblk sgdisk parted mkfs.ext2 mkfs.ext3 mkfs.ext4 mkfs.vfat mkfs.fat mkfs.xfs blkid e2fsck resize2fs mount xfs_growfs umount
```
mudler, line 32: this *should* be already included - BUT - I feel more safe being explicit here. I'm not sure if there is any reason to NOT do it.

mudler:
mmm seems the tpm.mount in tumbleweed is in /usr/lib/systemd/system/tmp.mount as opposed to leap where is in /usr/share/systemd/tmp.mount

packages/immutable-rootfs/30cos-immutable-rootfs/cos-loop-img.sh
```diff
@@ -9,16 +9,38 @@ function doLoopMount {
...
+
             exit 0
         else
             umount "${cos_state}"
         fi
     done
 }
 
+function dofsCheck {
+    # Iterate over current partitions
```
mudler, line 36: @davidcassany here while testing I corrupted COS_PERSISTENT in a way that no labels were properly displayed, but by running efsck manually I could restore them just fine. 

Since this could happen, in the current initramfs I could not hook into alread-existing loop - so I'm back at scanning all the partitions found. I'm not sure if this is maybe too invasive on some front and maybe better to make it opt-in - but on the other hand I don't see why someone would disable a fscheck integrity utility running on boot. I guess, in that case `fsck.mode=skip` could be passed, so I didn't felt to add any option to instrument it even more.

mudler:
who did forgot to bump initrd? me!

mudler:
ok, tested locally and seems to work just fine, I've enabled it now by default on the default `bootargs.cfg` that we ship. Will try now to add a test for it :crossed_fingers:

mudler:
oh my, tests went green

Itxaka (approved):

packages/immutable-rootfs/30cos-immutable-rootfs/module-setup.sh
davidcassany: I don't think there is any reason not to do it. However this is part of the `00systemd` module here https://github.com/dracutdevs/dracut/blob/1befc6416743a527824f0f2cc25471e86a6f8a79/modules.d/00systemd/module-setup.sh#L33-L41 so from a dracut perspective I guess the proper way of enforcing it would be to require `systemd` module in `depends` function. I doubt we explicitly require systemd module anywhere, assuming it is already there, which could not be the case.

So, I'd suggest adding `systemd` in depends, then keep this `systemd-fsck`inclusion here is good to explicitly tell we make use of systemd-fsck. I am wondering if shouldn't we use the `${systemdutildir}` variable instead of the absolute path, just incase there are divergences on distros or the installation path of systemd evolves.

davidcassany:
> mmm seems the tpm.mount in tumbleweed is in /usr/lib/systemd/system/tmp.mount as opposed to leap where is in /usr/share/systemd/tmp.mount

Yes this is just because Leap has not done the switch to tmpfs mount in `/tmp` by default yet. This is something we should check for teal now that I am thinking about it, whast is SLE for Rancher using for `/tmp` by default?

packages/base-dracut-modules/50-cos-initrd.conf
```diff
@@ -2,4 +2,4 @@ hostonly_cmdline="no"
 hostonly="no"
 compress="xz"
 omit_dracutmodules+=" multipath "
-add_dracutmodules+=" livenet dmsquash-live "
+add_dracutmodules+=" fs-lib livenet dmsquash-live "
```
davidcassany, line 5: mmm I am not so sure about it. If this is a dependency of `systemd-fsck` I doubt it should be here. Instead better add it in `depends()` in `packages/immutable-rootfs/30cos-immutable-rootfs/module-setup.sh`, so each installation of the immutable-rootfs module will also pull and include this dracut module.

packages/immutable-rootfs/30cos-immutable-rootfs/cos-loop-img.sh
```diff
@@ -9,16 +9,38 @@ function doLoopMount {
...
         [ -e "/tmp/cosloop-${label}" ] && continue
         > "/tmp/cosloop-${label}" 
+
         mount -t auto -o "${cos_root_perm}" "/dev/disk/by-label/${label}" "${cos_state}" || continue
         if [ -f "${cos_state}/${cos_img}" ]; then
-            losetup -f "${cos_state}/${cos_img}"
+
+            # FSCHECK if cos_root_perm == "ro" on both
+            if [ "$cos_root_perm" == "ro" ]; then
+               /usr/lib/systemd/systemd-fsck "/dev/disk/by-label/${label}"
```
davidcassany, line 18: This absolute path frightens me... I am not sure if we have the `${systemdutilsdir}` variable here, I doubt it though. Probably we could install a symlink under `/sbin` in module_setup.sh and just use `systemd-fsck` here?

packages/immutable-rootfs/30cos-immutable-rootfs/module-setup.sh
```diff
@@ -5,9 +5,9 @@ check() {
     return 255
 }
 
-# called by dracut
+# called by dracut 
 depends() {
-    echo rootfs-block dm
+    echo rootfs-block dm fs-lib
```
davidcassany, line 10: If fs-lib is here, then it should not be needed in `dracut.conf` file.

packages/immutable-rootfs/build.yaml
```diff
@@ -7,7 +7,13 @@ steps:
 {{ if .Values.distribution }}
 {{if eq .Values.distribution "opensuse" }}
 # Mount /tmp as tmpfs by default as set by systemd itself
-- cp /usr/share/systemd/tmp.mount /etc/systemd/system
+- | 
+    /bin/bash -c " \
+    if [ -e /usr/share/systemd/tmp.mount ]; then \
+      cp /usr/share/systemd/tmp.mount /etc/systemd/system; \
```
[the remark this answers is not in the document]
davidcassany: could be, note that TW alreay has this mountpoint by default, probably SLE for rancher too. The old logic was valid when opensuse meant leap, if we include TW, this logic is safer.

packages/grub2/config/bootargs.cfg.tmpl
```diff
@@ -2,7 +2,7 @@ set kernel=/boot/vmlinuz
 if [ -n "$recoverylabel" ]; then
     set kernelcmd="console=tty1 console=ttyS0 root=live:LABEL=$recoverylabel rd.live.dir=/ rd.live.squashimg=$img panic=5 rd.neednet=1 rd.cos.oemlabel=@OEM_LABEL@"
 else
-    set kernelcmd="console=tty1 console=ttyS0 root=LABEL=$label cos-img/filename=$img panic=5 security=selinux selinux=1 rd.neednet=1 rd.cos.oemlabel=@OEM_LABEL@"
+    set kernelcmd="console=tty1 console=ttyS0 root=LABEL=$label cos-img/filename=$img panic=5 security=selinux selinux=1 rd.neednet=1 rd.cos.oemlabel=@OEM_LABEL@ fsck.mode=force fsck.repair=yes"
```
davidcassany, line 5: I don't see any `fsck.repair` definition, hence I guess this is something that  `systemd-fsck` parses, right? If this something custom from our side I'd suggest defining it with the `rd.cos...` prefix, I think it is not the case though, just to double check.

davidcassany (approved):
Looks good to me, nice :+1:

Just a couple of nits regarding the module dependencies and `systemd-fsck` absolute path usage.

packages/immutable-rootfs/build.yaml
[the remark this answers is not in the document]
mudler: yup exactly, this happens to be in two different places - tw and leap have it differently

packages/immutable-rootfs/30cos-immutable-rootfs/cos-loop-img.sh
mudler: I know, it frightens me as well, but the fact that systemd-fsck is not having a binary in /sbin etc hold me off from doing any special setup. I just wonder why is not installed as a normal binary to begin with ?!

packages/grub2/config/bootargs.cfg.tmpl
mudler: yup, those are read by systemd-fsck directly, so we don't need to do any special handling ourselves :partying_face:

packages/immutable-rootfs/30cos-immutable-rootfs/cos-loop-img.sh
mudler: Seems it is? https://github.com/dracutdevs/dracut/blob/d502d2a816ba8f8329b3d8616bd2a7e82a0ad21f/modules.d/00systemd/module-setup.sh#L39

mudler: mmh but might not be just available when calling it inside scripts. I just wonder how dracut handles it as it used already?

mudler: should be fixed now!

packages/immutable-rootfs/30cos-immutable-rootfs/module-setup.sh
mudler: ouch good catch, I think it slipped in while I was testing

mjura (approved):
LGTM

davidcassany (approved):
Nice!

PR #1356, opened 2022-05-23
Workaround for live packages on tumbleweed

Itxaka:
Current grub implementation for live isos requires several x86 AND i386 packages
which are no longer available under tumbleweed.

This patch works around it by using a leap image to build those grub
live packages if we are on aarch64+tumbleweed

### DIFF
--- a/packages/livecd/grub2/build.yaml
+++ b/packages/livecd/grub2/build.yaml
@@ -1,7 +1,14 @@
+{{ if .Values.arch }}
+  {{if and (eq .Values.arch "aarch64") (ne .Values.codename "green") }}
+image: {{ .Values.live_image }}
+  {{else}}
 requires:
 - name: "tool"
   category: "distro"
   version: ">=0"
+{{end}}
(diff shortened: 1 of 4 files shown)


packages/livecd/grub2/build.yaml
```diff
@@ -1,7 +1,14 @@
+{{ if .Values.arch }}
+  {{if and (eq .Values.arch "aarch64") (eq .Values.distribution "opensuse") }}
+image: {{ .Values.live_image }}
+  {{else}}
```
mudler, line 4: I'm wondering.. wouldn't this apply for teal too? Shouldn't be values adapted there too?

mudler (approved):

packages/livecd/grub2/build.yaml
Itxaka: Yes indeed, I wanted to see if this built as expected before bringing it tomorrow morning in the meeting

PR #1358, opened 2022-05-24
Revert tool_image change to tumbleweed

Itxaka:
tool_image is used for static bin (golang) or artifacts (firmware, grub2
artifacts, live/grub artifacts) so it makes no sense to update it to
tumbleweed as that brings in more issues like missing packages

This patch reverts the tool_image to leap:15.3


mudler (approved):

Itxaka:
Everything seems to build correctly, teal was not changed so merging this to get packages published asap.

Issue #1258, opened 2022-04-26
elemental: support building device images


Itxaka:
This card is about to port required scripts from [https://github.com/rancher-sandbox/cOS-toolkit/tree/master/images](https://github.com/rancher-sandbox/cOS-toolkit/tree/master/images?rgh-link-date=2022-04-12T07%3A11%3A37Z) as elemental build-disk command option.

Current build-disk is command is very focused on building vanilla/recovery images. Device images are full images, with extra scripts and so on, and they look closer to a full install than a vanilla/recovery image.

Investigate how to port those device scripts to the new elemental system

PR #1360, opened 2022-05-25
Remove squashfs iso

### LINKED ISSUE #1285: Drop recovery image as default package included when building isos
Currently the CI creates images with the recovery included first, that then gets dropped during later stages.

This card is about to move it the other way around, so we have default images without recovery and we add the recovery in the CI steps that actually needs it

Tied to: https://github.com/rancher-sandbox/cOS-toolkit/issues/1270

davidcassany:
Fixes #1285 


davidcassany:
This PR incorporates two related changes:

* `recovery/cos-img` is no longer used to build the isos (even removed from packages tree), so the default `manifest.yaml` does not include a separate squashed recovery image.
* the `iso-squashfs-<flavor>` job is removed, since elemental can be instructed to install custom system and recovery-system images, instead of building an iso including an extra image we instruct elemental to use the installation media root squashfs image, which is the same image contents used for active, but in its squashfs form.
 
This can be achieved with the following `config.yaml`:

```yaml
install:
  recovery-sytem:
    fs: squashfs
    uri: file:/run/initramfs/live/rootfs.squashfs
```

With these changes we are omit to build an extra iso in CI and the above configuration file allows to use squashfs images without having to actually create the squashfs image at run time.

Note the same `config.yaml` as above without the image uri should also work, but it would create (running `mksquashfs`) the squashed image from the mounted data on each installation. The default uri is something like `uri: dir:/run/rootfsbase`, which is indeed the mountpoint for `/run/initramfs/live/rootfs.squashfs`.

davidcassany:
I am not very convinced about the packer config changes, but it seams to work. I find HCL config files (to call them something) pretty esoteric. If we manage to implement #1258 and use it to create images within the CI packer changes from this PR could be reverted.

mudler:
> I am not very convinced about the packer config changes, but it seams to work. I find HCL config files (to call them something) pretty esoteric. If we manage to implement #1258 and use it to create images within the CI packer changes from this PR could be reverted.

yep HCL isn't very nice indeed, it looks good here anyhow, In the future when #1258 will be delivered we can just keep packer more confined in our CI process, just as a separate build to guarantee that can be still used but no need to attach it to the full CI cycle anymore

mudler (approved):
Looking good!

davidcassany:
> yep HCL isn't very nice indeed, it looks good here anyhow, In the future when #1258 will be delivered we can just keep packer more confined in our CI process, just as a separate build to guarantee that can be still used but no need to attach it to the full CI cycle anymore

Exactly this is what I had in mind, keep the packer build, but just as simple **it can build** test, without the need of running multiple paths.

Itxaka (approved):
nice

davidcassany:
Requires https://github.com/rancher-sandbox/elemental/pull/229

davidcassany:
Rebased, included https://github.com/rancher-sandbox/elemental/pull/229 and added couple of little fixes in installation and recovery tests.

PR #1363, opened 2022-05-26
Link teal arm64 kernel to /boot/vmlinuz

### LINKED ISSUE #1342: [teal][arm64] building broken
This is your weekly arm64 broken card.

 - [x] installer is not picking up the proper arch so it tries to install x86 grub on arm64 https://github.com/rancher-sandbox/elemental/pull/209
 - [ ] kernel packages dont have the vmlinuz link so an installed system does not boot
 - [ ] upgrade tests wont work until the above is merged and the test version updated in sut.go

more to come?

Itxaka:
Otherwise installed systems wont boot as they are looking for that

Should probably fix #1342 


Itxaka:
Requesting reviews now, unfortunately until we publish this current fixed package, we wont be able to make the upgrade test jobs work, as this is the current version which fixes the boot, we cannot upgrade to an older version as expect it to boot. Once this is merged and published a follow up PR will come to bump the test version in order to make the upgrade tests work on arm64.

tests/sut/sut.go
```diff
@@ -31,9 +31,10 @@ const (
 
 	TimeoutRawDiskTest = 600 // Timeout to connect for recovery_raw_disk_test
 
-	Ext2 = "ext2"
-	Ext3 = "ext3"
-	Ext4 = "ext4"
+	Ext2         = "ext2"
```
mudler, line 34: mmm damn, we have been moving this file out from the repo with #1368 so we can share it with different projects, maybe can put those changes directly in the [testhelper](https://github.com/rancher-sandbox/ele-testhelpers/blob/main/vm/sut.go) repo?

Itxaka: sure, let me first use this to get the actual log of the upgrade tests because somehow is resetting to a wrong version and I have no visibility over that

mudler (approved):

tests/installer/installer_efi_test.go
```diff
@@ -220,7 +220,7 @@ var _ = Describe("cOS Installer EFI tests", func() {
 			})
 			It("forces efi", func() {
 				By("Running the installer")
-				out, err := s.Command("elemental install --force-efi /dev/sda")
+				out, err := s.Command(fmt.Sprintf("%s install --force-efi /dev/sda", sut.ElementalCmd))
```
mudler, line 223: mmm maybe we could even do an `s.Elemental` helper directly?

Itxaka: yep, that should be the endgame, `s.Elemental(arg1, arg2, arg3)`

PR #1364, opened 2022-05-27
Check source mount path in mountState() and mountOverlay()

### LINKED ISSUE #1311: Boot stalls when peristent folders are missing in rootfs
This card is about enhancing cOS to bail out in case a folder which is listed as persistent is missing in the rootfs

If a derivative describes a layout like the following:
```yaml
  rootfs:
    - if: '[ ! -f "/run/cos/recovery_mode" ]'
      name: "Layout configuration"
      environment_file: /run/cos/cos-layout.env
      environment:
        VOLUMES: "LABEL=COS_OEM:/oem LABEL=COS_PERSISTENT:/usr/local"
        OVERLAY: "tmpfs:25%"
        RW_PATHS: "/var /etc /srv"
        PERSISTENT_STATE_PATHS: >-
          /etc/systemd
          /etc/rancher
          /etc/ssh
          /etc/iscsi 
          /etc/cni
          /home
          /opt
          /root
          /usr/libexec
          /var/log
          /var/lib/rancher
          /var/lib/kubelet
          /var/lib/wicked
          /var/lib/longhorn
          /var/lib/cni
```
And any of the folder in `PERSISTENT_STATE_PATHS` are missing in the rootfs, a cOS derivative boot will get stuck without any prompt

We should fail in the initramfs stage if any of the folders are missing or at least display an  error message so it is clear what's going on.

N/A

**Additional context**
https://github.com/rancher-sandbox/rancher-node-image/pull/5

mjura:
Check all paths declared in PERSISTENT_STATE_PATHS and take care about
array reference for it.

Fixes #1311

### DIFF
--- a/packages/immutable-rootfs/30cos-immutable-rootfs/cos-mount-layout.sh
+++ b/packages/immutable-rootfs/30cos-immutable-rootfs/cos-mount-layout.sh
@@ -152,6 +152,8 @@ function mountOverlay {
 
     mount="${mount#/}"
     merged="/sysroot/${mount}"
+    # Check if source mount path does exist
+    [ ! -d "${merged}" ] && return
     if [ "${base##/run}" == "${base}"  ]; then
         base="/sysroot${base}"
     fi
@@ -180,9 +182,11 @@ function mountState {
     if [ "${state_bind}" = "true" ]; then
         mount="${mount#/}"
         base="/sysroot/${mount}"
+       # Check if source mount path does exist
+        [ ! -d "${base}" ] && return
         state_dir="/sysroot${state_target}/${mount//\//-}.bind"
         if ! mountpoint -q "${base}"; then
-            mkdir -p "${base}" "${state_dir}"
+            mkdir -p "${state_dir}"
             rsync -aqAX "${base}/" "${state_dir}/"
             mount -o defaults,bind "${state_dir}" "${base}"
             fstab_line="${state_dir##/sysroot} /${mount} none defaults,bind 0 0\n"


packages/immutable-rootfs/30cos-immutable-rootfs/cos-mount-layout.sh
```diff
@@ -109,10 +109,13 @@ function readCOSLayoutConfig {
...
     # applied when RW_PATHS is unset.
     if [ -n "${RW_PATHS+x}" ]; then
-        rw_paths=(${RW_PATHS})
+        rw_paths=(${RW_PATHS[@]})
     fi
     if [ -n "${PERSISTENT_STATE_PATHS}" ]; then
-        state_paths=(${PERSISTENT_STATE_PATHS})
+        state_paths=(${PERSISTENT_STATE_PATHS[@]})
+        for path in "${state_paths[@]}"; do
+            [ ! -d "${path}" ] && echo "Path ${path} from PERSISTENT_STATE_PATHS does not exist!" && exit 0
```
mudler, line 117: I think we should rather check that on the rootfs, in this way it's going to check the ones in the initramfs? cc @davidcassany

mjura: 1. I agree with assigning VARIABLES to array and I have dropped this change

2. I have added additional check to functions `mountOverlay`and `mountState`. Now it will check first if source mount path exists, if not then will return from function and continue

packages/immutable-rootfs/30cos-immutable-rootfs/cos-mount-layout.sh
```diff
@@ -150,6 +150,7 @@ function mountOverlay {
     local workdir
     local fstab_line
 
+    [ ! -d "${mount}" ] && echo "Source mount path ${mount} does not exist!" && return
```
davidcassany, line 153: We need to test against `${merged}` variable. Note this is executed inside initrd before switching the real root. At this stage the real root is at mounted at `/sysroot`, so in order to check if the directory is there we should test against `/sysroot/<dir>`, which is the `${merged}` value in this script.

In addition I think we will have to omit the `echo ` message. Note this method is expected to `echo` an fstab line of the mounted path, hence whatever we `echo` here will end up being part of the `/etc/fstab`.

packages/immutable-rootfs/30cos-immutable-rootfs/cos-mount-layout.sh
```diff
@@ -177,6 +178,7 @@ function mountState {
     local fstab_line
     local state_dir
 
+    [ ! -d "${mount}" ] && echo "Source mount path ${mount} does not exist!" && return
```
davidcassany, line 181: Same here as before, we should test against `${base}` and only check inside the `"${state_bind}" = "true"` path. Again we should not `echo` any message in this method. Line 187 `mkdir -p "${base}" "${state_dir}"` can be changed to `mkdir -p "${state_dir}"`, as we will know `${base}` exists at this stage.

davidcassany (changes requested):
I am fine with the spirit of the change, we are almost there. Just a couple of details to fix.

This code is sensitive and a bit convoluted as this is all happening on early boot before switching root.

packages/immutable-rootfs/30cos-immutable-rootfs/cos-mount-layout.sh
mjura: Thank you for your comments, I included all your suggestions

davidcassany (approved):
Looks good to me :+1:

mudler:
mmh package wasn't bumped, neither the cos collection, so test runs on this case didn't run at all!

PR #1368, opened 2022-05-27
Systemd timesyncd

fgiudici:


mudler (approved):

packages/cloud-config/oem/09_services.yaml
```diff
@@ -0,0 +1,8 @@
+name: "Default config"
+stages:
+   initramfs:
+     - name: "Default systemd config"
+       if: '[ -e "/sbin/systemctl" ] || [ -e "/usr/bin/systemctl" ] || [ -e "/usr/sbin/systemctl" ] || [ -e "/usr/bin/systemctl" ]'
```
Itxaka, line 5: what is wrong with systemctl lol, can it be in 4 different places?? :laughing:

Itxaka (approved):

packages/cloud-config/oem/09_services.yaml
mudler: argh yea, tried to be as much generic as possible here... maybe we should have some helpers in the system for such things :thinking:

kkaempf: The second one should be `/bin/systemctl`, no ?!
Otherwise the second and fourth are identical.

fgiudici: right... nice catch!

PR #1373, opened 2022-05-30
Bump cos and test helpers to fix arm64

Itxaka:
As only the latest 0.8.14-1 version has a fixed kernel for arm64, any
upgrade jobs that "upgraded" to an older version would fail, as after
the upgrade the VM will fail to boot due to the kernel issue.

Now that the issue is fixed and packages are published, we can bump the
cos-version to be tested to a know working version.


mudler (approved):
Looks good, nice catch!

PR #1374, opened 2022-05-30
Fix Autorelease

Itxaka:
Looks like scheduled events has minimal context so autoreleaser cannot infer the owner and the repo by itself


mudler (approved):
I was wondering indeed if something was wrong there.. nice catch, LGTM

Itxaka:
@mudler Im not even sure this will work...

PR #1383, opened 2022-05-31
Disable arm64 tests for now

Itxaka:
Signed-off-by: Itxaka <igarcia@suse.com>

PR #1384, opened 2022-05-31
Revert "Disable arm64 tests for now"

Itxaka:
Reverts rancher-sandbox/cOS-toolkit#1383

Once packages have been published for at least 2 different versions, we can merge this to re-enable tests for arm64

### DIFF
--- a/.github/config/master.yaml
+++ b/.github/config/master.yaml
@@ -52,7 +52,7 @@ flavors:
         repository: "releases"
         cache_repository: "build"
         organization: "quay.io/costoolkit"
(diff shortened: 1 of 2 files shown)


mudler:
jfyi there is no need to actually merge into master the skip test, I usually do that by pushing a branch with the skip_test enabled, and then associate that branch to be triggered in order to be published

mudler:
It didn't passed because the arm64 pipeline was somehow invalid https://github.com/rancher-sandbox/cOS-toolkit/actions/runs/2413330848

PR #1387, opened 2022-05-31
re enable arm64 tests

Itxaka:


Itxaka:
only affects master, merging

PR #1388, opened 2022-05-31
Use tool image with selinux policies

mudler:
Should fix https://github.com/rancher-sandbox/cOS-toolkit/runs/6671892893?check_suite_focus=true


mudler:
now... why the raw-disk-deploy test started to fail it's a mistery..

mudler:
ok, looks master have the same issue: https://github.com/rancher/elemental-toolkit/runs/6671951123?check_suite_focus=true, merging..

Issue #1365, opened 2022-05-27
Optimize CI tests suites


Itxaka:
With the team now growing, it seems that we have more PRs in parallel than ever, which results in a multitude of test jobs in CI.

Unfortunately, our test suite runs on macos and we are limited to a max of 5 runs in parallel.

We currently run 18 individual tests (9 squash, 9 non-squash) per teal PR x86 job. Each test takes between 9 and 41 minutes. We also run another 18 tests for each master merge. 

This results in a huge queue of test jobs in CI which really slows down things around, knowing that the test suite is meant to show any breaking changes on our PRs, this can result in spending a whole day waiting for the tests to fail and tell you where it failed.

We should try to find a way of running those tests somewhere else, somehow. either by using our custom runners, use the ubuntu runner provided by github, running aws/azure/whatever, or something else (dare I say it....maybe jenkins..... _runs_ ) altogether.

The main issue is that the tests use virtualbox as the provider. Now we have some examples of running some other tests on qemu, but those have no acceleration, so they may be even more slow?

We could start transitioning some of the more fast ones into qemu to compare, i.e test-features takes 9 minutes, test-smoke also 9 minutes, to compare the speed of one vs another, and leave the more heavy ones in virtualbox if they need it.

Itxaka:
argh, just remembered, one of the issues with qemu was that the ssh tunnel will get lost once we reboot, but that should be fixed if we use the proper networking like we do in other tests.

Itxaka:
network issues are resolved with the hack that we have in the arm64 tests of tying the mac address to an ip.

The problem is that it seems to be twice as slow as the virtualbox tests :/

davidcassany:
Yes we need to improve that, cos-toolkit CI costs too much. I also believe there is a great room of improvement by reducing the amount of tests we need. Having this squashfs and nonsquashfs path is way too heavy.

Probably we could start considering again some of the ideas from https://github.com/rancher-sandbox/cOS-toolkit/issues/1145, essentially either drop one of the two (`squashfs` or `nonsquashfs`) or eventually converge these two into a single boot path so no duplicated test are needed (I think there are chances to achieve that if we embed the fs image into the squashfs img).

Itxaka:
Yep, there is also a very weird thing In That we run boths paths but only require one for publishing artifacts, so IMHO those test not being gated mean absolutely nothing and should be either required or thrashed.

mudler:
[...]

PR #1467, opened 2022-06-01
ci: reduce tests running against squashfs

mudler:
Fixes: #1365

This leaves only recovery tests running on squashfs. It also makes it possible to tweak the test suites to run via pipeline config files


Itxaka (approved):

mjura (commented):
LGTM

PR #1473, opened 2022-06-03
Refactor raw disk

mudler:


mudler:
cc @Itxaka @davidcassany we picked up https://github.com/rancher/elemental-toolkit/issues/1346 for educational purposes (it looked like, one of the best fit to be simple enough to show the workflow). any feedback is really appreciated!

Itxaka:
This probably needs https://github.com/rancher/elemental-toolkit/pull/1500 in order for the build iso step to use the latest locally built elemental. Otherwise it will pick and run with the latest published, making it impossible to test those changes. In this case, its ignoring the latest changes to ignore repos from some arches, thus building the x86 iso with arm64 packages :)

Itxaka:
rebased

Itxaka:
Im confused, why is this failing? This works locally with no issues, so elemental doesnt seem to be the problem?

PR #1481, opened 2022-06-06
Move everything to yq4

### LINKED ISSUE #1379: Switch all to yq4
Currently our automation is double headed. 

There are pieces still leveraging yq3 (for example, pipelines generation) and some other using yq4.

This card is about to track all usage and converting to use yq4

Itxaka:
fixes #1379 

Also drops the non used img-builder script as that its not used anymore


.github/pipelines.sh
```diff
@@ -2,11 +2,18 @@
 set -e
 YQ="${YQ:-yq}"
 
+VERSION=$($YQ -V |rev|  cut -d " " -f 1 |rev| cut -d "." -f 1)
+if [[ "$VERSION" == "3" ]]; then
+    echo "yq version 3 detected, only version 4 is supported"
```
mudler, line 7: :heart:

.github/pipelines.sh
```diff
@@ -2,11 +2,18 @@
...
+    exit 1
+fi
+
 for i in $(ls config); do
     filename=$(basename -s .yaml $i)
-    for flavor in $(${YQ} r config/"$i" -p "p" 'flavors.[*]' | cut -d "." -f 2); do
-      for arch in $(${YQ} r config/"$i" -p "p" "flavors.$flavor.arches.[*]" | cut -d "." -f 4); do
-        ${YQ} r -X config/"$i" "flavors.$flavor.arches.$arch" | gomplate --left-delim "{{{" --right-delim "}}}" -V --datasource config="stdin:/?type=application/yaml" --file build.yaml.gomplate --out workflows/build-"$filename"-"$flavor"-"$arch".yaml
+    for flavor in $(${YQ} '.flavors|keys|join(" ")' config/"$i"); do
+      for arch in $(${YQ} ".flavors.$flavor.arches|keys|join(\" \")" config/"$i" ); do
```
mudler, line 14: huh, much uglier!

mudler (approved):

.github/pipelines.sh
Itxaka: Im not sure if leaving that in place and add it everywhere or just drop it altogether

.github/pipelines.sh
Itxaka: yes, I want to go afterwards and try to get it in a different way, Im pretty sure we can spit pairs of flavor+arch like we did on the img-builder script instead.

Although Im not sure which one is uglier lol, mixing bash and yaml ist not nice at all.

images/arm-img-builder.sh
```diff
@@ -216,12 +216,6 @@ if [ -z "$output_image" ]; then
 fi
 
 if [ -n "$manifest" ]; then
-  YQ_VERSION=$( yq -V | cut -d " " -f 3 | cut -d "." -f 1)
```
mudler, line 219: mm this seems to have broken the CI here: https://github.com/c3os-io/c3os/runs/6760298289?check_suite_focus=true#step:11:150

PR #1482, opened 2022-06-06
Drop COSIGN_REPOSITORY from our side

Itxaka:
There is no need to push to a different repo the signatures as now quay
supports the cosign signatures.

The value its still there as its picked by cosign, so in the end there
is still the possibility of using it when building a derivative for
example, but its not needed anymore on our side so lets drop it and
update the resigner as well.

Also drops 2 old jobs that were not working


Itxaka:
devices is expected to fail for now, Im triggering the resigner with the latest repository.yaml references to push them to the new repos, after that it should start working again

mudler (approved):
dang, I thought I have approved it :sweat_smile:

PR #1483, opened 2022-06-06
Resigner: Also sign repo own files

Itxaka:
We currently not signing the repo files (compilertree, metadata, etc...)
so the resigner its not really resigning the whole
releases/point-in-time files as we also check those files when trying to
install a package, a referenceID re-resigned will never work

### DIFF
--- a/.github/resign.go
+++ b/.github/resign.go
@@ -11,7 +11,8 @@ import (
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


Itxaka:
no need to wait for CI for this BTW.

mudler (approved):

mjura (commented):
LGTM

PR #1486, opened 2022-06-07
set SELinux in enforce mode

### LINKED ISSUE #1326: Elemental SELinux support
As part of https://github.com/rancher-sandbox/os2/pull/119 SELinux was disabled from cmdline options.

Previous versions of os2 (not based on SLE Micro for Rancher) enabled selinux, but as the base image (openSUSE) didn't had support for it, it was silently ignored.

With SLE Micro for Rancher that has full SELinux support, os2 was not capable of booting anymore, as we don't have SELinux profiles.

This card is about having basic SELinux profile which would allow cOS to be functional under SELinux environments.

Currently, from my tests, even booting is prevented, as access to `/usr/local` is denied for several components. To replicate, is enough booting os2 with the `security=selinux selinux=1` boot commands.

See also: https://rancher-sandbox.github.io/cOS-toolkit/docs/reference/selinux/ and check if the elemental-cli  can re-apply the SELinux labels after install/upgrade/reset.

## Action items
- Check if the elemental-cli SELinux relabel section is working correctly
- Create a basic SELinux profile for cOS

davidcassany:
This PR adds an elemental SELinux module, adds a selinux specific cloud config to relabel ephemeral file systems and sets SELinux in enforcing mode. 

The elemental selinux module is meant to be used for illustration purposes and to be extended/modified as required by specific use cases, it should **not** be used as is. It was created by simply feeding denials logged at boot time in permissive mode to `audit2allow` utility without any further fine tuning.

Fixes #1326


packages/cloud-config/oem/10_selinux.yaml
```diff
@@ -0,0 +1,16 @@
+# Default cOS OEM configuration file
```
davidcassany, line 1: All this file is pretty tied to the immutable layout so I am wondering to what extend this should be part or not of the immutable-rootfs dracut module and include there an early boot systemd unit that does this thing if selinux is active (`selinux=1` kernel parameter or active in `/etc/selinux/config`). 

Probably we could apply the restorecon to any `rw` path and `rw` paths are managed by the immutable rootfs module. Anyway I do not want to modify immutable rootfs module code now, as changes there tend to strike back sooner or later.

davidcassany:
I believe we are hitting https://github.com/SELinuxProject/selinux/issues/343 here, on my host I don't see that as I am using the btrfs driver, hence there are no filesystem changes between the base layer and the runtime layer. However I assume we have an overlayfs driver in our CI worker.

The fix for that is already included upstream since last march in SELinuxProject/selinux@c7a3b93e3 and it is released under 3.4 version. Currently we are on 3.1 in SLE, so there is a gap to fill somehow, I am trying to figure out if there are options to workaround it, like trying to all these files into upper overlayfs layer before libsemanage kicks in... however up to now I had no success.

davidcassany:
I believe we are hitting https://github.com/SELinuxProject/selinux/issues/343 here, on my host I don't see that as I am using the btrfs driver, hence there are no filesystem changes between the base layer and the runtime layer. However I assume we have an overlayfs driver in our CI worker.

The fix for that is already included upstream since last march in SELinuxProject/selinux@c7a3b93e3 and it is released under 3.4 version. Currently we are on 3.1 in SLE, so there is a gap to fill somehow, I am trying to figure out if there are options to workaround it, like trying to all these files into upper overlayfs layer before libsemanage kicks in... however up to now I had no success.

davidcassany:
This PRs has few follow ups that should be tracked in separate cards IMHO:

- [ ] Add a relabel step in immutable rootfs dracut module after initramfs if selinux is activated
- [ ] Discuss configuration options in `config.yaml` related to SELinux (for instance, should we relabel persistent data on upgrade?, enable/disable modules?)

mudler (approved):

packages/cloud-config/oem/10_selinux.yaml
```diff
@@ -0,0 +1,16 @@
...
+# consider copying this file to /usr/local/cloud-config or
+# copy the file with a prefix starting by 90, e.g. /oem/91_custom.yaml
+name: "SELinux"
+stages:
+   initramfs:
+     - name: "Relabelling"
+       commands:
+       - |
+           load_policy -i
+           restorecon -R -i -v /etc /root /opt /srv /var /home /usr/local /oem
```
Itxaka, line 16: This doesnt need an input file? How does that work? uses the default file in the selinux dir (/var/selinux or whatever)

mudler:
> This PRs has few follow ups that should be tracked in separate cards IMHO:
> 
>     * [ ]  Add a relabel step in immutable rootfs dracut module after initramfs if selinux is activated
> 
>     * [ ]  Discuss configuration options in `config.yaml` related to SELinux (for instance, should we relabel persistent data on upgrade?, enable/disable modules?)
> 
>     * [ ]  Document SELinux usage from a derivative perspective (it is unclear to me where this should go, as part of elemental-toolkit docs?):
>       
>       * `elemental-selinux` does not install the elemental module `semodule -i /usr/share/elemental/selinux/elemental.pp` is needed.
>       * Limitations of overlayfs driver for libsemanage prior to v3.4 and possible work arounds
>       * Point to elemental module sources (`/usr/share/elemental/selinux/elemental.te`) to be extended/modified or substituted by some other custom module
>       * For the time being document on boot relabeling in `initramfs` stage.

Can you create relevant cards? This looks fine. For the docs, I'd use elemental-toolkit as docs here would apply for every derivative, and config-wise this spans over multiple areas (from cloud config to the Dockerile) so that would make the best place to document it imho.

mudler:
...nice work! nailing down SELinux support isn't at all a trivial thing :sweat_smile:

packages/elemental-selinux/build.yaml
```diff
@@ -0,0 +1,10 @@
+{{if eq .Values.codename "teal" }}
+requires:
+- name: "base"
+  category: "distro"
+  version: ">=0"
+
+steps:
+- mkdir -p /usr/share/elemental/selinux
+- cp elemental.pp elemental.te /usr/share/elemental/selinux
```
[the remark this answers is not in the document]
davidcassany: Yes, this one of the parts to be documented under the specific docs.

davidcassany:
Merging, all tests went green after rebasing. However worth to mention that the arm recovery tests took around more than 2h 30m to pass... 

We should configure no squashfs compression at all for arm builds...

PR #1490, opened 2022-06-07
Add slack post on master/nigthly/release jobs

Itxaka:
Configurable in the jobs config by setting slack_on_failure: false this
will post a message with the job and a link to the job in case of
failure.

Currently enabled for nigthly-teal, master-teal and releases-all


mudler (approved):
nais!!

Itxaka:
tested and seems to work.

PR #1500, opened 2022-06-10
Call deps_ci after adding the build repo

Itxaka:
Otherwise we cannot bump packages and use them, i.e. elemental in all
phases of the build


frelon (approved):
Nice :+1:

PR #1501, opened 2022-06-10
Bump elemental client and initrd to include latest changes

davidcassany:

### DIFF
--- a/packages/cos/collection.yaml
+++ b/packages/cos/collection.yaml
@@ -2,7 +2,7 @@ packages:
   - &cos
     name: "cos"
     category: "system"
-    version: 0.8.14-16
+    version: 0.8.14-17
     description: "cOS base image, used to build cOS live ISOs"
     brand_name: "cOS"
     labels:
--- a/packages/initrd/definition.yaml
+++ b/packages/initrd/definition.yaml
@@ -1,4 +1,4 @@
 name: "dracut-initrd"
 category: "system"
-version: 0.3-2
+version: 0.3-4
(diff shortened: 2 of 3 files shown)


davidcassany:
I realized the generated initrd was not including the latest changes from the immutable-rootfs package as the intird was not recreated on the build in this [run](https://github.com/rancher/elemental-toolkit/runs/6816450272?check_suite_focus=true), hence the changes from the PR were not included within the resulting image.

Itxaka (dismissed):

davidcassany:
I noted there is an error on `elemental-cli build-iso` error while parsing arguments. With the latest changes in build-iso we omitted the option of setting all sources and ISO details from the manifest, so it can also work if no arguments are passed and the `manifest.yaml` includes the minimum data. A PR with the fix on elemental-cli side is coming.

davidcassany:
Bumped `elemental-cli` to include https://github.com/rancher/elemental-cli/pull/256

davidcassany:
@Itxaka I am not sure what to do with this PR. I don't want to step over what you are doing in arm64, as far as I understand if we publish a new elemental-cli the revert commit here should not be needed anymore right?

Shouldn't we just merge a version bump on elemental-cli and wait until it is published?

Itxaka:
It's now published, arm64 was lagging behind, but the latest elemental-cli is on the repos and on master the manifest now uses the channel stuff for the packages so it should be ok, you should be able to only bump the initrd stuff and forget about Elemental cli

PR #1505, opened 2022-06-13
Fix repo file output name for releases

Itxaka:


Itxaka:
no need to wait for anything

PR #1507, opened 2022-06-13
Split recovery tests in individual tests

Itxaka:
Using labels so we can run them individually for arm64, as they take too
long

It also uses a bit of trickery to determine if we are running the tests
with labels, which would indicate that we are running them individually
and skips the reset part to speed them up even further.

Currently only done for arm64, x86 tests are still run on the whole
suite

### DIFF
--- a/.github/config/master.yaml
+++ b/.github/config/master.yaml
@@ -69,7 +69,10 @@ flavors:
         skip_tests: false
         run_tests:
           squashfs:
-            - "test-recovery"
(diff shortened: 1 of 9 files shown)


mudler (approved):

davidcassany (approved):
Looks good :+1:

Itxaka:
This has already passed before and it works as expected, only fixed one of the tests targets running 2 tests instead of one.

PR #1511, opened 2022-06-14
Revert "Call deps_ci after adding the build repo"

Itxaka:
Reverts rancher/elemental-toolkit#1500

This works as long as we are building on the same arch. Once we cross arches (arm64 build, running on x86) all hell breaks loose as it will install local deps for CI from the arm64 branch :(


Itxaka:
This only hits the issue when a dependency is bumped and built for arm64. Otherwise it will install from the default repo and we will not hit this.

frelon (approved):

PR #1514, opened 2022-06-15
Revert "Revert manifest until new elemental is pushed"

Itxaka:
Reverts rancher/elemental-toolkit#1513

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -1,11 +1,11 @@
 iso:
   rootfs:
-    - system/cos
+    - channel:system/cos
   uefi:
-    - live/grub2-efi-image
+    - channel:live/grub2-efi-image
   image:
-    - live/grub2
-    - live/grub2-efi-image
+    - channel:live/grub2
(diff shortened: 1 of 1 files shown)


Itxaka:
Now that the new elemental has been pushed and will be consumed on the iso/raw disk steps, we can update the manifest.yaml

Itxaka:
oh yeah, this wont run any jobs because the manifest.yaml is not included in the files to trigger the jobs....

frelon (approved):

davidcassany:
Ouch, I was about to say this will conflict with #1501... 😅 as I had to include that in there.

Also I am afraid with the current steps order in build-iso job, the new elemental won't be used. In #1501 I had to revert the revert of `make ci_deps` steps.

Itxaka:
yes, new elemental wont be used and we have to choose between using the new elmental or fully breaking arm64 (again lol): https://github.com/rancher/elemental-toolkit/issues/1515

Issue #1515, opened 2022-06-15
New elemental packages are not inmediatly consumed by CI


Itxaka:
Build iso and build raw images are using the default elemental-cli from the releases-teal repos, which means that if we bump them to use new features, they wont be used until the PR has been merged and the artifact published.

An easy fix would be to instead consume the elemental-cli in those steps from the published docker image, as latest should always point to the latest released, so we could cut a release when new breaking changes are done on the cli side and the CI would consume that.

We tried adding the build repo, but due to cross-compiling that would still fail. Like we run the commands on x86, but if we are building for arm64, we cannot use the local build dir as repo, as that will install a wrong arch version.

davidcassany:
Yes, using the docker image sounds like a reasonable effective and simple solution.

I have also been wondering if we aren't abusing of the `manifest.yaml` file in elemental-toolkit. I mean, there we have the build stanza that does not belong to an elemental OS manifest, it is just for our own automation:

```yaml
# Additional packages to build                                                  
build:                                                                          
  - utils/nerdctl                                                               
  - utils/k9s                                                                   
  - utils/jq                                                                    
  - selinux/rancher                                                             
  - selinux/k3s                                                                 
  - recovery/cos                                                                
  - system/grub2-artifacts                                                      
  - system/grub2-efi-image                                                      
  - toolchain/elemental-cli  
```

I think this should not be part of the manifest and we should handle package builds we need in CI beyond the OS tree in a different way/place. Probably it is enough to strip it to another file. I was wondering about that because I was thinking on building the CI dependencies separately always with the host arch, not the target arch.

Also something that is probably worth double checking is why, with the current setup, luet attempts to install packages from another arch on cross builds, the expected behavior should be that packages are simply not found. Looks like we might have a wrong (or not precise enough) repo configuration somewhere.

Itxaka:
Couple of things, first, I don't think we are using that package list anymore as we use the build.go facility to identify packages missing and build those so it should be easy to drop. Second is that in order for luet to enable/disable a repo based on arch, the repo must be disabled and have a arch key. Then it compares it to the running arch and if they match it enables it. That would allow us to have it on the deps_ci part and automatically enable/disable stuff. Not sure how would that affect cross repo builds, as we don't use the system repos it should be safe enough.

davidcassany:
> Couple of things, first, I don't think we are using that package list anymore as we use the build.go facility to identify packages missing and build those so it should be easy to drop

[...]

PR #1518, opened 2022-06-16
Add cut utility in initrd, fixes #1517

### LINKED ISSUE #1517: #1516 introduced missing "cut" initrd dependency, breaking boot
**cos-toolkit version:**
master

amd64

**Describe the bug**
#1516 introduced a loop with cut command not found. System appears to hang until it drops to the rescue prompt.

**To Reproduce**
Build from latest green amd64 image and attempt to boot.

**Expected behavior**
Boot normally.

**Logs**
image

mount:
image

davidcassany:

### DIFF
--- a/packages/cos/collection.yaml
+++ b/packages/cos/collection.yaml
@@ -2,7 +2,7 @@ packages:
   - &cos
     name: "cos"
     category: "system"
-    version: 0.8.14-18
+    version: 0.8.14-19
     description: "cOS base image, used to build cOS live ISOs"
     brand_name: "cOS"
     labels:
--- a/packages/immutable-rootfs/30cos-immutable-rootfs/module-setup.sh
+++ b/packages/immutable-rootfs/30cos-immutable-rootfs/module-setup.sh
@@ -25,7 +25,7 @@ install() {
     declare initdir="${initdir}"
 
     inst_multiple \
-        mount mountpoint elemental sort findmnt rmdir findmnt rsync
+        mount mountpoint elemental sort findmnt rmdir findmnt rsync cut
 
     # Include utilities required for cos-setup services,
     # probably a devoted cos-setup dracut module makes sense
--- a/packages/immutable-rootfs/definition.yaml
+++ b/packages/immutable-rootfs/definition.yaml
@@ -1,6 +1,6 @@
 name: "immutable-rootfs"
 category: "system"
-version: 0.6-4
+version: 0.6-5
 requires:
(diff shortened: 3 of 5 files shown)


davidcassany:
This PR as is will break arm64 builds due to e420dce37f0603dc20d935c2b061b9daccb1dccf and 9546c1d5bf7ace227699bdbb48db0cb33f3e0420 can be merged as is as this is just to validate the elemental-cli fix from rancher/elemental-cli#260 solves the issue.

mudler (approved):
Looks good!

.github/build.yaml.gomplate
```diff
@@ -277,6 +276,7 @@
         run: |
           export DOCKER_INSTALL=true
           sudo -E make add_local_repo
+      {{{ tmpl.Exec "make" "deps_ci" }}}
```
mudler, line 279: we have been into this already several times... hoping it's the good one :crossed_fingers:

davidcassany:
Ok this turns to be tricky, in our repos we have published a broken image. Due to #1519 the error this PR fixes was not caught on CI and a non bootable `system/cos` was published.  This PR includes the fix for #1519 so the proper artifacts are test and it also includes the fix for #1517 to make `system/cos` bootable again. However this will not pass the tests because our upgrade/install might install the latest published image, which is the broken one, hence they will fail.

Then even if we force merge there is the problem that the CI on master will fail and hence not publish the image.

@mudler any idea on how to work around that?

.github/build.yaml.gomplate
davidcassany: No this not the good one... 😞  this will break arm64 builds, we are still trapped on #1515, I just added it to see if I could manage to get a green CI for x86_64 at least. But the CI won't be green due to https://github.com/rancher/elemental-toolkit/pull/1518#issuecomment-1157652688

Pretty disappointed, we have several things to digest out of this issue chain of issues.

mudler: correct, indeed I'd treat https://github.com/rancher/elemental-toolkit/issues/1515 separately and forget about it at the moment. We weren't capable of doing that before, so even if frustrating, I'd say to just let it go..

davidcassany: yes gonna remove this commit, just I considered it when I thought it was enough to get a green CI...

davidcassany:
@mudler ok I think for now your approach is the best one. Merge this and then branch from master to publish directly.

PR #1523, opened 2022-06-17
Check filesystem on all found devices in initqueue

davidcassany:

### DIFF
--- a/packages/immutable-rootfs/30cos-immutable-rootfs/cos-loop-img.sh
+++ b/packages/immutable-rootfs/30cos-immutable-rootfs/cos-loop-img.sh
@@ -14,6 +14,9 @@ function doLoopMount {
         # Ensure run system-fsck, at least, for the root partition
         systemd-fsck "${partdev}"
 
+        # Only run systemd-fsck if root is already found
+        [ "${found}" == ok ] && continue
+
         mount -t auto -o "${cos_root_perm}" "${partdev}" "${cos_state}" || continue
         if [ -f "${cos_state}/${cos_img}" ]; then
 
@@ -22,7 +25,7 @@ function doLoopMount {
             # attempt to run systemd-fsck on the loop device
             systemd-fsck "${dev}"
 
-            exit 0
+            found="ok"
         else
             umount "${cos_state}"
         fi
@@ -36,6 +39,7 @@ PATH=/usr/sbin:/usr/bin:/sbin:/bin
 declare cos_img=$1
 declare cos_root_perm="ro"
 declare cos_state="/run/initramfs/cos-state"
+declare found=""
 
 [ -z "${cos_img}" ] && exit 1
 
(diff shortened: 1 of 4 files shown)


davidcassany:
@mudler finally this ended amost as you first implementation 😅 At the end there only a couple of minor changes:

* devices detected by `lsblk` and do not relay on udev links (e.g. `/dev/disk/by-partuuid`) which could be missing if corrupted and MSDOS vs GPT differ on available identifiers.
* It does the check in the main loop so if the script is executed more than once the filesystem check and loop image check are not executed over the same devices again

mudler (approved):

mudler:
yeah it is not obvious until you deep dive into it, quite time consuming also to try that out... anyhow , looks good!

davidcassany:
> yeah it is not obvious until you deep dive into it, quite time consuming also to try that out... anyhow , looks good!

Yes indeed. In fact the current implementation only works for the test cases if all partitions are found in `cos-loop-img` stage, for which there are no guarantees. Since all partitions are in the same device udev catches all them at once, but if something like COS_PERSISTENT is pointing to another device that appears later than COS_STATE we are screwed, not only for the systemd-fsck, also for the `cos-mount-layout.sh` script and `initramfs` stage.

So a good practice I guess it should be keeping all the partitions that are required at boot (initrd stage) into a single device and keep any other persistent device out of the layout config and add them as regular systemd units in initramfs stage. I guess this is not much of a problem for the edge, but for the cloud I think this is a valid concern.

mudler:
> > yeah it is not obvious until you deep dive into it, quite time consuming also to try that out... anyhow , looks good!
> 
> Yes indeed. In fact the current implementation only works for the test cases if all partitions are found in `cos-loop-img` stage, for which there are no guarantees. Since all partitions are in the same device udev catches all them at once, but if something like COS_PERSISTENT is pointing to another device that appears later than COS_STATE we are screwed, not only for the systemd-fsck, also for the `cos-mount-layout.sh` script and `initramfs` stage.
> 
> So a good practice I guess it should be keeping all the partitions that are required at boot (initrd stage) into a single device and keep any other persistent device out of the layout config and add them as regular systemd units in initramfs stage. I guess this is not much of a problem for the edge, but for the cloud I think this is a valid concern.

That would be enough imho to mention in the docs as specific case. It is a fine tuning-based on the derivative scope, we could also provide an example so we don't miss the gained knowledge here

packages/immutable-rootfs/30cos-immutable-rootfs/cos-loop-img.sh
```diff
@@ -14,6 +14,9 @@ function doLoopMount {
         # Ensure run system-fsck, at least, for the root partition
         systemd-fsck "${partdev}"
 
+        # Only run systemd-fsck if root is already found
+        [ "${found}" == ok ] && continue
```
kcburge, line 18: Should this no be quoted "ok"?

PR #1560, opened 2022-07-07
Set fixed tool_images

Itxaka:
Looks like the 15.3 image is a moving target, which gets updated from
time to time and older images dropped so we cannot count on it. Instead
use a point releases which should not be dropped and should not be
updated once its released


mudler (approved):

mudler:
Seems to pass build on arm64, merging

PR #1570, opened 2022-07-21
Move to github runners for x86 jobs

Itxaka:
Current self-hosted runners migth not be available for a couple of days


Itxaka:
Already run correctly where the self-hosted runners would have run.

fgiudici (approved):
LGTM, thanks @Itxaka

PR #1594, opened 2022-08-31
bump teal-x86_64.yaml base os

Itxaka:

### DIFF
--- a/packages/cos/collection.yaml
+++ b/packages/cos/collection.yaml
@@ -2,7 +2,7 @@ packages:
   - &cos
     name: "cos"
     category: "system"
-    version: 0.8.14-24
+    version: 0.9.0
     description: "cOS base image, used to build cOS live ISOs"
     brand_name: "cOS"
     labels:
@@ -10,13 +10,10 @@ packages:
   - !!merge <<: *cos
     name: "cos-container"
     description: "cOS container image, used to build cOS derivatives from scratch"
-    version: 0.8.14-25
   - !!merge <<: *cos
     category: "recovery"
     brand_name: "cOS recovery"
     description: "cOS recovery image, used to boot cOS for troubleshooting"
-    version: 0.8.14-24
   - !!merge <<: *cos
     name: "cos-img"
     category: "recovery"
-    version: 0.8.14-24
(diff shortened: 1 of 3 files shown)


davidcassany (commented):
Shouldn't we also bump the builder to leap 15.4? I know it is unlikely to make any difference, but even if it is just for consistency

Itxaka:
yep, sorry was too eager and added the bump and missed bumping cos, bumping arm64, too images etc...

Althougth as far as I can remember I tried to bump it some time ago and failed miserably.

davidcassany:
Weird, looks like we are not passing the flavor to the docker builds. Orange, blue and green are all attempting to build with teal values. This should have not been altered in this PR, feels like an unrelated issue.

Itxaka:
yeah, docker builds are totally broken, not sure if we should just disable them like they were just a couple of months ago.

davidcassany (approved):
Looks good to me

A bit concerned about the failures on the bios installer tests 🤔 I doubt it has direct relation with the changes here though

Given the vast majority of tests are passing I think can disable docker flavors builds and probably focus on failing teal tests in follow up PRs

Itxaka:
> A bit concerned about the failures on the bios installer tests thinking I doubt it has direct relation with the changes here though

yup, gonna wait until is re-run, according to the logs, there was a disconnection so it could not copy the proper files into the VM so it failed.

Gotta wait for a while thought, its run on macos workers are we have none available right now ¬_¬

Itxaka:
ah, I see!

Its just that due to https://github.com/rancher/elemental-cli/pull/287 we now start the custom files from 90, and before we just copied the cloud config files into 99-custom.yaml, so in the test we check for that specific file, thus it wont work.

Will merge this and send a PR to fix the tests.

davidcassany:
> Its just that due to [rancher/elemental-cli#287](https://github.com/rancher/elemental-cli/pull/287) we now start the custom files from 90, and before we just copied the cloud config files into 99-custom.yaml, so in the test we check for that specific file, thus it wont work.

Good catch! Damn, we are overlooking this repository... I wish we had a cleaner relation between `elemental` and  `elemental-toolkit` repos...

Itxaka:
To be honest, on testing just getting transient/new files by specific name is pretty bad as its error prone. We should not get that file directly but just look at the files in /oem and check all of them instead :D

Although a cross project testing like zuul does would be really nice in this case :D

PR #1599, opened 2022-09-01
Set a point release image

Itxaka:
simple version images gets replaced over time, which means that our sha
is not valid anymore for those tags.

Use a point release image instead which dont get replaced so we can use
them over time

### DIFF
--- a/values/teal-arm64.yaml
+++ b/values/teal-arm64.yaml
@@ -1,5 +1,6 @@
 
 # skopeo inspect docker://registry.suse.com/suse/sle-micro-rancher/5.2:latest --override-arch arm64 | jq '.Digest'
+
 image: registry.suse.com/suse/sle-micro-rancher/5.3@sha256:fb4b77662cd848be843a5ed76699e6117ec6d108e5eef22eea6f96c33c451671
 distribution: "opensuse"
 codename: "teal"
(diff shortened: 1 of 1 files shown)


Itxaka:
should fix the current failing arm jobs as they cant find the image sha anymore. It took 3 hours to start failing again :D

davidcassany (approved):
Looks good the me! :+1:

Just to completely understand the issue. Correct me if wrong, the sha usage was required to enforce the use of arm built right? As these references are pointing to multi-arch builds and without the sha it just pulls the x86.

I am wondering if this shouldn't be a luet feature request, I mean, pulling the image for the correct arch based on the arch we are building for.

Itxaka:
> Just to completely understand the issue. Correct me if wrong, the sha usage was required to enforce the use of arm built right? As these references are pointing to multi-arch builds and without the sha it just pulls the x86.

Yes, indeed.

> I am wondering if this shouldn't be a luet feature request, I mean, pulling the image for the correct arch based on the arch we are building for.

Also yes. I think I talked with @mudler about this a looooong time ago when introducing the arm64 flavor but that was an eternity, luet has moved so much that maybe this is not an issue anymore?

also lack of arm64 hardware to do proper testing makes this a pain in the ass.

PR #1600, opened 2022-09-02
split pushing artifacts from signing

Itxaka:
if signing fails, we end up with no artifacts properly pushed so first
publish artifacts and then sign them

This reuses the resigner binary to sign the full tree of artifacts

### DIFF
--- a/.github/build.yaml.gomplate
+++ b/.github/build.yaml.gomplate
@@ -650,6 +650,7 @@
       FLAVOR: {{{ $flavor }}}
       ARCH: {{{ $config.arch }}}
       FINAL_REPO: {{{$config.organization}}}/{{{$config.repository}}}-{{{ $flavor }}}{{{- if ne $config.arch "x86_64"}}}-{{{$config.arch}}}{{{end}}}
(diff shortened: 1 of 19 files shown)


Itxaka:
@davidcassany any chance you can have a look at this? This splits the pushing of the artifacts and signing into its own separate steps, so in case the signing fails we dont get a broken half-published repo.

Also reuses the signer.go util under the .github dir to sign all remote artifacts, signs the missing repository files and adds extra metadata into the signatures with the referenceID of the repo and the team name.

davidcassany:
> @davidcassany any chance you can have a look at this? This splits the pushing of the artifacts and signing into its own separate steps, so in case the signing fails we dont get a broken half-published repo.
> 
> Also reuses the signer.go util under the .github dir to sign all remote artifacts, signs the missing repository files and adds extra metadata into the signatures with the referenceID of the repo and the team name.

Sorry 🙏🏽 I forgot about this PR, looking it now

davidcassany (approved):
Looks good to me :+1:

PR #1604, opened 2022-09-13
Add shim to grub2-efi packages

Itxaka:


packages/grub2/build.yaml
```diff
@@ -27,10 +36,10 @@ steps:
     {{if eq .Values.tool_image_distribution "opensuse"}}
       {{if .Values.arch }}
         {{if eq .Values.arch "x86_64"}}
-- cp /usr/share/grub2/x86_64-efi/grub.efi /EFI/BOOT/bootx64.efi
+- cp /usr/share/efi/x86_64/* /EFI/BOOT/
```
Itxaka, line 39: here we should not care about the naming as elemental-cli already do its stuff

packages/livecd/grub2/build.yaml
```diff
@@ -34,10 +34,16 @@ steps:
 {{if eq .Values.name "grub2-efi-image"}}
   {{if .Values.arch }}
     {{if eq .Values.arch "x86_64"}}
-- mkdir -p /EFI/BOOT && cp /usr/share/grub2/x86_64-efi/grub.efi /EFI/BOOT/bootx64.efi
+- mkdir -p /EFI/BOOT
+- cp /usr/share/efi/x86_64/shim.efi /EFI/BOOT/bootx64.efi
```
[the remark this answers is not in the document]
kkaempf: Please add this knowledge as comments to the code.

packages/grub2/build.yaml
kkaempf: Please add this knowledge as comments to the code.

packages/grub2/build.yaml
```diff
@@ -5,6 +5,15 @@ requires:
 - name: "cos-config-tmpl"
   category: "utils"
   version: ">=0"
+{{if eq .Values.name "grub2-efi-image"}}
+  {{if .Values.tool_image_distribution}}
+    {{if eq .Values.tool_image_distribution "opensuse"}}
```
kkaempf, line 10: Is Teal 'opensuse' ?!

kkaempf (changes requested):

packages/grub2/build.yaml
Itxaka: Packages are tool independent, comment is here because the package change affect us directly.

In any case this still need changes to be backwards compatible so it can't go out like this.

davidcassany (commented):
I am wondering if the shim inclusion into the EFI image should be an opt-in flag or not. May it hit is us somehow always including the shim image?

Itxaka:
> I am wondering if the shim inclusion into the EFI image should be an opt-in flag or not. May it hit is us somehow always including the shim image?

Currently testing this. It should not, as the shim is just a forwarder (In a rough sense) for the grub.efi image, so its only a step in the middle.

packages/grub2/build.yaml
Itxaka: no, this is the tool image which is used across all flavors to provide system-agnostic packages mainly. So for things like the grub artifacts or efi, we only use one source, which is the artifacts from the tool image directly.

We dont really care about the flavor in here as we generate the same package for all flavors.

packages/livecd/grub2/build.yaml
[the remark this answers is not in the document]
Itxaka: done

Itxaka:
Added backwards compatible stuff, so in any case, using your own stuff for grub or using elemental, the result will be the same.

Instead of getting the grub to boot directly you will get shim -> grub, but that works out the same in all cases.

packages/grub2/build.yaml
davidcassany: I kind agree with Klaus this is disturbing... I bet this was done this way because at a time we were not able to `zypper in` into the Teal image the grub packages we needed, the same is happening now with `shim` and `syslinux`.

Itxaka:
now, one missing thing is for teal to support shim packages.

As with teal, we dont install any packages on the base image, so any new version of elemental-cli that uses teal will fail as it wont find the required shim artifacts.

The package has already been asked to get in so it would be a while until it trickles down. At that point, the elemental-cli can be bumped.

packages/grub2/build.yaml
Itxaka: unfortunately this package is only used for raw devices images. Im not even sure if this package should exist or change as its not used anywhere..... we mainly use the grub2-install to generate the grub stuff under elemental-cli so..... I would not change this for now other than adding any extra artifacts that can be added unless we find a use for this.

Itxaka: Is this a blocking issue still @kkaempf ?

kkaempf: No, and it never was. It's just confusing to have a reference to 'opensuse' here.

kkaempf (approved):
ack - as this affects github builds only

kkaempf (commented):
What ? Why ? Nooo 😨 

(stupid GH - this comment was for the Teal 5.2 downgrade)

Itxaka:
> What ? Why ? Nooo fearful
> 
> (stupid GH - this comment was for the Teal 5.2 downgrade)

just checking to see if the same issue arises under 5.2 or its introduced by our packages (Im guessing something else has changed under 5.3 like the NetworkManager stuff)

kkaempf (approved):
https://build.opensuse.org/package/view_file/isv:Rancher:Elemental/systemd-presets-branding-Elemental/50-default-Elemental.preset?expand=1 😝

davidcassany (approved):
LGTM, note that EFI installer tests will fail since the ISO is no longer using EFI.


I'd suggest to merge it if the only failures are the installer tests on EFI.

Itxaka:
> I'd suggest to merge it if the only failures are the installer tests on EFI.

yep, but I would like to modify the test to capture this and check that we are failing on the ISO part. Unfortunately it doesnt spit the error properly so I cant see what the error is to be able to modify the tests properly....Im not sure if I would reach that, getting impatient

PR #1607, opened 2022-09-19
Update grub config for live ISOs

davidcassany:
Update grub configuration to align it with newest elemental-cli

### DIFF
--- a/packages/livecd/grub2/build.yaml
+++ b/packages/livecd/grub2/build.yaml
@@ -8,7 +8,7 @@ prelude:
 - zypper in -y --no-recommends syslinux grub2-i386-pc grub2-x86_64-efi
   {{end}}
   {{if eq .Values.arch "aarch64"}}
-- zypper in -y --no-recommends syslinux-x86_64 grub2-i386-pc grub2-arm64-efi
+- zypper in -y --no-recommends grub2-arm64-efi
   {{end}}
 - zypper cc
 - grub2-mkimage -O i386-pc -o /usr/share/grub2/i386-pc/core.img -p /boot/grub2 -d /usr/share/grub2/i386-pc ext2 iso9660 linux echo configfile search_label search_fs_file search search_fs_uuid ls normal gzio png fat gettext font minicmd gfxterm gfxmenu all_video xfs btrfs lvm luks gcry_rijndael gcry_sha256 gcry_sha512 crypto cryptodisk test true loadenv part_gpt part_msdos biosdisk vga vbe chain boot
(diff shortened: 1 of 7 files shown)


davidcassany:
I made it a draft because it builds the elemental-cli from rancher/elemental-cli#316 which is not merged yet, just for testing.

However it can be already reviewed.

.github/build.yaml.gomplate
```diff
@@ -300,10 +300,8 @@
         with:
           name: build-{{{ $flavor }}}-{{{ $config.arch }}}
           path: build
-      - name: Add build repository
-        run: |
-          export DOCKER_INSTALL=true
-          sudo -E make add_local_repo
+      {{{ tmpl.Exec  "make" "add_local_repo" }}}
+      {{{ tmpl.Exec  "make" "upgrade_deps" }}}
```
davidcassany, line 304: I am wondering why haven't we done something like that before... feels like I am missing some side effect... Couldn't be that simple solving #1515... 🤔

packages/golang/build.yaml
```diff
@@ -6,7 +6,7 @@ requires:
 prelude:
 {{ if .Values.tool_image_distribution }}
 {{if eq .Values.tool_image_distribution "opensuse" }}
-- zypper in -y curl git && zypper install -y -t pattern devel_basis
+- zypper in -y curl git gzip && zypper install -y -t pattern devel_basis
```
davidcassany, line 9: Not sure what happened here, but apparently we need to include `gzip` now... I guess this is another unexpected change on teal base image 😓

packages/livecd/grub2/build.yaml
```diff
@@ -8,7 +8,7 @@ prelude:
 - zypper in -y --no-recommends syslinux grub2-i386-pc grub2-x86_64-efi
   {{end}}
   {{if eq .Values.arch "aarch64"}}
-- zypper in -y --no-recommends syslinux-x86_64 grub2-i386-pc grub2-arm64-efi
+- zypper in -y --no-recommends grub2-arm64-efi
```
davidcassany, line 11: I have not tested the arm64 ISO yet, but I resist to accept we need syslinux-x86_64 and grub2-i386 on arm64.

packages/livecd/grub2/config/grub_live.cfg
```diff
@@ -21,9 +21,9 @@ if [ -f ${font} ];then
 fi
 menuentry "cOS" --class os --unrestricted {
     echo Loading kernel...
-    $linux ($root)/boot/kernel.xz cdroot root=live:CDLABEL=COS_LIVE rd.live.dir=/ rd.live.squashimg=rootfs.squashfs console=tty1 console=ttyS0 rd.cos.disable
+    $linux ($root)/boot/kernel cdroot root=live:CDLABEL=COS_LIVE rd.live.dir=/ rd.live.squashimg=rootfs.squashfs console=tty1 console=ttyS0 rd.cos.disable
     echo Loading initrd...
-    $initrd ($root)/boot/rootfs.xz
+    $initrd ($root)/boot/initrd
```
davidcassany, line 26: This was a long standing TODO we had in elemental-cli. These paths used to be misleading, there was nothing forcing or telling us to use `xz` compression, this is an arbitrary name set by us and we are not forcing `xz` in any way. It is likely to be the default for SLE15-SP4, but, for instance it is not for TW, there dracut defaults to `zst`.

packages/golang/build.yaml
kkaempf: Is that also needed in `builder-image` ? 🤔

packages/livecd/grub2/build.yaml
kkaempf: Huh ? Isn't the syslinux rpm just named `syslinux` ? Where's `syslinux-x86_64` coming from ?

.github/build.yaml.gomplate
Itxaka: this bites us in the ass down the line. I already did this some time ago, and my thinking was exactly the same LOL

I think it was something to do with cross arch build in which the deps were upgraded with wrong arch packages? Something like that, maybe its now fixed because luet knows better about arches?

packages/livecd/grub2/build.yaml
Itxaka: cant build an hybrid iso from an arm64 os then no? arent the needed files in the syslinux package and i386 packages, thus we also need them on x86_64? I mean, we are mixing x86_64 with 386 and the only reason is that we need files from those package AFAIk? Otherwise if we can drop it from here, we can drop it from the x86_64 package. 

We need to make sure that x86 and arm64 packages have exactly the same files if we support both arches

davidcassany:
Closing in favor of #1613, equivalent changes are included there.

PR #1613, opened 2022-10-04
Adapt to the latest elemental-cli

davidcassany:


.github/build.yaml.gomplate
```diff
@@ -300,10 +300,7 @@
         with:
           name: build-{{{ $flavor }}}-{{{ $config.arch }}}
           path: build
-      - name: Add build repository
-        run: |
-          export DOCKER_INSTALL=true
-          sudo -E make add_local_repo
+      {{{ tmpl.Exec  "make" "add_local_repo" }}}
```
davidcassany, line 303: I am still tempted to add here an additional step to upgrade repositories so we use the built artifacts to build the iso. I know this will break cross arch ISO builds, however I am wondering if it wouldn't be easier to not run cross arch builds and relay on pure arch specific CI pipes?

.github/build.yaml.gomplate
```diff
@@ -575,7 +572,7 @@
   {{{ $config := (datasource "config") }}}
   {{{ $flavor := index . "flavor" }}}
   {{{ $subset := index . "subset" }}}
-  {{{- range slice "efi" "bios" }}}
+  {{{- range slice "bios" }}}
```
Itxaka, line 575: isnt this the other way around??

packages/livecd/grub2/build.yaml
```diff
@@ -6,13 +6,13 @@ prelude:
 {{ if .Values.arch }}
   {{if eq .Values.arch "x86_64"}}
 - zypper in -y --no-recommends syslinux grub2-i386-pc grub2-x86_64-efi shim
+- grub2-mkimage -O i386-pc -o /usr/share/grub2/i386-pc/core.img -p /boot/grub2 -d /usr/share/grub2/i386-pc ext2 iso9660 linux echo configfile search_label search_fs_file search search_fs_uuid ls normal gzio png fat gettext font minicmd gfxterm gfxmenu all_video xfs btrfs lvm luks gcry_rijndael gcry_sha256 gcry_sha512 crypto cryptodisk test true loadenv part_gpt part_msdos biosdisk vga vbe chain boot
+- cat /usr/share/grub2/i386-pc/cdboot.img /usr/share/grub2/i386-pc/core.img > /usr/share/grub2/i386-pc/eltorito.img
   {{end}}
   {{if eq .Values.arch "aarch64"}}
-- zypper in -y --no-recommends syslinux-x86_64 grub2-i386-pc grub2-arm64-efi shim
+- zypper in -y --no-recommends grub2-arm64-efi shim
```
davidcassany, line 13: to me it feels all this x86 binaries here are just legacy...

.github/build.yaml.gomplate
davidcassany: this what I wanted to ask you, did we already shitf everything to EFI?

Itxaka: I think so, bios deprecated, efi by default, we leave the option to have bios in there but not support it no?

davidcassany: yes, but I mean, are our VMs in the CI for all tests booting in EFI?

.github/build.yaml.gomplate
Itxaka: The problem with pure arch specific pipes is that we then add extra jobs for anything that requires arm64 and we only have one worker ready, which means that we increase the time for arm64 jobs. Although currently there is not that much movement on arm64 so it may be easier to just dont do cross arch anymore and avoid this stupid problem of deps which may be more important.

Anyway, this is just a speed VS deps issue. I would go with fixing the deps issue, which is more problematic than having to way half an hour more IMO. If in the future we get more heavy development in the toolkit and it requires more arm64 specific development, we can also temporarily increase the arm64 workers, while we cannot fix the dep problems easily.

.github/build.yaml.gomplate
Itxaka: good point... pretty sure packer isnt even ready, at least the x86 part. aarch64 is efi as that is the only possible way but not the x86.

And the same for the tests, vagrant for aarch64 is efi ready as there is no bios on aaarch64, but x86 it just uses virtuabox with defaults, so no, its bios.

The change its "easy" but tweaking is gonna be a nightmare. In any case, I would go full EFI, even if it breaks the pipeline. Then we can rework the pipeline to work with that, drop uneeded stuff (bye bye vbox packer) etc..

.github/build.yaml.gomplate
davidcassany: Ok just revisited and I think that adding the proper built arch in `add_local_repo` should do the trick. The problem is that the luet interface does not allow to set repo architecture form the command line, hence the yaml file for the repository should be modified manually. I guess should test this approach by a manual hack on the `add_local_repo`  target and if it works then code the correspondent arch flag in `luet add repo` command.

.github/build.yaml.gomplate
davidcassany: oh you are brave man :wink: 

You convinced me, lets assume a broken pipe for now and move to EFI on follow up PRs until it is fixed.

Itxaka: Already testing going the QEMU only path for everything with EFI: https://github.com/rancher/elemental-toolkit/pull/1624 ;)

Itxaka:
weird, still failed..

Itxaka:
FYI I canceled the workflow after the packer box failed to boot as that indicates that the iso is wrong, but the workflow keeps stuck somehow.

davidcassany:
> FYI I canceled the workflow after the packer box failed to boot as that indicates that the iso is wrong, but the workflow keeps stuck somehow.

This is not accurate, packer will fail for x86, as it uses the default firmware for qemu, I bet this is BIOS. Only the efi installer tests had a chance to pass. No other test has a chance to pass in this PR for x86.

Itxaka:
argh sorry, I retriggered the failed jobs then. Was too eager to get an iso in order to test building locally

Itxaka (approved):
lets goooo

PR #1619, opened 2022-10-11
Drop ami publishing

Itxaka:
Not needed anymore, all images were clean up.

Signed-off-by: itxaka <igarcia@suse.com>


kkaempf (approved):
Dropping code is always good ;-)

Itxaka:
Can only see the effect on release so merging already

PR #1620, opened 2022-10-11
Drop uneeded stuff

Itxaka:
Drops several uneeded jobs that are not needed anymore plus the ghrunner docker image which we no longer use

Signed-off-by: itxaka <igarcia@suse.com>


kkaempf (approved):
More removals, more good

davidcassany (approved):
lgtm :+1:

PR #1628, opened 2022-10-18
signatures: protect against broken tree

Itxaka:
If the repo tree is broken we got no protection and panic miserably. Instead just print the erro and exit.


frelon (approved):
LGTM

PR #1637, opened 2022-11-03
Use go 1.18 in tests

davidcassany:

### DIFF
--- a/.github/build.yaml.gomplate
+++ b/.github/build.yaml.gomplate
@@ -92,7 +92,7 @@
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-            go-version: '1.17'
+            go-version: '1.18'
   {{{- end }}}
   {{{- if $config.local_runner }}}
       - name: Install make
@@ -440,7 +440,7 @@
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-            go-version: '1.17'
+            go-version: '1.18'
       - uses: actions/checkout@v2
       - name: Download vagrant box
         uses: actions/download-artifact@v2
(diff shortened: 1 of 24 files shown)


Itxaka (approved):

davidcassany:
I am not entirely sure this is the best approach but seams to work. Essentially I could not find why in ginkgo and gomega docs they stick with the `go get` instead of `go install`. I could see some back and forth in their docs around the two recently due to some people having issues with go1.18, but I could not see any reasoning on why they still stick with `go get` for omega... 🤷🏽‍♂️

Itxaka:
> I am not entirely sure this is the best approach but seams to work. Essentially I could not find why in ginkgo and gomega docs they stick with the `go get` instead of `go install`. I could see some back and forth in their docs around the two recently due to some people having issues with go1.18, but I could not see any reasoning on why they still stick with `go get` for omega... 🤷🏽‍♂️

I think `go install` is meant to install binaries, while `go get` just downloads the given package. 

So the main difference is that `go install` will look for a `main` module to compile it to binary while `go get` just ignores if there is a `main` module. Thus gomega which is just a matcher does not contain a `main` module so it can be installed by `go install`

Now, I dont understand why when you install ginkgo, gomega is not build as part of it if it has it as a dep, nor I want to understand go deps anymore. Its all just a mess. We should probably use the `go run github.com/onsi/ginkgo/v2/ginkgo` from the tests dir so it uses the versions in the go.mod, although Im not even sure that will fix it.

davidcassany:
> I think `go install` is meant to install binaries, while `go get` just downloads the given package.

Yes, understand this is the case for go1.18. Prior that I think `go get` also used to build binaries, starting from g1.18 `go get` is no longer building the binaries, this is what I understood. Probably we can drop the `go install` for gomega... 

> We should probably use the go run github.com/onsi/ginkgo/v2/ginkgo from the tests dir so it uses the versions in the go.mod, although Im not even sure that will fix it.

I think we are already doing that 🤔

davidcassany:
I think I will run a separate PR to try to better fix it, if not we can just merge this as it seams to work for now and it is a simple change.

davidcassany:
closing in favor of #1638

davidcassany:
closing in favor of #1638

PR #1640, opened 2022-11-04
Centralize go installation in CI jobs

davidcassany:

### DIFF
--- a/.github/build.yaml.gomplate
+++ b/.github/build.yaml.gomplate
@@ -45,7 +45,6 @@
       - uses: actions/checkout@v3
       - run: |
           git fetch --prune --unshallow
-  {{{ tmpl.Exec "runner_cleanup" }}}
       - name: Install CI plugins
         run: |
             sudo cp -rfv .github/plugins/* /usr/bin/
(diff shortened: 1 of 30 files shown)


.github/build.yaml.gomplate
```diff
@@ -86,8 +86,9 @@
           timeout_minutes: 5
           max_attempts: 3
           command: |
-            curl -L https://golang.org/dl/go1.17.5.linux-{{{ $arch }}}.tar.gz -o go1.17.5.linux-{{{ $arch }}}.tar.gz
-            sudo rm -rf /usr/local/go && sudo tar -C /usr/local -xzf go1.17.5.linux-{{{ $arch }}}.tar.gz
+            curl -L https://golang.org/dl/go1.18.8.linux-{{{ $arch }}}.tar.gz -o go1.18.8.linux-{{{ $arch }}}.tar.gz
```
Itxaka, line 89: we should probably create a template that returns the go version so we can use it everywhere and we we want to bump up the go version we can do in a single place before running the pipelines.sh otherwise we gonna keep changing the version manually :/

davidcassany: Yes I also thought about that, but first I wanted to check if my approach is actually working. So far I already found an inconsistency at `.github/workflows/build-main-teal-arm64.yaml`, as it runs on ubuntu-latest but it renders the steps for self hosted runners... the build gomplate is not well prepared to distinguish this sort of cases were we run some cross-arch stuff... So arm with this PR is definitely broken.

davidcassany: Ok, now I think it should be fine. I'd merge this and check if the publish job is now fixed. If so a follow up PR to template the golang version should be easy and with little chances to break again something.

Itxaka (approved):

PR #1644, opened 2022-11-15
Grub2 config rework

Itxaka:
- Merge config files
 - Merge custome files
 - Store everything in STATE dir
 - Store everything under grub2 dir
 - Update docs to clarify grub2 config location
 - Bump required packages
 - Load custom/env file from the state part directly, no need for search

cc @davidcassany 

### DIFF
--- a/docs/content/en/docs/Customizing/configure_grub.md
+++ b/docs/content/en/docs/Customizing/configure_grub.md
@@ -7,7 +7,7 @@ description: >
   GRUB 2 Configuration
 ---
 
-COS is set to deploy a persistent `grub.cfg` into the `COS_RECOVERY` partition during
+COS is set to deploy a persistent `grub.cfg` into the `COS_STATE` partition during
 the system installation or image creation. COS grub configuration
 includes three menu entries: first for the main OS system, second for the
 fallback OS system and a third for the recovery OS.
@@ -28,7 +28,7 @@ menuentry "cOS" --id cos {
 ```
 
 {{% alert title="Kernel parameters" %}}
-The kernel parameters are not part of the persistent `grub.cfg` file stored in `COS_RECOVERY` partition. 
+The kernel parameters are not part of the persistent `grub.cfg` file stored in `COS_STATE` partition. 
 Kernel parameters are sourced from the loop device of the OS image to boot. This is mainly to
 keep kernel parameters consistent across different potential OS images or
 system upgrades. 
@@ -90,12 +90,15 @@ Use `grub2-editenv` command line utility to define the desired values.
 For instance use the following command to reboot to recovery system only once:
 
 ```bash
-> grub2-editenv /run/cos-state/grub2/grub_oem_env set next_entry=recovery
+> grub2-editenv /oem/grub_oem_env set next_entry=recovery
 ```
+{{% alert title="Note" %}}
+The examples below make of the `COS_OEM` device.
+{{% /alert %}}
 
 ### Default boot entry
 
-The default grub configuration loads the `/grubenv` of any available device
+The default grub configuration loads the `/grub_oem_env` from the COS_OEM partition
 and evaluates on `next_entry` variable and `saved_entry` variable. By default
 none is set.
 
@@ -107,18 +110,16 @@ boot. If `next_entry` variable is set this is only being used once, GRUB2 will
 unset it after reading it for the first time. This is helpful to define the menu entry
 to reboot to without having to make any permanent config change.
 
-Use `grub2-editenv` command line utility to define desired values.
-
 For instance use the following command to reboot to recovery system only once:
 
 ```bash
-> grub2-editenv /run/cos-state/grub2/grub_oem_env set next_entry=recovery
+> grub2-editenv /oem/grub_oem_env set next_entry=recovery
 ```
 
 Or to set the default entry to `fallback` system:
 
 ```bash
-> grub2-editenv /run/cos-state/grub2/grub_oem_env set saved_entry=fallback
+> grub2-editenv /oem/grub_oem_env set saved_entry=fallback
 ```
 
 ## Boot menu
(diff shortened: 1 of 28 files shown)


docs/content/en/docs/Customizing/configure_grub.md
```diff
@@ -176,7 +171,7 @@ By default Elemental boots into active, and if there are failures will boot into
 It is possible to override the default fallback logic by setting `default_fallback` as grub environment, consider for example:
 
 ```bash
-> grub2-editenv /oem/grubenv set default_fallback="2 0 1"
+> grub2-editenv /run/cos-state/grub2/grub_oem_env set default_fallback="2 0 1"
```
davidcassany, line 174: Again, I would limit to document variable meaning and do no instruct to directly change it. The we could have a docs section to modify grubenv vars as an after-install/upgrade hook.

packages/cloud-config/README-cloud-config-boot-assessment.md
```diff
@@ -10,22 +10,22 @@ The boot assessment process works in the following way:
...
+- `/grub2/boot_assessment` - contains the GRUB env sentinel variables
+- `/grub2/grub_boot_assessment` - contains the GRUB logic for booting into fallback
 
-Note: The extra GRUB logic is installed and sourced into `/grubcustom` inside `COS_STATE`.
+Note: The extra GRUB logic is installed and sourced into `/grub2/grubcustom` inside `COS_STATE`.
 
 When a boot failure is detected and the fallback is automatically selected, it will be created a `/run/cos/upgrade_failure` sentinel file during the boot process, which is accessible under the `boot` cloud-init stage.
 
-To enable boot assessment besides upgrades, `enable_boot_assessment_always` can be set to `yes` in the grub environment ( for example in `/oem/grubenv` ).
+To enable boot assessment besides upgrades, `enable_boot_assessment_always` can be set to `yes` in the grub environment ( for example in `/run/cos-state/grub2/grubenv` ).
```
davidcassany, line 20: I think `/run/cos-state/grub2` is wrong, should be `/run/initramfs/cos-state`, however this is a runtime path, we should not relay on this on our docs. Also there is a long standing TODO in elemental-cli to change this default to be aligned with other paths under `/run/cos/....`

The same I would move the enablement/disable docs to an upgrade/install hook.

packages/grub2/config/grub.cfg.tmpl
```diff
@@ -1,27 +1,15 @@
 set timeout=10
 
-# Load custom env file
-set env_file="/grubenv"
-search --no-floppy --file --set=env_blk "${env_file}"
+# Load env file if file exists and size its not 0
+search --no-floppy --label --set=customenv @STATE_LABEL@
```
davidcassany, line 4: Why no call it `state_blk` instead of `customenv`? We can also reuse this variable in cos and fallback menus. So we also skip the search in there.

davidcassany (commented):
What about also changing the menus like, I think this should work

```
menuentry "${display_name}" --id cos {
  set img=/cOS/active.img
  set label=@ACTIVE_LABEL@
  loopback loop0 /$img
  source (loop0)/etc/cos/bootargs.cfg
  linux (loop0)$kernel $kernelcmd ${extra_cmdline} ${extra_active_cmdline}
  initrd (loop0)$initramfs
}
```

packages/grub2/config/grub.cfg.tmpl
davidcassany: I am wondering if at this stage $root is already set, probably it is essentially set to the current blk device. If so this is already state and there is no need to re-search it again (it was already searched to read the cfg file). So probably we can even remove this one and relay on $root.

Itxaka: Well I had variable issues with reusing vars in the grub.cfg file when I was testing some other things.

First I want to check that this passes the tests, then I will have a look at this!

docs/content/en/docs/Customizing/configure_grub.md
```diff
@@ -117,13 +112,13 @@ Use `grub2-editenv` command line utility to define desired values.
 For instance use the following command to reboot to recovery system only once:
 
 ```bash
-> grub2-editenv /oem/grubenv set next_entry=recovery
+> grub2-editenv /run/cos-state/grub2/grub_oem_env set next_entry=recovery
```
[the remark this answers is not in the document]
Itxaka: whole docs need changing after this to rely into the hooks instead of this.

packages/grub2/config/grub.cfg.tmpl
davidcassany: According to the grub2 [manual](https://www.gnu.org/software/grub/manual/grub/grub.html#config_005fdirectory) we should have already properly set the `config_directory` after loading the config file. So files could be checked under this directory and event avoid the STATE search. This is relevant as it allows us to choose the STATE label at install time without having to touch the in image `grub.cfg`.

Itxaka:
> I like it +1 Probably we can clean it a little further and I think documentation requires a separate follow up PR for wider changes.
> 
> What about also changing the menus like, I think this should work. We where setting root but not using it anywhere, which makes me think we probably don't event need the former search of the state label.
> 
> ```
> menuentry "${display_name}" --id cos {
>   set img=/cOS/active.img
>   set label=@ACTIVE_LABEL@
>   loopback loop0 /$img
>   source (loop0)/etc/cos/bootargs.cfg
>   linux (loop0)$kernel $kernelcmd ${extra_cmdline} ${extra_active_cmdline}
>   initrd (loop0)$initramfs
> }
> ```

yep. Not sure why we use the root in there, pretty sure it was used for something before. I mean `set root=($root)` ??? plus then we override root in the kernel args so Im not even sure if at this point is that even needed....

packages/grub2/config/grub.cfg.tmpl
Itxaka: oh wow, that is handy...so then we only need to change the root if we are on recovery?? Pretty nice!

Itxaka:
until discussion, this is not clear so closing.

Issue #1650, opened 2022-11-20
Creating embedded images


wanshaoyuan:
```
docker build -t derivative:latest .
```
Error:

```
failed to acquire exclusive app lock: open /var/lock/luet.lock: no such file or directory
```

add
```
ENV LUET_NOLOCK=true
```

frelon:
@wanshaoyuan Could you share the Dockerfile used?

PR #1651, opened 2022-11-21
Add LUET_NOLOCK for example Dockerfile

frelon:
As pointed out in #1650  


davidcassany (approved):
Looks good :+1:

PR #1653, opened 2022-11-23
Update documentation

frelon:
Update docs to use NetworkManager and `nmcli` instead of `wicked`

Also remove mentions of AMI:s in releases since those are dropped in #1619 


kkaempf (approved):
lgtm

kkaempf:
@juadk @ldevulder - do we have tests for such cloud-config scenarios ?

PR #1654, opened 2022-11-24
Fixes cos-setup systemd unit paths

davidcassany:
`/lib/systemd/system` is not the path used for most major distributions. In fact Fedora and Ubuntu ship their `/lib` as a symlink to `/usr/lib`, hence the real path is `/usr/lib/systemd/system`. In openSUSE Leap there is no such link and both exist `/usr/lib/` and `/lib`, being `/usr/lib` the default path for systemd. This also applies for Teal.

From systemd PoV `/lib/system/systemd` is a valid path, this why elemental-toolkit does not break. However I believe that if Ubuntu, Fedora and openSUSE they all use a different and equal path, we should use that.

### DIFF
--- a/packages/immutable-rootfs/30cos-immutable-rootfs/module-setup.sh
+++ b/packages/immutable-rootfs/30cos-immutable-rootfs/module-setup.sh
@@ -22,7 +22,6 @@ install() {
     declare moddir=${moddir}
     declare systemdutildir=${systemdutildir}
     declare systemdsystemunitdir=${systemdsystemunitdir}
-    declare initdir="${initdir}"
 
     inst_multiple \
         mount mountpoint sort rmdir findmnt rsync cut
--- a/packages/cos-setup/02-cos-setup-initramfs.conf
+++ b/packages/cos-setup/02-cos-setup-initramfs.conf
@@ -1,4 +1,4 @@
-install_items+=" /lib/systemd/system/cos-setup-initramfs.service /etc/systemd/system/initrd.target.requires/cos-setup-initramfs.service "
-install_items+=" /lib/systemd/system/cos-setup-rootfs.service /etc/systemd/system/initrd-fs.target.requires/cos-setup-rootfs.service "
+install_items+=" /usr/lib/systemd/system/cos-setup-initramfs.service /etc/systemd/system/initrd.target.requires/cos-setup-initramfs.service "
+install_items+=" /usr/lib/systemd/system/cos-setup-rootfs.service /etc/systemd/system/initrd-fs.target.requires/cos-setup-rootfs.service "
 install_items+=" /etc/hosts "
 add_dracutmodules+=" network "
--- a/packages/cos-setup/build.yaml
+++ b/packages/cos-setup/build.yaml
@@ -3,11 +3,11 @@ requires:
   category: "distro"
   version: ">=0"
 steps:
-- mkdir -p /lib/systemd/system
+- mkdir -p /usr/lib/systemd/system
 - mkdir -p /etc/dracut.conf.d
 - cp 02-cos-setup-initramfs.conf /etc/dracut.conf.d
-- cp -rfv *.service /lib/systemd/system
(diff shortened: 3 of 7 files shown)


davidcassany:
This fixes a current bug I casually saw in Staging builds logs, for some reason dracut does not fail and only reports the error. This has been there for a while it is actually also present in our Stable 😞 

```
[  113s] [0m[91mdracut-install: ERROR: installing '/lib/systemd/system/cos-setup-initramfs.service'
[  113s] [0m[91mdracut-install: ERROR: installing '/lib/systemd/system/cos-setup-rootfs.service'
[  113s] [0m[91mdracut: FAILED: /usr/lib/dracut/dracut-install -D /var/tmp/dracut.j51Hd9/initramfs -a /lib/systemd/system/cos-setup-initramfs.service /etc/systemd/system/initrd.target.requires/cos-setup-initramfs.service /lib/systemd/system/cos-setup-rootfs.service /etc/systemd/system/initrd-fs.target.requires/cos-setup-rootfs.service /etc/hosts /usr/lib/udev/rules.d/54-before-scsi-sg3_id.rules /etc/ssl/ca-bundle.pem /etc/ssl/certs /etc/ssl/ct_log_list.cnf /etc/ssl/ct_log_list.cnf.dist /etc/ssl/engdef.d /etc/ssl/engines.d /etc/ssl/openssl.cnf /etc/ssl/private /usr/lib/udev/rules.d/11-dm-parts.rules /usr/lib/tmpfiles.d/multipath.conf
```

This happens because SLE Micro installs unit files in `/usr/lib/systemd/system` by default, the same is valid for Fedora and Ubuntu, but there this is not a huge issue because these two ship `/lib` as a symlink to `/usr/lib`, hence the path is still valid.

I am currently trying to figure out how it is it never popped up, to my understanding this issue cloud-init execution at boot. Still digging, but in any case, this fix applies.

Itxaka:
but then our derivatives based of teal packages still run those services no? We would have seen issues otherwise??? Im confused!

packages/immutable-rootfs/30cos-immutable-rootfs/module-setup.sh
```diff
@@ -22,7 +22,6 @@ install() {
     declare moddir=${moddir}
     declare systemdutildir=${systemdutildir}
     declare systemdsystemunitdir=${systemdsystemunitdir}
-    declare initdir="${initdir}"
```
Itxaka, line 25: THANK YOU. This kept printing an error if Im not mistaken and was fedup of seeing it

Itxaka (approved):

davidcassany:
> but then our derivatives based of teal packages still run those services no? We would have seen issues otherwise??? Im confused!

me too, now I am trying to understand what the hell happened here and why we haven't seen that before

packages/immutable-rootfs/30cos-immutable-rootfs/module-setup.sh
davidcassany: Yes indeed, you are right, this was a pretty annoying warning

davidcassany:
@Itxaka mistery solved, we were just lucky not to face any issue, the key is in the dracut.conf files:

```
install_items+=" /lib/systemd/system/cos-setup-initramfs.service /etc/systemd/system/initrd.target.requires/cos-setup-initramfs.service "
```

This install_items calls includes the symlink and the linked file. Apparently dracut when installing a symlink installs the symlink itself and whatever the link was pointing to,  so despite the wrong path and the failure installing the unit file with a wrong path the unit was still there because the activation link (created with `systemclt enable`) was correct.

Now I am wondering if removing the `/usr/lib/systemd/system/cos-setup-initramfs.service` in the install items list or not... 🤔 Less configuration tends to be better, but also having explicit inclusion of the unit and its activation link makes it easier to understand. I think I'll gonna leave it as it is in this PR for now, including both, the file and the symlink.

PR #1657, opened 2022-11-25
Bump yip and separate it into its own package

Itxaka:
Versioning has changed upstream to include a v in front of the version, so its no longer possible to have it on the same template as the others


frelon (approved):
Nice :+1:

PR #1659, opened 2022-11-26
Set serial baudrate to 115200

kkaempf:
This is the standard rate on Raspberry Pi.

Signed-off-by: Klaus Kämpf <kkaempf@suse.de>

### DIFF
--- a/packages/livecd/grub2/config/grub_live.cfg
+++ b/packages/livecd/grub2/config/grub_live.cfg
@@ -21,7 +21,7 @@ if [ -f ${font} ];then
 fi
 menuentry "cOS" --class os --unrestricted {
     echo Loading kernel...
-    $linux ($root)/boot/kernel cdroot root=live:CDLABEL=COS_LIVE rd.live.dir=/ rd.live.squashimg=rootfs.squashfs console=tty1 console=ttyS0 rd.cos.disable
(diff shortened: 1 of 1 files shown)


Itxaka:
Have you tried adding the value `tty` in the registration with those extra values? Im not sure if it works but we may be able to tweak it to add the baudrate or any random options to the tty line in grub.

Itxaka:
wait this is the toolkit :D

kkaempf:
this is required when booting from the ~ISO~image on RPi with serial console. RPi's bootloader, u-boot, and grub all use 115200 but the kernel doesn't (🤦🏻)

Itxaka:
if you want this to be shipped yu need to bump the versions :D

Itxaka:
@kkaempf if the packages are not bumped then they are not pushed and not available to install, so your changes are lost until the packages version is changed :)

Itxaka:
also, not tested due to no bumped version

kkaempf:
> Sorry, you need to bump the grub package that you changed, otherwise ci will pick the old version

🤔 
We should document this somewhere ...

Itxaka:
> > Sorry, you need to bump the grub package that you changed, otherwise ci will pick the old version
> 
> thinking We should document this somewhere ...

well it kind of makes sense, and pretty sure its documented on the luet docs :D

kkaempf:
How about [this](https://github.com/rancher/elemental-toolkit/pull/1659/commits/df58f25234f33981e0a1d12e2a94e6abc4a71053) ?

Itxaka (approved):

kkaempf:
closing for a better approach 🤞🏻

PR #1668, opened 2022-11-29
[ci] compress raw image before upload

Itxaka:
Instead of transferring 2Gb just transfer around 500Mb. affect the raw image creation, raw image test and release steps


Itxaka:
should also fix the release not being able to upload the raw image as its over the size limits (2Gb)

kkaempf (approved):
👍🏻

PR #1692, opened 2022-12-28
Move elemental-cli tracking to release versions

### LINKED ISSUE #1691: Move elmental-cli package to track versions again
Once upon a time we were developing elemental-cli with a lot of changes that affected the toolkit very much and we needed to have an updated package almost daily without having to cut new releases of elemental-cli. Hence, the council of wise men reunited and accorded to move the elemental-cli into a non-versioned package tracking so we would be able to easily test new packages on the toolkit.

But what was once a pot of boiling changes and new features has now turn into a very robust cli with less breaking changes and more stable improvements. Thus the council of wise men have to reunite once again and decide if the tracking should be back to version of elemental-cli.

Itxaka:
Now that its stable, we should just build stable releases

fixes #1691 

requires sign off by @davidcassany as I know he likes to try new stuff on elemental-cli

also moves it from a collection to a package. we are not enabling fips on it for now, so it doesnt make sense to have a collection...yet.


davidcassany (approved):
Looks good to me,

I wonder how to test elemental-cli changes in that case though 🤔 In any case I agree this is nicer and for testing we could always switch version by commit hash within the test PR, not very nice, but good enough if it is just for testing some sensitive elemental-cli change.

davidcassany:
> I guess by changing the tracking back to hash? Indeed its a pita

Yest that was my idea, it shouldn't be that hard neither happen that often. I agree we must move to releases though.  Not sure if we could find a way to relay on `git.hash` if provided instead of version. That would make it easier in PRs to give it a test.

PR #1693, opened 2022-12-29
Extreme cleanup of workers

Itxaka:
Again running out of space on the workers, maximize the cleanup as much as possible to remove things not needed

### DIFF
--- a/.github/build.yaml.gomplate
+++ b/.github/build.yaml.gomplate
@@ -66,8 +66,29 @@
           sudo rm -Rf root oem efi || true
           docker system prune -f -a --volumes || true
   {{{- else }}}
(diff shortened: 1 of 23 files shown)


kkaempf (approved):
slimmer workers are always welcome !

Itxaka:
I dont think we can do much more than this tbh, this frees about 30Gb of space which is the max we can get. There is a hacky workaround to get another 9gb free by creating an lvm volume with the root disk + temp disk but that makes the whole thing slower. This should get better once we start dropping packages from the toolkit as ther will be not that much space needed for a full rebuild.

Itxaka:
I added the output of listing the snapd packages so we can see in the future if packages are added that way as it seems to be the future of ubuntu

PR #1698, opened 2023-01-11
Split sign into build and run

Itxaka:
Build the signer without root privs so it works on any runner that installs go as the current user.

Also run the signer as root in a different target so we can build it once without privs and then run the binary with root as its needed due to luet chowning files


fgiudici (approved):
looks good here, thanks!

PR #1718, opened 2023-01-26
Use findmnt in update-grub-hook

frelon:
Fixes rancher/elemental#632

### DIFF
--- a/packages/cloud-config/oem/01_upgrade_grub_hook.yaml
+++ b/packages/cloud-config/oem/01_upgrade_grub_hook.yaml
@@ -13,10 +13,7 @@ stages:
       name: "Set grub env variables"
       commands:
       - |
-        statemnt="/run/cos/state"
-        if [ ! -f "${statemnt}/grub2/grub.cfg" ]; then
-          statemnt="/run/initramfs/cos-state"
-        fi
+        statemnt="$(findmnt /dev/disk/by-label/COS_STATE -n -o target | head -n1)"
 
         grub2-editenv ${statemnt}/grub_oem_env set state_label=COS_STATE
         grub2-editenv ${statemnt}/grub_oem_env set active_label=COS_ACTIVE


packages/cloud-config/oem/01_upgrade_grub_hook.yaml
```diff
@@ -13,10 +13,7 @@ stages:
       name: "Set grub env variables"
       commands:
       - |
-        statemnt="/run/cos/state"
-        if [ ! -f "${statemnt}/grub2/grub.cfg" ]; then
-          statemnt="/run/initramfs/cos-state"
-        fi
+        statemnt="$(findmnt /dev/disk/by-label/COS_STATE -n -o target | head -n1)"
```
davidcassany, line 16: Honestly, it doesn't convince me... I know this fixes the current issue, but I'd like to avoid having to refer to device literals and mount points here.

My idea was more aligned with setting and documenting hooks to be capable finding partitions at known places (arbitrary static paths):

```
/run/cos/recovery
/run/cos/state
/run/cos/workingtree  # current image being deployed
```
I know we are not yet there and this sort of changes are tricky for backward compatibility but this is how I'd like to see it at some point in time.

Also, in very short term I'd like to code the above TODO and set `elemental install` to create these labels and also tweak this script to get the labels from the state file. In that case we would be back to the same former issue. How to get the path of the `state.yaml` file without knowing the filesystem label?

I am also curious to understand why the `/host/run` to `/run` bind mount is not functional... :thinking: could it be related to the after-upgrade-chroot? probably it gets unmounted there, it should appear in the logs.

davidcassany (commented):
Not convinced, however I am fine by having a fill the gap solution if we don't see a clear better path. But probably it would be better to have such a fix within the elemental repository and not in elemental-toolkit? this is pretty specific to the suc-upgrade and the weird setup it run elemental

packages/cloud-config/oem/01_upgrade_grub_hook.yaml
frelon: Yep, it's a workaround until rancher/elemental-cli#412 is solved (or this hook is moved into the cli)

frelon:
Closing this in favor of rancher/elemental-cli#414

PR #1719, opened 2023-01-27
Remove after-install from upgrade-grub-hooks

frelon:
The labels are now set during install/upgrade/reset by elemental-cli in rancher/elemental-cli#414.


davidcassany (commented):
I am not sure we should remove the upgrade hook.

I don't think it is needed for elemental-teal, but for upgrades on system where the elemental-cli of the previous version runs the upgrade process the `after-upgrade-*` steps should be kept. So rather than fully deleting it I'd probably just reduce it to the upgrade steps.

Together with that we can drop the `Requires:       elemental-upgrade-hooks = %{version}-%{release}` from spec, so the `elemental-toolkit` RPM does not strictly require it and it won't be included in Elemental-teal.

frelon:
That's a good point, I will remove the `after-install` stage and move it to upgrade instead.

PR #1724, opened 2023-02-06
Fix immutable rootfs setup

davidcassany:
This commit ensures variables defined in loops are also declared as local in each function. Otherwise some variables are crossing from one method to another.

The issue poped up adding multiple device mountpoints within the kernel command line.


frelon (approved):
LGTM

PR #1731, opened 2023-02-16
fix network stage execution constraints

davidcassany:
This makes sure the network stage is actually executed when `network-online.target` is either in failed or succeeded status, but not before.


fgiudici (approved):
Looks good here

PR #1734, opened 2023-02-17
Fix bootargs.cfg

davidcassany:
This commit ensures empty variables are evaluated as empty strings, as it was noted unquoted variables in conditional tests were not evaluated as empty strings leading to weird behavior on special corner cases.

### DIFF
--- a/packages/grub2/config/bootargs.cfg
+++ b/packages/grub2/config/bootargs.cfg
@@ -1,6 +1,6 @@
 # TODO we could sanity check $partlabel is set here so we can error out before even attempting to boot
 set kernel=/boot/vmlinuz
-if [ ${label} == ${system_label} ]; then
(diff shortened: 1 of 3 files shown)


packages/grub2/config/bootargs.cfg
```diff
@@ -1,6 +1,6 @@
 # TODO we could sanity check $partlabel is set here so we can error out before even attempting to boot
 set kernel=/boot/vmlinuz
-if [ ${label} == ${system_label} ]; then
+if [ "${label}" == "${system_label}" ]; then
```
davidcassany, line 3: I got an unexpected behavior when system_label was an empty string, the comparison was always returning true. Using quotes just fixed it.

frelon (approved):
LGTM

PR #1735, opened 2023-02-28
Deprecate system/cos-config

frelon:
The package is used as a base for cloud-config and grub2, use distro/tools instead and bump versions.


frelon:
@davidcassany Yep, forgot to mark this as draft/wip. Thanks for the pointers!

packages/grub2/build.yaml
```diff
@@ -28,8 +27,7 @@ steps:
 
 {{if eq .Values.name "grub2-efi-image"}}
 - mkdir -p /EFI/BOOT
-- cp config/grub_efi.cfg.tmpl /EFI/BOOT/grub.cfg.tmpl
-- cos-tmpl-render /EFI/BOOT/grub.cfg.tmpl
+- cp config/grub_efi.cfg /EFI/BOOT/grub.cfg
```
davidcassany, line 30: If this file is not copied here then the `grub2-mkimage` call should also not make any reference to it: `-c /EFI/BOOT/grub.cfg` parameter can be dropped.

davidcassany: Probably drop whole `grub2-efi-image`

packages/grub2/config/grub_efi.cfg
```diff
@@ -1,4 +1,4 @@
-search --no-floppy --label --set=root @RECOVERY_LABEL@
+search --no-floppy --label --set=root ${recovery_label}
```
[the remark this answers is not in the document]
davidcassany: As said, probably better drop whole `grub2-efi-image`

davidcassany (commented):
Looks good, just a couple of comments:

* I'd fully drop `packages/cos-config`
* I'd fully drop `grub-efi-image` from `packages/grub2`. I can't see where this package is being used, I'd say this is no longer used since last refactor Itxaka did on EFI setup.

In any case do not spend much time on `grub-efi-image`, if dropping is tricky just leave it with a fixed value `COS_RECOVERY` and that's it. I think this is only for the disk builds which are already broken.

davidcassany:
I also realized the spec file will require some little adaptation too https://github.com/rancher/elemental-toolkit/blob/f0b9d4da0793c91cbbbb6fc2e88be6716f772585/.obs/specfile/elemental-toolkit.spec#L183-L184 I am not sure why the build error is not reported back to github from OBS 🤔 On success it does, I had expected it would also do it on failure...

frelon:
Seems like `grub-efi-image` is still used when building raw disks and it will take some work to move away from that (bootloader-in-rootfs maybe?).

Anyway I'm keeping the package and hardcoding the recovery-partition label for now.

davidcassany (approved):
LGTM

davidcassany:
> Seems like `grub-efi-image` is still used when building raw disks and it will take some work to move away from that (bootloader-in-rootfs maybe?).
> 
> Anyway I'm keeping the package and hardcoding the recovery-partition label for now.

That's good, build-disk command needs to revisited to be properly aligned with install command and use the same/equivalent partitions API.

PR #1736, opened 2023-03-01
Make sure STATE is mounted after initqueue executed fsck

davidcassany:
This PR is to ensure fsck is executed before mounting STATE partition. On some rare cases, on reboot, due to rancher/elemental#587 there is a dirty bit in STATE partition causing the STATE partition mount to fail, while this should have been already fixed by fsck it was not as because the STATE mount was not waiting for initqueue service.

This together with rancher/elemental#707 and this [OBS submission](https://build.opensuse.org/request/show/1068496) fixes some sporadic but critical issues during upgrades caused the boot assessment (rancher/elemental#708)


frelon (approved):
LGTM

Issue #1738, opened 2023-03-06
[epic] Drive elemental-toolkit to v1.x


davidcassany:
/This is an epic to mostly track actionable items towards https://github.com/rancher/elemental-toolkit/discussions/1716

The work done here can be based on top of this [PoC](https://github.com/davidcassany/elemental-toolkit/pull/3) started in Hackweek.

Actionable prior merge items:
- [x] Set a working branch with new _package_ names if suitable
- Expand the PR test matrix to cover (keep all of them easily executable on a local env)
  - [x] rancher/elemental-toolkit#1756
- [x] Adapt OBS builds of the elemental-toolkit package

- [x] Set a nightly build workflow on main (same test matrix as in PRs) x86_64 and aarch64
- [x] Drop `luet`

After merging:
- [x] #1762
- [x] rancher/elemental-toolkit#1766
- [x] rancher/elemental-toolkit#1754
- [x] rancher/elemental-toolkit#1755
- [x] #1792
- [x] #1793
- [x] https://github.com/rancher/elemental-toolkit/issues/1805 (move green example to 15.5 and add TW)

mbologna:
- [ ] Do the actual release

PR #1741, opened 2023-03-13
Refactor grub.cfg

frelon:
Remove search commands in order to speed up boot. Search can in some cases be very slow and should be used sparingly.

This means we will only have grub-config files on state partition (instead of searching on any device).

Also removes the grubmenu file from loading since it's used in the same way as grubcustom.

### DIFF
--- a/packages/cloud-config/oem/08_boot_assessment.yaml
+++ b/packages/cloud-config/oem/08_boot_assessment.yaml
@@ -78,10 +78,9 @@ stages:
          permsisions: 0600
          content: |
             set extra_active_cmdline="rd.emergency=reboot rd.shell=0 panic=5 systemd.crash_reboot systemd.crash_shell=0"
-            set boot_assessment="/boot_assessment"
-            search --no-floppy --file --set=boot_assessment_blk "${boot_assessment}"
-            if [ "${boot_assessment_blk}" ]; then
-              load_env -f "(${boot_assessment_blk})${boot_assessment}"
+            set boot_assessment_file="/boot_assessment"
+            if [ -f "${boot_assessment_file}" ]; then
+              load_env -f "${boot_assessment_file}"
             fi
             if [ "${enable_boot_assessment}" = "yes" -o "${enable_boot_assessment_always}" = "yes" ]; then
               if [ -z "${selected_entry}" ]; then
@@ -90,7 +89,7 @@ stages:
                   set extra_passive_cmdline="upgrade_failure"
                 else
                   set boot_assessment_tentative="yes"
-                  save_env -f "(${boot_assessment_blk})${boot_assessment}" boot_assessment_tentative
+                  save_env -f "${boot_assessment_file}" boot_assessment_tentative
                 fi
               fi
             fi
--- a/packages/grub2/config/grub.cfg
+++ b/packages/grub2/config/grub.cfg
@@ -19,7 +19,7 @@ if [ "${next_entry}" ]; then
   set default="${next_entry}"
   set selected_entry="${next_entry}"
   set next_entry=
-  save_env -f "${env_file}" next_entry
+  save_env -f "(${oem_blk})${env_file}" next_entry
 else
   set default="${saved_entry}"
 fi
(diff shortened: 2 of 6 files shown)


packages/grub2/config/grub.cfg
```diff
@@ -74,12 +57,8 @@ menuentry "${display_name} (fallback)" --id fallback {
...
 
 menuentry "${display_name} recovery" --id recovery {
-  # label and recoverylabel are kept around for backward compatibility
-  if [ -n "${system_label}" ]; then
-    set label=${system_label}
-  else
-    set recoverylabel=${recovery_label}
-  fi
+  # label is kept around for backward compatibility
+  set label=${system_label}
```
davidcassany, line 61: I do not entirely remember why label and recoverylabel were required in terms of backward compatibility. However I suspect that if label is still required recoverylabel is also. Need to look at it closer to asses it.

davidcassany (commented):
This looks really good. IMHO we just need to verify if the recoverylabel is required to boot older images.

Is there anything else to keep it as a draft?

davidcassany:
@frelon to ensure this is being used within the CI you should also bump the `cos` package version. Otherwise it is likely that the `cos` package including the grub2 changes is not being used. Once doing that I believe the squashfs tests will fail due to dropping the `recoverylabel` variable.

davidcassany (approved):
Looks good :+1:

We can then do a quick follow up PR for updating the docs https://rancher.github.io/elemental-toolkit/docs/customizing/configure_grub/ I would not get much into detail, just make sure there are no big inconsistencies. After all elemental-toolkit docs will require a deep review in the scope of #1738

packages/grub2/config/grub.cfg
```diff
@@ -1,35 +1,18 @@
...
-if [ "${oem_blk}" ] ; then
-  load_env -f "(${oem_blk})${oem_env_file}"
-fi
+set custom_file="/grubcustom"
 
-if [ "${env_blk}" ] ; then
-  load_env -f "(${env_blk})${env_file}"
-fi
+load_env -f "${oem_env_file}"
+load_env -f "${env_file}"
```
davidcassany, line 8: on a closer look, I believe the last variables file to load should be `/grub_oem_env`, so we ensure no custom variables from grubenv are overwriting those. `grubenv` should be the RW file stored in `/oem`. Also we should check the file exists before reading it.

packages/grub2/config/grub.cfg
```diff
@@ -1,35 +1,25 @@
...
+if [ -f "${oem_env_file}" ]; then
+  load_env -f "${oem_env_file}"
 fi
 
-if [ "${env_blk}" ] ; then
-  load_env -f "(${env_blk})${env_file}"
+search --no-floppy --set oem_blk --label "${oem_label}"
+
+if [ -f "(${oem_blk})${env_file}" ]; then
+  load_env -f "(${oem_blk})${env_file}"
```
davidcassany, line 14: I'd load fist the `env_file` and after the `oem_env_file`, so the `oem_env_file` has preference in case of a variable overlap.

davidcassany (commented):
Just a couple of comments

packages/grub2/config/grub.cfg
frelon: Hmm, isn't the `oem_label` set in `oem_env_file`?

davidcassany: mm good point 😅

packages/cloud-config/oem/08_boot_assessment.yaml
```diff
@@ -90,7 +89,7 @@ stages:
                   set extra_passive_cmdline="upgrade_failure"
                 else
                   set boot_assessment_tentative="yes"
-                  save_env -f "(${boot_assessment_blk})${boot_assessment}" boot_assessment_tentative
+                  save_env -f "${boot_assessment_file}" boot_assessment_tentative
```
davidcassany, line 92: Damn it, this exact line is checked on tests, in fact the tests expect the boot_assessment_blk literal. I guess the literal the test checks could be changed to `set boot_assessment_tentative="yes"` or something similar

PR #1742, opened 2023-03-15
Update GRUB2 docs

frelon:
Grub configuration has changed to no longer search for files.

It will load grub_oem_env file from state partition and use that to find the oem partition and load grubenv from there.

Also adds documentation for grub_oem_env variables.

### DIFF
--- a/docs/content/en/docs/Customizing/configure_grub.md
+++ b/docs/content/en/docs/Customizing/configure_grub.md
@@ -2,20 +2,20 @@
 title: "GRUB"
 linkTitle: "GRUB"
 weight: 4
-date: 2017-01-05
+date: 2023-03-14
 description: >
   GRUB 2 Configuration
 ---
 
-COS is set to deploy a persistent `grub.cfg` into the `COS_RECOVERY` partition during
-the system installation or image creation. COS grub configuration
+Elemental is set to deploy a persistent `grub.cfg` into the `COS_RECOVERY` partition during
+the system installation or image creation. Elemental grub configuration
 includes three menu entries: first for the main OS system, second for the
 fallback OS system and a third for the recovery OS.
 
 For example the main OS system menu entry could be something like:
 
 ```
-menuentry "cOS" --id cos {
+menuentry "Elemental" --id elemental {
   search.fs_label COS_STATE root
   set img=/cOS/active.img
   set label=COS_ACTIVE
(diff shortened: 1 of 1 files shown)


docs/content/en/docs/Customizing/configure_grub.md
```diff
@@ -94,13 +96,13 @@ For instance use the following command to reboot to recovery system only once:
 ```
 
 {{% alert title="Note" %}}
-The examples below make of the `COS_OEM` device, however it could use any device
-detected by GRUB2 that includes the file `/grubenv`. First match wins.
+The examples below make of the `COS_STATE` device, only files in the state
+partition will be used when booting.
```
davidcassany, line 100: I am not sure about this note, in fact `/grubenv` is stored within the OEM paratition. In fact this should be only file stored in a RW partition, the other grub files are in state and supposed to be RO. Probably the note should simply state that `/oem/grubenv` is the only file under a RW partition, others are in state.

""

docs/content/en/docs/Customizing/configure_grub.md
```diff
@@ -28,10 +28,10 @@ menuentry "cOS" --id cos {
 ```
 
 {{% alert title="Kernel parameters" %}}
-The kernel parameters are not part of the persistent `grub.cfg` file stored in `COS_RECOVERY` partition. 
-Kernel parameters are sourced from the loop device of the OS image to boot. This is mainly to
-keep kernel parameters consistent across different potential OS images or
-system upgrades. 
+The kernel parameters are not part of the persistent `grub.cfg` file stored in
+`COS_RECOVERY` partition. Kernel parameters are sourced from the loop device of
```
davidcassany, line 32: The parameters are now always in COS_STATE, this was already wrong before...

davidcassany (approved):
Good work :+1: just a couple of comments.

fgiudici (approved):
looks good here!

docs/content/en/docs/Customizing/configure_grub.md
frelon: Oh I missed that! Nice catch!

docs/content/en/docs/Customizing/configure_grub.md
frelon: Yep, good point! Added oem partition in the note.

PR #1743, opened 2023-03-16
Wait for all persistent volumes

davidcassany:
This commit adds a wait logic for all persistent volumes configured within the immutable-rootfs dracut module. It has timeout of 60 seconds for each volume.

Note at this stage the root device is already found and mounted, since all volumes are expected to be part of the same disk it is really unlikely the wait is needed.

In addition this commit does not ignore any mountpoint in case of errors. It keeps creating the fstab as expected if the mountpoint could not be mounted or prepared.


packages/immutable-rootfs/30cos-immutable-rootfs/cos-mount-layout.sh
```diff
@@ -216,12 +216,25 @@ function mountPersistent {
...
 
+function waitForPath {
+    local path=$1
+    local count=0
+
+    while [ ! -e "${path}" ]; do
+        sleep 1
+        udevadm settle
+        (( count++ ))
+        [ "${count}" -ge "${timeout}" ] && return 1
```
fgiudici, line 227: nit: maybe we can move the `timeout` var declaration inside the function: it is just used here.

fgiudici (approved):
LGTM (just one nit ok to skip)

PR #1745, opened 2023-04-12
grub: Delete loopback device on fallback

frelon:
During boot if the active menu-entry fails the fallback will also fail with the error-message "device already exists".

This commit ensures we delete the loopback-device if it exists before booting to fallback or recovery.

Part of https://github.com/rancher/elemental/issues/708

### DIFF
--- a/packages/grub2/config/grub.cfg
+++ b/packages/grub2/config/grub.cfg
@@ -43,24 +43,30 @@ insmod gfxterm
 insmod loopback
 insmod squash4
 
+set loopdev="loop0"
+
 menuentry "${display_name}" --id cos {
   # label is kept around for backward compatibility
   set label=${active_label}
   set img=/cOS/active.img
-  loopback loop0 /$img
-  source (loop0)/etc/cos/bootargs.cfg
-  linux (loop0)$kernel $kernelcmd ${extra_cmdline} ${extra_active_cmdline}
-  initrd (loop0)$initramfs
+  loopback $loopdev /$img
+  source ($loopdev)/etc/cos/bootargs.cfg
+  linux ($loopdev)$kernel $kernelcmd ${extra_cmdline} ${extra_active_cmdline}
+  initrd ($loopdev)$initramfs
 }
 
 menuentry "${display_name} (fallback)" --id fallback {
   # label is kept around for backward compatibility
   set label=${passive_label}
   set img=/cOS/passive.img
-  loopback loop0 /$img
-  source (loop0)/etc/cos/bootargs.cfg
-  linux (loop0)$kernel $kernelcmd ${extra_cmdline} ${extra_passive_cmdline}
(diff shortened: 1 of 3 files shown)


davidcassany (commented):
mmm interesting, this raises when the failure happens within the grub2 code I guess.

How did you triggered this error? Ideally we should add  a new test case for the boot assessment tests, in any case, before adding tests we should discuss when and where to  include them. I am resistant to include new tests to the current project structure, it is just too complex.

frelon:
> Also I think that it requires a bump on cos package to make sure the CI pulls this change. The CI just rebuilds the modified packages without taking into account dependencies IIRC... disappointed

Ah, bumped it now!

packages/grub2/config/grub.cfg
```diff
@@ -43,24 +43,30 @@ insmod gfxterm
...
   set label=${passive_label}
   set img=/cOS/passive.img
-  loopback loop0 /$img
-  source (loop0)/etc/cos/bootargs.cfg
-  linux (loop0)$kernel $kernelcmd ${extra_cmdline} ${extra_passive_cmdline}
-  initrd (loop0)$initramfs
+  # if loopdev already exists, we've probably already tried booting once, delete it and try again.
+  if [ -e ($loopdev) ]; then
+    loopback -d $loopdev
+  fi
```
davidcassany, line 65: Couldn't we move this check in out of the menu setup just below the `set loopdev="loop0"` line?

I think it should not hurt if this is executed in all cases and it would make the code smaller and cleaner.

Probably I am missing some detail though

frelon: Hmm, it feels like grub runs only stuff in the "menuentry" when falling back.. here is a log with moving the `loopback -d ...` out of the menuentry:

 ```
 Booting `Elemental'

error: ../../grub-core/fs/fshelp.c:258:file `/boot/vmlinuz' not found.
error: ../../grub-core/loader/i386/efi/linux.c:207:you need to load the kernel
first.

Press any key to continue...

   Falling back to `Elemental'

error: ../../grub-core/disk/loopback.c:98:device name already exists.
error: ../../grub-core/fs/fshelp.c:258:file `/boot/vmlinuz' not found.
error: ../../grub-core/loader/i386/efi/linux.c:207:you need to load the kernel
first.

Press any key to continue...

   Falling back to `Elemental (fallback)'

error: ../../grub-core/disk/loopback.c:98:device name already exists.
error: ../../grub-core/fs/fshelp.c:258:file `/boot/vmlinuz' not found.
error: ../../grub-core/loader/i386/efi/linux.c:207:you need to load the kernel
first.

Press any key to continue...

```

davidcassany (approved):

frelon:
Marking this as blocked until after release of new version of elemental-teal.

PR #1760, opened 2023-05-10
Removing fallback test of a corrupted OS without upgrading

davidcassany:


tests/fallback/fallback_test.go
```diff
@@ -86,69 +87,5 @@ var _ = Describe("Elemental booting fallback tests", func() {
 				return out
 			}, 5*time.Minute, 10*time.Second).Should(ContainSubstring("upgrade_failure"))
 		})
-
-		It("without upgrades boots in fallback when rootfs is damaged, triggered by missing files", func() {
```
davidcassany, line 90: I fail to see the actual value of this test, probably I am missing something... but to me it feels like doing the same as the previous one but without calling `elemental upgrade` first, I don't see how this is that relevant in terms of the boot assessment logic. I'd say we can get rid of this test.

frelon (approved):
LGTM

PR #1763, opened 2023-05-12
Merge elemental-cli

frelon:
Merge rancher/elemental-cli into rancher/elemental-toolkit and all changes needed to get it working.


frelon:
The merge was done using the following commands:

```
git switch -c merge-cli
git remote add cli git@github.com:rancher/elemental-cli.git
git fetch cli
git merge cli/main --allow-unrelated-histories
```

And then fixing a couple of merge conflicts.

After that to get the cli building and using the same go.mod for tests i renamed the module to `github.com/rancher/elemental-toolkit` and had to change imports in the cli.

Will also have to go through the actions and more work on the Makefile to get it green.

davidcassany (approved):
@frelon we went through it a couple of times in the past days and finally the tests are as we wanted 😉 I really like the idea of dropping grsync. Current rsync approach is easier to maintain and extend :+1:

I'd say any further fine tuning, if any, should be done in follow up PRs even if it is just for clarity.

Nice job!

PR #1779, opened 2023-05-31
Add aarch64 workflow

### LINKED ISSUE #1754: Set aarch64 workflow
Make a workflow for arm64 parallel to the current one in x86 to run tests ondemand, on prs (if certain label is in) and on weekly or daily basis. We should reuse the already defined gcp runners for that.

frelon:
This commit adds aarch64 to the build-matrix for PRs and checks if the PR is labeled with arm64 before running.

Fixes #1754


davidcassany (approved):
Nice :+1:

PR #1789, opened 2023-06-16
Clean up rsync progress display

frelon:
The current implementation has a memory leak, use Ticker instead.

Also add `--partial`, `--progress`, `--human-readable` flags to rsync call.

### DIFF
--- a/pkg/elemental/elemental_test.go
+++ b/pkg/elemental/elemental_test.go
@@ -498,8 +498,8 @@ var _ = Describe("Elemental", Label("elemental"), func() {
 			runner.SideEffect = func(cmd string, args ...string) ([]byte, error) {
 				if cmd == "rsync" {
 					rsyncCount += 1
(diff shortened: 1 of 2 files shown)


davidcassany (approved):
Looks good to me :+1:

ldevulder:
@frelon just out of curiosity: how Ticker helps to remove/reduce the memory leak?

frelon:
@ldevulder ran some tests with the different implementation, and it was because I used a break instead of a return in the select, which breaks from the select and causes an infinite loop instead of breaking from the loop:

```
	stop := make(chan bool)
	go func() {
		for {
			select {
			case <-stop:
				break // <- this should be a return
			case <-time.After(5 * time.Second):
				fmt.Println(message)
			}
		}
	}()
```

So the Ticker implementation is not needed, but I will keep it since it's already merged!

PR #1794, opened 2023-07-03
Build disk command cherry pick

davidcassany:
Port from https://github.com/davidcassany/elemental-toolkit/pull/4 

This PR adds in elemental client the `build-disk` command.

With this command we can eventually build an image that includes partitions:
* EFI
* OEM
* Recovery
* State

Having State partition to match the minimum size (to reduce resulting image size), only includes config files no image.

Then the State partition could be expanded on first boot to desired size (`build-disk` command already pre-appends the required cloud-config files for that to happen) and then the Persistent partition created at the end with all the available space (or some desired specific size too). 

This setup can be executed without running a single `mount` (thanks to squashfs usage), meaning this disk could be built in a container or Dockerfile (like we do with ISOs).

This is the build-disk command help message:
```
Build a disk image using the given image (experimental and subject to change)

Usage:
  elemental build-disk image [flags]

Flags:
      --cosign                           Enable cosign verification (requires images with signatures)
      --cosign-key string                Sets the URL of the public key to be used by cosign validation
      --date                             Adds a date suffix into the generated disk file
      --expandable                       Creates an expandable image including only the recovery image
  -h, --help                             help for build-disk
      --local                            Use an image from local cache
  -n, --name string                      Basename of the generated disk file
  -o, --output string                    Output directory (defaults to current directory)
  -x, --squash-compression stringArray   cmd options for compression to pass to mksquashfs. Full cmd including --comp as the whole values will be passed to mksquashfs. For a full list of options please check mksquashfs manual. (default value: '-comp xz -Xbcj ARCH')
      --squash-no-compression            Disable squashfs compression. Overrides any values on squash-compression
  -t, --type string                      Type of image to create (default "raw")
      --unprivileged                     Makes a build runnable within a non-privileged container, avoids mounting filesystems (experimental)
```

The command includes disks types for public cloud as this is taken from former `build-disk` command back in a day. I doubt we should consider those (it is also hard to test and validate in a CI/CD fashion), so included here because we actually had the logic coded (I have not test it myself, it was back in a day more than a year ago). So I'd probably remove those and consider qcow2, vmdk and ova for the future.


davidcassany:
It can be tested with the following process:

write a `manifest.yaml` as:
```yaml
disk:
  unprivileged: true
  expandable: true
  recovery-system:
    fs: squashfs
  type: raw
```

Then build the example OS with `make build-os`.

Then call the resulting image with:

```bash
docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v $(pwd)/build:/build -v $(pwd)/manifest.yaml:/manifest.yaml     --entrypoint /usr/bin/elemental ${REPO}:${VERSION} --debug build-disk --config-dir / -n elemental     --local --squash-no-compression -o /build ${REPO}:${VERSION}
```

Where `REPO` and `VERSION` are the repository and tag of the resulting image of executing `make build-os`. This will produce an image as `build/elementa.raw`. This image can be dded to a bigger disk with:

```bash
$ dd if=build/elemental-disk.raw of=somedisk.img bs=4M conv=notrunc
```

Boot `somedisk.img` with qemu and the system will create missing partitions and expand them after first boot and then reboot.

davidcassany:
Improvements I foresee that require a further elemental-toolkit refactor:

* Do not install grub configuration in State partition (all in EFI partition?): this would allow us to actually build disks with no state partition and with a single recovery partition that suits the recovery image size. Then, on boot, expand recovery, create state and create persistent.
* Consider squashfs as the default filesystem for recovery

davidcassany:
> > * Do not install grub configuration in State partition (all in EFI partition?): this would allow us to actually build disks with no state partition and with a single recovery partition that suits the recovery image size. Then, on boot, expand recovery, create state and create persistent.
> 
> Agreed! Moving the boot configuration to EFI partition makes sense.

Yes, I am just wondering how to run the move in a sane way to allow upgrades, messing with grub setup is a PITA 😅 I was wondering if shouldn't we have a grub reinstall command, so we can eventually call it in an after-upgrade-hook or something like that. The problem is that any error when handling grub artifacts is likely to turn lead the system into an unrecoverable state.

kkaempf:
related issue: https://github.com/rancher/elemental/issues/947

frelon (approved):
Very nice work!

LGTM and then we can try using this instead of the yanky `elemental install /dev/loop0` for building images in the Makefile :+1:

shanduur (approved):

PR #1801, opened 2023-07-14
Drop SLE Micro version refrences from the code

davidcassany:
These are the changes required in this repository for the builds I prepared in [here](https://build.opensuse.org/project/show/home:dcassany:ElementalNewDev). 

Note this changes image and charts URLs, so this is likely to be a breaking change from CI perspective.

Part of rancher/elemental#815

### DIFF
--- a/.obs/dockerfile/Dockerfile
+++ b/.obs/dockerfile/Dockerfile
@@ -1,6 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
 # Define the names/tags of the container
-#!BuildTag: rancher/elemental-builder-image:latest
 #!BuildTag: rancher/elemental-builder-image:%VERSION%
 #!BuildTag: rancher/elemental-builder-image:%VERSION%-%RELEASE%
 #
--- a/.obs/dockerfile/Dockerfile
+++ b/.obs/dockerfile/Dockerfile
@@ -1,23 +1,28 @@
 # SPDX-License-Identifier: Apache-2.0
 # Define the names/tags of the container
-#!BuildTag: rancher/elemental-builder-image/5.3:latest
-#!BuildTag: rancher/elemental-builder-image/5.3:%VERSION%
-#!BuildTag: rancher/elemental-builder-image/5.3:%VERSION%-%RELEASE%
+#!BuildTag: rancher/elemental-builder-image:latest
+#!BuildTag: rancher/elemental-builder-image:%VERSION%
+#!BuildTag: rancher/elemental-builder-image:%VERSION%-%RELEASE%
 #
 
-FROM suse/sle15:15.4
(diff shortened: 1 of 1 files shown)


.obs/dockerfile/Dockerfile
```diff
@@ -1,23 +1,28 @@
 # SPDX-License-Identifier: Apache-2.0
 # Define the names/tags of the container
-#!BuildTag: rancher/elemental-builder-image/5.3:latest
-#!BuildTag: rancher/elemental-builder-image/5.3:%VERSION%
-#!BuildTag: rancher/elemental-builder-image/5.3:%VERSION%-%RELEASE%
+#!BuildTag: rancher/elemental-builder-image:latest
```
kkaempf, line 3: Hmm, this might fail while we build two versions (like `5.3` and `5.4`) in parallel.

We should probably just drop the `:latest` tag ?!

kkaempf (changes requested):

.obs/dockerfile/Dockerfile
davidcassany: Done, we keep `latest` and also the slemicro version, but it resolves dynamically to 5.3 or 5.4

.obs/dockerfile/Dockerfile
```diff
@@ -1,23 +1,29 @@
...
 LABEL org.opencontainers.image.description="Provides Elemental Client and required dependencies for installation media building"
 LABEL org.opencontainers.image.version="%VERSION%"
-LABEL org.opencontainers.image.url="https://github.com/rancher/elemental-cli"
+LABEL org.opencontainers.image.url="https://github.com/rancher/elemental-toolkit"
 LABEL org.opencontainers.image.created="%BUILDTIME%"
 LABEL org.opencontainers.image.vendor="SUSE LLC"
-LABEL org.opensuse.reference="%%IMG_REPO%%/rancher/elemental-builder-image/5.3"
+LABEL org.opensuse.reference=$IMAGE_REPO
 LABEL org.openbuildservice.disturl="%DISTURL%"
 LABEL com.suse.supportlevel="techpreview"
```
fgiudici, line 28: the supportlevel here is "techpreview": should we bump it to l2 (I don't remember if it really was l2 or something elese 😓 )?

fgiudici (approved):
lgtm

.obs/dockerfile/Dockerfile
davidcassany: not sure, this image is not customer facing, this is is just the ISO builder image, the one including `elemental` to actually call `elemental build-iso`. This is just part of internal tooling to build the vanilla ISO. This was left as is in purpose.

kkaempf (approved):

PR #1810, opened 2023-07-27
Add suite and fix unit tests for features

davidcassany:
Features tests were not included in a test suite. Ginkgo was complaining about that but not reporting an error (it was returning 0 🤷🏽‍♂️ ) neither reporting the result of the features tests. I assume it went unnoticed.

### DIFF
--- a/pkg/features/features_test.go
+++ b/pkg/features/features_test.go
@@ -17,11 +17,18 @@
 package features_test
 
 import (
+	"testing"
+
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
 	"github.com/rancher/elemental-toolkit/pkg/features"
 )
+
+func TestTypes(t *testing.T) {
+	RegisterFailHandler(Fail)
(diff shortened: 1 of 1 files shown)


pkg/features/features_test.go
```diff
@@ -17,24 +17,26 @@
...
 		It("Returns an empty list of features", func() {
 			feats, err := features.Get([]string{})
 			Expect(err).ToNot(HaveOccurred())
 			Expect(feats).To(BeEmpty())
 		})
-		It("Returns all features", func() {
-			feats, err := features.Get([]string{})
-			Expect(err).ToNot(HaveOccurred())
-			Expect(len(feats)).To(Equal(len(features.All)))
-		})
```
davidcassany, line 37: This test was essentially wrong, does the same exact thing as the previous one but expects a different output 😓 I guess this is an unnoticed leftover as the feature tests were not properly executed by ginkgo as they were not included in any test suite.

anmazzotti (approved):
So this never ran before?
Nice catch.

PR #1812, opened 2023-08-02
Drop builder image in OBS

davidcassany:
This commit drops the builder image in OBS and adds xorriso and mtools to the requirements of elemental-cli.

This allows using the OS image to build the ISO from itself.


kkaempf:
I'm not too happy about mixing build-time with run-time requirements.

kkaempf (approved):
Approving for now.
To be revisited when we're asked to minimize image sizes.

davidcassany:
> I'm not too happy about mixing build-time with run-time requirements.

Sure, this was just an attempt I did. The idea is to reduce the number of images we manage, so we keep only the images from `scratch` (elemental-operator, seedimage-builder) and from busybox (the image containing the iso) and the images of the OS itself. So there is no image based on an OS generic image and all images we have and customer facing.

The expense of it is adding xorriso and mtools to the current elemental-teal, which causes an increase of around ~2MB compressed and an increase of around ~2MB when uncompressed too. In terms of size it is not painful, its more of a conceptual matter if we should distinguish the requirements for elemental-cli for runtime commands (install, upgrade, reset, etc.) vs buildtime commands (build-iso and probably build-disk in the future).

PR #1815, opened 2023-08-09
Modify boot dir structure to cope with PXE SUMA/Uyuni defaults

anmazzotti:
Closes rancher/elemental#721

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -109,8 +109,8 @@ const (
 	SELinuxTargetedPolicyPath  = SELinuxTargetedPath + "/policy"
 
 	// Kernel and initrd paths are arbitrary and coupled to grub.cfg
-	ISOKernelPath    = "/boot/kernel"
-	ISOInitrdPath    = "/boot/initrd"
+	ISOKernelPath    = "/boot/x86_64/loader/linux"
+	ISOInitrdPath    = "/boot/x86_64/loader/initrd"
 	ISORootFile      = "rootfs.squashfs"
 	ISOEFIImg        = "uefi.img"
 	ISOLabel         = "COS_LIVE"
--- a/pkg/live/common.go
+++ b/pkg/live/common.go
@@ -33,9 +33,9 @@ const (
 
 	// TODO document any custom BIOS bootloader must match this setup as these are not configurable
 	// and coupled with the xorriso call
-	isoLoaderPath = "/boot/x86_64/loader"
-	isoHybridMBR  = isoLoaderPath + "/boot_hybrid.img"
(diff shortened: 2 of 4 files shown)


pkg/constants/constants.go
```diff
@@ -109,8 +109,8 @@ const (
 	SELinuxTargetedPolicyPath  = SELinuxTargetedPath + "/policy"
 
 	// Kernel and initrd paths are arbitrary and coupled to grub.cfg
-	ISOKernelPath    = "/boot/kernel"
-	ISOInitrdPath    = "/boot/initrd"
+	ISOKernelPath    = "/boot/x86_64/loader/linux"
```
kkaempf, line 112: Can we get the actual architecture here instead of hard-coding `x86_64` ?!

(Yes, this is technically irrelevant as only x86_64 supports PXE booting in SUSE Manager, but it would make for cleaner code 😉 )

pkg/constants/constants.go
```diff
@@ -109,8 +109,8 @@ const (
...
+	ISOInitrdPath    = "/boot/x86_64/loader/initrd"
```
kkaempf, line 113: same here

pkg/live/common.go
```diff
@@ -33,18 +33,18 @@ const (
 
 	// TODO document any custom BIOS bootloader must match this setup as these are not configurable
 	// and coupled with the xorriso call
-	isoLoaderPath = "/boot/x86_64/loader"
-	isoHybridMBR  = isoLoaderPath + "/boot_hybrid.img"
-	isoBootFile   = isoLoaderPath + "/eltorito.img"
+	IsoLoaderPath = "/boot/x86_64/loader"
```
kkaempf, line 36: and here

kkaempf (changes requested):
:+1: otherwise

pkg/constants/constants.go
anmazzotti: Oh apologies, you mentioned this is only supported for `x86_64`, so we can use the `/boot/x86_64/loader` dir in `amd64` case, or revert to the previous `/boot` for all other cases

kkaempf: No, I wouldn't revert for non-x86_64. Then we would have two code pathes.

kkaempf: > I guess I could fetch runtime.GOARCH and map amd64 (if any) to x86_64, but then should I also map the other values? Would that make sense?

I see, probably too much effort. (I had hoped for a simple fix)

Let's merge as-is

kkaempf (approved):

PR #1820, opened 2023-09-05
Remove block: prefix in immutable-rootfs

frelon:
The wait_for_dev dracut function will wait for the device with that exact name. In later versions 'block:' is added in parse-rootfs in dracut which means we will wait for a device that will never exist.

### DIFF
--- a/pkg/features/embedded/immutable-rootfs/usr/lib/dracut/modules.d/30elemental-immutable-rootfs/parse-elemental-cmdline.sh
+++ b/pkg/features/embedded/immutable-rootfs/usr/lib/dracut/modules.d/30elemental-immutable-rootfs/parse-elemental-cmdline.sh
@@ -43,7 +43,7 @@ esac
 
 info "root device set to root=${root}"
 
-wait_for_dev -n "${root}"
+wait_for_dev -n "${root#block:}"
 
 # Only run filesystem checks on force mode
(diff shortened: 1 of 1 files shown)


pkg/features/embedded/immutable-rootfs/usr/lib/dracut/modules.d/30elemental-immutable-rootfs/parse-elemental-cmdline.sh
```diff
@@ -43,7 +43,7 @@ esac
 
 info "root device set to root=${root}"
 
-wait_for_dev -n "${root}"
+wait_for_dev -n "${root#block:}"
```
davidcassany, line 46: I assume the change is backward compatible, I mean it will also work on old dracut versions, right?

davidcassany (approved):
Looks good to me

pkg/features/embedded/immutable-rootfs/usr/lib/dracut/modules.d/30elemental-immutable-rootfs/parse-elemental-cmdline.sh
frelon: Yes, it will only remove block: prefix if it exists. :+1:

PR #1823, opened 2023-09-06
Add tumbleweed example

frelon:
Pretty much copy pasted from the green example and remove the tmpfs unit.


davidcassany (approved):
Cool! :+1:

PR #1828, opened 2023-09-15
Remove deprecated cmdline options

frelon:
Remove tty, force-efi, force-gpt and partition-layout deprecated flags and set the default grub entry to "Elemental".


davidcassany (approved):
LGTM

PR #1832, opened 2023-09-20
Move grub bootargs.cfg into new feature

### LINKED ISSUE #1831: Elemental init command should differentiate between generic grub config and kernelcmd setup
**elemental-toolkit version:**

v1.0.0

irrelevant

**Describe the bug**

It is likely the case that derivatives require their own kernelcmd setup, hence decoupling it from the `grub-config` feature I think it makes a lot of sense.
In fact, in former elemental-toolkit package we used to pack them separately, there was the `elemental-grub-config` package and the `elemental-grub-bootargs` package. So installing `elemental-grub-config` was not colliding with anyother file.

**To Reproduce**

Use current elemental-teal from Dev and then elemental-init command will overwrite the kernelcmd configuration provided by the `elemental` package.

**Expected behavior**

Elemental-teal image includes the kernelcmd setup defined [here](https://github.com/rancher/elemental/blob/main/framework/files/etc/cos/bootargs.cfg).

**Logs**

none

**Additional context**

I am also wondering if `elemental init` should fail in case it overwrites an existing file, or at least, have some sort of `strict` of `force` flag to control this behavior and fail or not when overwriting depending on them.

frelon:
The default bootargs often end up overwriting bootargs in derivative, this commit moves the bootargs.cfg file into grub-default-bootargs feature.

Fixes #1831


davidcassany (approved):
LGTM thanks :+1:

PR #1833, opened 2023-09-20
Worflow was still trying to build the already dropped builder image

davidcassany:
This is a follow up of adb3dcd1470c79adfff865a12c6a17884eabc617


frelon (approved):
LGTM

PR #1834, opened 2023-09-21
Change examples kernel cmdline

frelon:
Add noapic option to try to get around flaky tests that run into kernel panic.


davidcassany:
I think to get around the kernel panic we could use the following

```
no_timer_check  [X86,APIC] Disables the code which tests for
                broken timer IRQ sources.
```

I think this way we are not defeating the poweroff and reboot flags. Also I would consider using an after-install hook to run  a `grub2-editenv` to set the extra command line parameters. This way we can eventually limit this parameter to CI tests. I'll to see if I manage to set it up.

frelon:
Closing this since we won't merge as is.

Issue #1830, opened 2023-09-19
Use non-FHS path as default for persistent state storage


frelon:
Right now we mount the persistent partition to `/usr/local` and put state-overlays/binds in `/usr/local/.state`.

This hides any applications installed in a derivative in `/usr/local` and I would propose we use some other path as default for this (`/.persistent` or `/run/elemental/persistent` comes to mind).

Users would still be able to make `/usr/local` into a persistent path, but our default would not mess with users expectations of a new system.

davidcassany:
Yes we were wondering about that in the past and I think it actually makes a lot of sense. I think I'd vote to `/run/elemental/persistent`. This would also allow as to move as much as possible files introduced with `elemental init` into `/etc` or `/usr/local`, so we do not mess with distro provided stuff.

PR #1837, opened 2023-09-26
Add mount command

### LINKED ISSUE #1781: Implement immutable-rootfs module in elemental-cli
In order to get better testability and hopefully some new features we should implement the immutable-rootfs dracut module in the elemental-cli.

frelon:
The mount command mounts the system and is meant to run in an initrd to actually mount the root filesystem and use systemd to switch-root into it.

It also optionally writes an /etc/fstab file to the newly mounted system so that systemd will mount the system after switching root.

The command is used in the new dracut module elemental-rootfs, which will coexist with immutable-rootfs (they are functionally the same) until immutable-rootfs can be deprecated.

Fixes #1781, #1830

### DIFF
--- a/pkg/action/mount.go
+++ b/pkg/action/mount.go
@@ -34,14 +34,8 @@ func RunMount(cfg *v1.RunConfig, spec *v1.MountSpec) error {
 
 	e := elemental.NewElemental(&cfg.Config)
 
-	allParts, err := utils.GetAllPartitions()
-	if err != nil {
-		cfg.Logger.Errorf("Error getting all partitions: %s", err.Error())
-		return err
-	}
-
 	cfg.Logger.Debug("Fscking partitions")
-	if err := e.FsckPartitions(allParts); err != nil {
+	if err := RunFsck(cfg, spec, e); err != nil {
 		cfg.Logger.Errorf("Error fscking partitions: %s", err.Error())
 		return err
 	}
@@ -135,6 +129,21 @@ func applyLayoutConfig(cfg *v1.RunConfig, spec *v1.MountSpec) error {
 	}
 
 	return nil
+}
+
+func RunFsck(cfg *v1.RunConfig, spec *v1.MountSpec, e *elemental.Elemental) error {
+	if !spec.RunFsck {
+		cfg.Logger.Debug("Skipping fsck")
+		return nil
+	}
+
+	allParts, err := utils.GetAllPartitions()
+	if err != nil {
+		cfg.Logger.Errorf("Error getting all partitions: %s", err.Error())
+		return err
+	}
+
+	return e.FsckPartitions(allParts)
 }
 
 func MountOverlay(cfg *v1.RunConfig, sysroot string, overlay v1.OverlayMounts) error {
--- a/pkg/action/mount.go
+++ b/pkg/action/mount.go
@@ -30,6 +30,8 @@ import (
 	"github.com/rancher/elemental-toolkit/pkg/utils"
 )
 
+const overlaySuffix = ".overlay"
+
 func RunMount(cfg *v1.RunConfig, spec *v1.MountSpec) error {
 	cfg.Logger.Info("Running mount command")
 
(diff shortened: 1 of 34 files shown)


frelon:
Pushing this as a rough draft for now, still missing the following pieces:

- [ ] Mount persistent overlays + bind mount
- [ ] Reading kernel cmdline
- [ ] Configuration from env-vars
- [ ] Write sentinel files (recovery_mode, active_mode, passive_mode)
- [ ] Tests

davidcassany (commented):
Looks so promising 😉

pkg/features/embedded/elemental-rootfs/usr/lib/dracut/modules.d/30elemental-rootfs/module-setup.sh
```diff
@@ -0,0 +1,45 @@
...
+# called by dracut
+install() {
+    declare moddir=${moddir}
+    declare systemdutildir=${systemdutildir}
+    declare systemdsystemunitdir=${systemdsystemunitdir}
+
+    inst_multiple \
+        mount mountpoint sort rmdir findmnt rsync cut realpath basename lsblk
+
+    inst_simple "/etc/elemental/config.yaml"
```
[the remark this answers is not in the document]
frelon: Yep, in my opinion generating the `cos-layout.env` is a very convoluted way of going about it.. the big win of generating the `/run/cos/cos-layout.env` from cloud-init is that we can check if we are running in recovery mode and then generate another config, but with the new command we can already check which mode we are running in. In my mind this should be handled inside the code-path.

We could also add a setting for mounting in recovery:

```
elemental:
  mount:
    rw-paths: []
    persistent:
      ..
    recovery:
      rw-paths: [] # or some other setting...
```

And we could use the values from `cos-layout.env` for backwards compatibility... WDYT?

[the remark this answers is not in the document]
frelon: After discussing this we will add backwards compatibility through reading the `/run/cos/cos-layout.env` as well.

pkg/action/mount.go
```diff
@@ -0,0 +1,318 @@
...
+
+	cfg.Logger.Debugf("Mounting image %s", spec.Image.File)
+	if err := e.MountImage(spec.Image); err != nil {
+		cfg.Logger.Errorf("Error mounting image %s: %s", spec.Image.File, err.Error())
+		return err
+	}
+
+	if spec.RunCloudInit {
+		cfg.Logger.Debug("Running rootfs cloud-init stage")
+		err = utils.RunStage(&cfg.Config, "rootfs", cfg.Strict, cfg.CloudInitPaths...)
```
frelon, line 51: This feels a little hacky since all the other stages are installed in the `elemental-setup` feature. Maybe we should move the units from that feature into elemental-rootfs and then try to `systemd-cat` this stage into the journal to get more streamlined handling of the different stages? Or maybe move the fs-dependant stages into the mount-command.

davidcassany (approved):
Very nice! LGTM

frelon:
@davidcassany I ended up bumping go to 1.21 as well since I needed it for the `slices` module, let me know if you want to stay on 1.20!

pkg/action/mount.go
```diff
@@ -0,0 +1,340 @@
...
+	default:
+		return fmt.Errorf("unknown overlay type '%s'", overlay.Type)
+	}
+
+	if err := cfg.Mounter.Mount(overlaySource, constants.OverlayDir, overlayFS, overlayOpts); err != nil {
+		cfg.Logger.Errorf("Error mounting overlay: %s", err.Error())
+		return err
+	}
+
+	for _, path := range overlay.Paths {
```
Itxaka, line 180: a small issue we encountered was bind paths not being ordered, so something like this mistake in the layout file:

```
        /usr/share/pki/trust/anchors
	/usr/share/pki/trust
```

would shadow the first bind with the second! Just a heads up as we encountered this :) Patch looks awesome :+1:

frelon: @Itxaka thanks for the heads up! I will add a commit to sort these in correct order! :+1:

frelon:
Closing in favor of #1866

PR #1842, opened 2023-10-11
Ensure the cache is renewed on source changes

davidcassany:
This commit adds to the key cache validation a hash based on the elemental client code.

In addition it also provides a dockerignore file to prevent adding to docker build environments some built artifacts, such as disks, isos, etc.


frelon (approved):
LGTM!

PR #1843, opened 2023-10-11
Improve CI stability

davidcassany:
~This is brute force attempt to check if adding this simple no_timer_check kernel parameter is solving the tests flakiness.~

[EDIT]

Adds several improvements:

* Check VM pid on 'EventuallyConnects'. So it does not wait for a
  command to succeed if the underlaying VM crashed and it is not
  running.

* Does not use '-daemonize' flag of qemu, now it simply runs on the
  background and the stdout and stderr are redirected to vmstdout file.

* Does not install QEMU, the runner already has a recent QEMU version
  installed. This saves several minutes on each macos job.

* Fixes some of the stability issues on macOS by disabling hugepages on
  the kernel. This is not supported on macOS.

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -157,7 +157,6 @@ jobs:
       - if: ${{ env.ARCH == 'x86_64' }} 
         name: Run VM script dependencies
         run: |
-          brew update; brew upgrade qemu
           brew install bash coreutils
       - if: ${{ env.ARCH == 'x86_64' }} 
         name: Prepare test (x86_64)
@@ -177,6 +176,13 @@ jobs:
         with:
           name: serial-${{ env.ARCH }}-${{ env.FLAVOR }}-${{ matrix.test }}.log
           path: tests/serial.log
+          if-no-files-found: error
+      - name: Upload qemu stdout for ${{ matrix.test }}
+        uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: vmstdout-${{ env.ARCH }}-${{ env.FLAVOR }}-${{ matrix.test }}.log
+          path: tests/vmstdout
           if-no-files-found: error
       - name: Stop test VM
         if: always()
@@ -216,7 +222,6 @@ jobs:
       - if: ${{ env.ARCH == 'x86_64' }} 
         name: Run VM script dependencies
         run: |
-          brew update; brew upgrade qemu
           brew install bash coreutils
       - if: ${{ env.ARCH == 'x86_64' }} 
         name: Prepare test (x86_64)
@@ -236,6 +241,13 @@ jobs:
         with:
           name: serial-${{ env.ARCH }}-${{ env.FLAVOR }}-installer.log
           path: tests/serial.log
+          if-no-files-found: error
+      - name: Upload qemu stdout for installer tests
+        uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: vmstdout-${{ env.ARCH }}-${{ env.FLAVOR }}-installer.log
+          path: tests/vmstdout
           if-no-files-found: error
       - name: Stop test VM
         if: always()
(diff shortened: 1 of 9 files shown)


kkaempf:
What is the "timer sync issue" ? Can you add a reference to the original bug report please ?

kkaempf:
Ah, #1834 has some background

davidcassany:
> Ah, #1834 has some background

Yes sorry, this was just an attempt to verify if this parameter was solving the flaky CI tests. The thing is that this seams to be an issue only appearing in MacOS runners, hence likely to be a problem with hvf and qemu. It is pretty annoying as pretty often we get a kernel panic a boot... 😞 

We haven't ever seen it in any other env beyond MacOS github runners. So we actually need to make a PR to check if the issue is solved.

davidcassany:
Still a draft as it builds on top of https://github.com/rancher-sandbox/ele-testhelpers/pull/38, hence this should be merged first and then readapt go.mod on this project again.

.github/workflows/build_and_test.yaml
```diff
@@ -157,7 +157,6 @@ jobs:
       - if: ${{ env.ARCH == 'x86_64' }} 
         name: Run VM script dependencies
         run: |
-          brew update; brew upgrade qemu
```
frelon, line 160: Yep, this was added since there was an update to qemu which was not picked up by our runner, which broke the build... Probably not needed now!

frelon (approved):
This looks great! :rocket:

.github/workflows/build_and_test.yaml
davidcassany: Yes I guess we were suffering MacOS runner upgrades. I remember there used to be a very old qemu version when we first started using it, now I see it comes with Qemu v8 by default.

scripts/run_vm.sh
```diff
@@ -6,19 +6,20 @@ SCRIPT=$(realpath -s "${0}")
 SCRIPTS_PATH=$(dirname "${SCRIPT}")
 TESTS_PATH=$(realpath -s "${SCRIPTS_PATH}/../tests")
 
+: "${ELMNTL_PREFIX:=}"
```
davidcassany, line 9: I added a prefix to the files so we could consider at some point running in parallel multiple tests on the same machine without overlapping files.

scripts/run_vm.sh
```diff
@@ -67,12 +68,12 @@ function start {
         ;;
   esac
 
-  [ "hvf" == "${ELMNTL_ACCEL}" ] && accel_arg="-accel ${ELMNTL_ACCEL}" && firmware_arg="-bios ${ELMNTL_FIRMWARE} ${firmware_arg}"
+  [ "hvf" == "${ELMNTL_ACCEL}" ] && accel_arg="-accel ${ELMNTL_ACCEL}" && firmware_arg="-bios ${ELMNTL_FIRMWARE} ${firmware_arg}" && cpu_arg="-cpu max,-pdpe1gb"
```
davidcassany, line 71: I found this `-pdpe1gb` flag surfing the net... Apparently this disables hugepages and for some reason this is not supported in MacOS 🤷🏽‍♂️ I do believe this is the real change that fixes flakytests.

make/Makefile.test
```diff
@@ -25,38 +25,41 @@ endif
...
+.PHONY: test-stop
+test-stop:
+	@scripts/run_vm.sh stop
+
 .PHONY: test-installer
 test-installer: prepare-installer-test
-	$(GINKGO) run $(GINKGO_ARGS) ./tests/installer
-	$(GINKGO) run $(GINKGO_ARGS) ./tests/smoke
+	VM_PID=$$(scripts/run_vm.sh vmpid) go run $(GINKGO) $(GINKGO_ARGS) ./tests/installer
+	VM_PID=$$(scripts/run_vm.sh vmpid) go run $(GINKGO) $(GINKGO_ARGS) ./tests/smoke
```
davidcassany, line 38: Probably there are more elegant ways pass the VM pid to the test code, but so far I opted for this simple solution. This is not strictly required and can be omitted on local runs.

PR #1846, opened 2023-10-18
Make use of partx instead of partprobe

davidcassany:
This commit makes use of partx to notify the kernel on partition updates. This fixes the installation over loop devices on some distributions (e.g. Leap 15.5).

### DIFF
--- a/.obs/specfile/elemental-toolkit.spec
+++ b/.obs/specfile/elemental-toolkit.spec
@@ -36,6 +36,8 @@ Requires:       udev
 Requires:       xfsprogs
 Requires:       xorriso
 Requires:       mtools
+Requires:       util-linux
+Requires:       gptfdisk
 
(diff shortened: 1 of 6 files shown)


davidcassany:
I am facing the same or very similar issue as it was reported in #1841 reported while trying to install to a loop device. So far I noted the process does not work on Leap 15.5 hosts, feels like there is something updated in parted or some related tooling. Feels like it doesn't notify the kernel anymore for /dev/loop devices 🤷🏽‍♂️ Interestingly enough I can't see such an issue on my main host (running Leap 15.3, yes I should update it...).

This PR works fine on both Leap 15.3 and Leap 15.5, moreover relies on partx which is part of util-linux. This opens the door to actually completely remove `parted` from our dependency set and use `sgdisk` instead.

frelon (approved):

PR #1851, opened 2023-10-27
Add installer test

davidcassany:
This commit adds an installer test and moves the smoke test to be executed on the installed system rather than the disk build.


frelon (approved):
LGTM!

PR #1852, opened 2023-10-30
Move recovery hostname to cloud-config-defaults

frelon:
Setting recovery hostname is mostly a nice-to-have and can be destructive so move it to cloud-config-defaults feature.


anmazzotti (approved):
Nice!

davidcassany (approved):
LGTM

PR #1853, opened 2023-10-31
Configurable build-disk deploy command

frelon:
This PR allows the deployment-command used when building expandable disk-images to be configured by the user.


davidcassany (approved):
LGTM :+1:

PR #1856, opened 2023-11-09
Add support for RISC-V 64bit architecture

ldevulder:
Basic support of riscv64 architecture. Shim is not supported yet, will be added later.


kkaempf (approved):
lgtm

Dockerfile
```diff
@@ -27,31 +28,13 @@ RUN go build \
...
     -o /usr/bin/elemental
 
-FROM opensuse/leap:$LEAP_VERSION AS elemental
+FROM $BUILD_OS_DISTRI:$BUILD_OS_VERSION AS elemental
 # This helps invalidate the cache on each build so the following steps are really run again getting the latest packages
 # versions, as long as the elemental commit has changed
 ARG ELEMENTAL_COMMIT=""
 ENV ELEMENTAL_COMMIT=${ELEMENTAL_COMMIT}
 
-RUN ARCH=$(uname -m); \
```
frelon, line 36: Not really sure about this change, removing the dependencies from the elemental-toolkit container will break all currently documented approaches of building derivatives (using the toolkit container to run commands).

Dockerfile
```diff
@@ -1,7 +1,8 @@
-ARG GO_VERSION=1.20
-ARG LEAP_VERSION=15.5
+ARG GOLANG_IMAGE
+ARG BUILD_OS_DISTRI
+ARG BUILD_OS_VERSION
 
-FROM golang:${GO_VERSION}-alpine as elemental-bin
```
frelon, line 4: Right now the workflow for building multi-platform images includes `make PLATFORM=linux/amd64,linux/arm64 ... build`. Would probably need to do something akin to https://docs.docker.com/build/building/multi-platform/#cross-compilation to get multi-platform working with this change.

frelon (changes requested):
Really nice work, psyched to see riscv support!

That said, this PR breaks current multi-platform workflow. I would like to be able to write `make PLATFORM=linux/amd64,linux/arm64,linux/riscv64 build` and get a multi-platform image that can be pushed and has all dependencies needed to build other systems.

Dockerfile
ldevulder: This Dockerfile is only to compile the elemental CLI, not one of the provided examples. But yes if the toolkit container is used then it may fails.

I will check the docs and find something else, or use a less minimal images (using BCI images means no `zypper` command).

Dockerfile
ldevulder: I will have a look. The main issue is that RISC-V is not always supported (no official Golang image yet for example).

ldevulder: I modified and split the build: now we can compile the elemental-cli binary on all architecture using cross-compiling.

There is now two dockerfile: one for the binary and another for the toolkit.

Dockerfile
ldevulder: As we discussed I revert this. With the split there is now a `Dockerfile.toolkit` which does this and it works for all architectures.

ldevulder:
Here the grub entry with the latest fixes, no more hardcoded `x86_64` (checked on `x86_64`  and `riscv64`):
```
$ sudo mount -o loop build/elemental-tumbleweed.riscv64.iso /mnt
mount: /mnt: WARNING: source write-protected, mounted read-only.
$ ls /mnt
boot  EFI  rootfs.squashfs
$ ls /mnt/boot/
grub2  riscv64
$ cat /mnt/boot/grub2/grub.cfg
[...]
menuentry "Elemental" --class os --unrestricted {
		echo Loading kernel...
		$linux ($root)/boot/riscv64/loader/linux cdroot root=live:CDLABEL=COS_LIVE rd.live.dir=/ rd.live.squashimg=rootfs.squashfs console=tty1 console=ttyS0 rd.cos.disable cos.setup=/run/initramfs/live/iso-config
		echo Loading initrd...
		$initrd ($root)/boot/riscv64/loader/initrd
	}
[...]
```

ldevulder:
> I would like to be able to write `make PLATFORM=linux/amd64,linux/arm64,linux/riscv64 build`

It should be doable, but keep in mind that qemu docker configuration will be needed to be able to build the full OS stack (or native client on docker).

But the CLI can be compiled for all supported arch directly (thanks to cross-compiling) from `x86_64` (test on my laptop) with the `build-cli-with-docker` target.

davidcassany (approved):
LGTM :+1: Cool too see riscv 

I think this PR should be merged before #1858

Dockerfile.toolkit
```diff
@@ -0,0 +1,35 @@
+# Elemental CLI is built in another Dockerfile that support multi-arch
```
frelon, line 1: I'm not sure about this Dockerfile split, from what I can tell it can be a single Dockerfile that uses --target to split out which image to build.

Makefile
```diff
@@ -10,18 +10,27 @@ PLATFORM?=linux/$(ARCH)
 IMAGE_SIZE?=20G
 PACKER_TARGET?=qemu.elemental-${ARCH}
 REPO?=local/elemental-$(FLAVOR)
+BIN_REPO?=local/elemental-bin
```
frelon, line 13: I'm not sure I see the use case for a new repo, where should this be published and why not use the elemental-toolkit image?

.goreleaser.yaml
```diff
@@ -4,14 +4,15 @@ builds:
...
       - -s -w
-      - -X github.com/rancher/elemental-cli/internal/version.version={{.Tag}}
-      - -X github.com/rancher/elemental-cli/internal/version.gitCommit={{.Commit}}
+      - -X github.com/rancher/elemental-toolkit/internal/version.version={{.Tag}}
+      - -X github.com/rancher/elemental-toolkit/internal/version.gitCommit={{.Commit}}
     goos:
       - linux
     goarch:
       - amd64
       - arm
```
kkaempf, line 13: arm (32-bit) should be dropped imho

ldevulder: I will remove it with the next modifications.

frelon (approved):
Very nice! 👍

.goreleaser.yaml
ldevulder: Done!

Makefile
ldevulder: Removed!

PR #1857, opened 2023-11-15
Remove existing file during build-disk

frelon:
The first commit makes build-disk command behave like build-iso in that it will remove an existing raw disk before writing the new disk and log a warning.

Second commit changes the build-iso Makefile target to use the toolkit image instead of the target derivative image to be consistent with build-disk target

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -44,6 +44,7 @@ jobs:
           username: ${{ secrets.ELEMENTAL_BOT_GITHUB_USERNAME }}
           password: ${{ secrets.ELEMENTAL_BOT_GITHUB_TOKEN }}
       - name: Build toolkit image
+        run: |
           make TOOLKIT_REPO=ghcr.io/${{ github.workspace }}/${{ github.repository }}/elemental-cli VERSION=$(git describe --abbrev=0 --tags) build push-toolkit
           make PLATFORM=linux/arm64,linux/amd64 TOOLKIT_REPO=ghcr.io/${{ github.workspace }}/${{ github.repository }}/elemental-cli VERSION=latest build push-toolkit
       - name: Build green example
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ build-iso: build-os
 	@echo Building ${ARCH} ISO
 	mkdir -p $(ROOT_DIR)/build
 	$(DOCKER) run --rm -v /var/run/docker.sock:/var/run/docker.sock -v $(ROOT_DIR)/build:/build \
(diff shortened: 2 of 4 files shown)


davidcassany (commented):
With the exception of the second commit all good to me.

Regarding the second commit (c8b76f25da5c6734844b552c0262e1e09cdcd788) I have mixed feelings with it. My idea to solve that, was to add a new `RunNoError` (or similar) into the runner interface and in that interface simply log as warnings any error, but never return error. Which would still call it always but do not present it as an error in logs.

Note this error message is also present on install processes, which is also ugly. There the suggested approach here would still log the red ugly error.

So probably I'd be more in favor of log it as a warning|debug message. I have an implementation [here](https://github.com/rancher/elemental-toolkit/commit/ceb4e5250378b3d33627072948d824c0d4b8f474). Probably we can leave that for a separate PR.

frelon:
> Regarding the second commit ([c8b76f2](https://github.com/rancher/elemental-toolkit/commit/c8b76f25da5c6734844b552c0262e1e09cdcd788)) I have mixed feelings with it. My idea to solve that, was to add a new `RunNoError` (or similar) into the runner interface and in that interface simply log as warnings any error, but never return error. Which would still call it always but do not present it as an error in logs.

That feels so hacky though.. We could consider moving the logging out of the runner for this reason, but I do think that only running partx on files in `/dev` makes sense. Will have to try it with the install command also to see what happens!

frelon:
For now I could remove the second commit to get the other changes merged!

davidcassany:
> > Regarding the second commit ([c8b76f2](https://github.com/rancher/elemental-toolkit/commit/c8b76f25da5c6734844b552c0262e1e09cdcd788)) I have mixed feelings with it. My idea to solve that, was to add a new `RunNoError` (or similar) into the runner interface and in that interface simply log as warnings any error, but never return error. Which would still call it always but do not present it as an error in logs.
> 
> That feels so hacky though.. We could consider moving the logging out of the runner for this reason, but I do think that only running partx on files in `/dev` makes sense. Will have to try it with the install command also to see what happens!

Well, I am not convinced either, but this isn't much different than calling `Run` and ignore the error. On exception based languages I'd code `Run` with a flag `RaiseOnError` defaulting to `true`. Anyway as long as I am aware this is only happening for `partx` that I am aware of and I'd like to apply the best effort run concept (swallow errors) for `blkdeactivate` calls, this way we can drop LVM requirement from the spec and simply keep it as a recommended tool. 

Another option could be turning all runner logs into debug or warning level.

Alternatively we could also have a `quiet`  flag we set and unset on the call we want to ignore any kind of message.

I think having the logger within the runner is helpful for debugging purposes as we see the exact commands that are executed.

davidcassany (approved):

PR #1858, opened 2023-11-17
Grub refactor and other distros

davidcassany:


cmd/build-iso.go
```diff
@@ -140,8 +138,6 @@ func NewBuildISO(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 	c.Flags().String("overlay-iso", "", "Path of the overlayed iso data")
 	c.Flags().String("label", "", "Label of the ISO volume")
 	c.Flags().Bool("bootloader-in-rootfs", false, "Fetch ISO bootloader binaries from the rootfs")
-	c.Flags().Var(firmType, "firmware", "Firmware to install for: 'efi' or 'bios'. (defaults to 'efi')")
```
kkaempf, line 143: I wouldn't remove options but just make them output a warning.

examples/blue/Dockerfile
kkaempf, line 1: Shouldn't "blue" (and "orange") be added to CI flavors in order to give both some (minimal) testing ?
Otherwise these examples might rod pretty quickly.

kkaempf (changes requested):
I can't (easily) review a PR with 47(!) changed files

examples/blue/Dockerfile
davidcassany: Yes, the blue is already added to the CI, as well as TW (which wasn't already), orange is not there because network configuration is a must in the CI and I could not easily figure out how to configure network.

davidcassany:
@kkaempf alright, orange flavor is finally added too. Now the test matrix is huge and, unfortunately, the macos environment we use is still flaky a bit, hence with such a big matrix is not rare that one or two tests fail. Retrying makes them pass.

On the good side tests are easily reproducible on the local environment, hence if there is a non environment related issue we can easily reproduce it without the need of github actions.

cmd/build-iso.go
davidcassany: Done, now if a non supported firmware (basically anything but `efi`) is provided it will just error out with a non supported error message. If `--firmware efi` is provided then the deprecated warning appears but it does not error out. `--firmware` option, as it is deprecated it does not appear in help message.

examples/orange/Dockerfile
```diff
@@ -0,0 +1,86 @@
...
+
+# Adding specific network configuration based on netplan
+ADD 05_network.yaml /system/oem/05_network.yaml
+
+# Arrange bootloader binaries into /boot/efi/EFI/elemental 
+# this way elemental installer can easily fetch them
+RUN mkdir -p /boot/efi/EFI/elemental && \
+    cp /usr/lib/grub/x86_64-efi-signed/gcdx64.efi.signed /boot/efi/EFI/elemental/grubx64.efi && \
+    cp /usr/lib/shim/shimx64.efi.signed.latest /boot/efi/EFI/elemental/shimx64.efi && \
+    cp /usr/lib/shim/mmx64.efi /boot/efi/EFI/elemental/mmx64.efi
```
davidcassany, line 83: @frelon in this section I mostly exercise this concept of a custom path (`/boot/efi/EFI/elemental`) to arrange the bootloader binaries in a way that elemental installer will find them. However I not convinced about this path `/boot/efi/EFI/elemental`, as `/boot/efi` is usually the mountpoint of EFI partition in many distributions, so I am wondering if shouldn't we use something like `/etc/elemental/bootloader` or something else. Do you have any better idea? The more I think about `/boot/efi/EFI/elemental` the more I dislike it.


Probably a structure like like:


```
/etc/elemental/config.yaml
/etc/elemental/config.d/extra_config.yaml
/etc/elemental/bootloader/grub.cfg
/etc/elemental/bootloader/bootargs.cfg
/etc/elemental/bootloader/shim.efi
/etc/elemental/bootloader/grub.efi
/etc/elemental/bootloader/mmx64.efi
...
```

frelon: Yep, I agree that `/boot/efi` is not ideal! It seems at least tumbleweed puts some bootloader stuff in `/usr/lib/bootloader`.. Maybe we could use something like `/usr/lib/bootloader/elemental` for efi apps and bootloader config?

frelon: I'm not opposed to using `/etc/elemental/bootloader` either!

davidcassany: interesting, just checked the `/usr/lib/bootloader` is owned by the package perl-Bootloader (perl utilities for various bootloaders). I don't want to get mixed with this. But you got a point, probably we should use something like `/usr/lib/elemental/bootloader`. My resistance of `/etc/elemental` is that it feels odd adding binaries in there, `/etc` should include only configuration files.

What about:

```
/etc/elemental/config.yaml
/etc/elemental/config.d/extra_config.yaml
/etc/elemental/bootloader/grub.cfg
/etc/elemental/bootloader/bootargs.cfg
/usr/lib/elemental/bootloader/shim.efi
/usr/lib/elemental/bootloader/grub.efi
/usr/lib/elemental/bootloader/mmx64.efi
```

Anyway we can run this change on a follow up PR.

frelon: Looks reasonable to me!

davidcassany:
Adding blocked label as I think we should sort #1856 first, as the two PR conflict and I do believe it is easier to rebase this PR than the other way round.

examples/orange/Dockerfile
davidcassany: Finally used `/usr/lib/elemental/bootloader` after rebasing over the riscv64 support.

davidcassany:
rebased on top of #1856 and #1859

kkaempf (approved):
Unreviewable with 52 changed files :-(

davidcassany:
This PR has been split in two, first the refactor itself, and then additional examples of other distros. This way the PR is slightly better to manage and also ends up in more organized and better git history.

Refactor PR #1867

frelon (approved):

PR #1859, opened 2023-11-21
Add target for rpi disk

frelon:
This commit adds a flavor for tumbleweed raspberry pi image that uses a after-disk hook to copy firmware into the EFI partition.


Dockerfile
```diff
@@ -21,7 +21,8 @@ ARG ELEMENTAL_VERSION=0.0.1
 ARG ELEMENTAL_COMMIT=""
 ENV ELEMENTAL_VERSION=${ELEMENTAL_VERSION}
 ENV ELEMENTAL_COMMIT=${ELEMENTAL_COMMIT}
-RUN go build \
+ARG TARGETOS TARGETARCH
```
kkaempf, line 24: I'd split this into two lines for readability

Makefile
```diff
@@ -48,7 +48,7 @@ build-cli:
 .PHONY: build-os
 build-os: build
 	$(DOCKER) build examples/$(FLAVOR) --platform ${PLATFORM} ${DOCKER_ARGS} --build-arg TOOLKIT_REPO=$(TOOLKIT_REPO) --build-arg VERSION=$(VERSION) --build-arg REPO=$(REPO) -t $(REPO):$(VERSION)
-
+
```
kkaempf, line 51: Oops

Makefile
```diff
@@ -71,12 +71,25 @@ build-disk: build-os
 	mkdir -p $(ROOT_DIR)/build
 	$(DOCKER) run --rm -v /var/run/docker.sock:/var/run/docker.sock -v $(ROOT_DIR)/build:/build \
 		--entrypoint /usr/bin/elemental \
-		${TOOLKIT_REPO}:${VERSION} --debug build-disk --unprivileged --expandable -n elemental-$(FLAVOR).$(ARCH) --local \
+		${TOOLKIT_REPO}:${VERSION} --debug build-disk --platform $(PLATFORM) --unprivileged --expandable -n elemental-$(FLAVOR).$(ARCH) --local \
```
kkaempf, line 74: Uhm, we already have TARGETOS and TARGETARCH - and PLATFORM seems to be a combination of both ?!

Makefile
```diff
@@ -71,12 +71,25 @@ build-disk: build-os
...
 
+.PHONY: build-rpi-disk
+build-rpi-disk: build-os
+ifneq ("$(PLATFORM)","linux/arm64")
+	@echo "Cannot build raspberry pi disk for $PLATFORM"
+	@exit 1
+endif
+	@echo Building ${ARCH} disk
+	mkdir -p $(ROOT_DIR)/build
+	$(DOCKER) run --rm -v /var/run/docker.sock:/var/run/docker.sock -v $(ROOT_DIR)/examples:/examples -v $(ROOT_DIR)/build:/build \
```
kkaempf, line 88: Would that (/var/run/docker.sock) work with podman ?

Makefile
```diff
@@ -71,12 +71,25 @@ build-disk: build-os
...
+		--entrypoint /usr/bin/elemental \
+		${TOOLKIT_REPO}:${VERSION} --debug build-disk --platform $(PLATFORM) --cloud-init-paths /examples/tumbleweed-rpi --unprivileged --expandable -n elemental-$(FLAVOR).aarch64 --local \
```
kkaempf, line 90: Please start with SLE Micro for Rancher, that's our primary target. tumbleweed is secondary

Makefile
frelon: Hehe, TARGETOS and TARGETARCH are set in Dockerfile based on what platform is passed in.. Generally what we use is `--platform` but the fun begins when we are doing cross-compilation in Dockerfiles :see_no_evil:

Makefile
frelon: Looks like you can use `/run/podman/podman.sock`.. Will probably try to hide this behind a variable!

cmd/build-disk.go
```diff
@@ -108,7 +108,8 @@ func NewBuildDisk(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 	c.Flags().Bool("expandable", false, "Creates an expandable image including only the recovery image")
 	c.Flags().Bool("unprivileged", false, "Makes a build runnable within a non-privileged container, avoids mounting filesystems (experimental)")
 	c.Flags().VarP(imgType, "type", "t", "Type of image to create")
-	c.Flags().StringSliceP("cloud-init", "c", []string{}, "Cloud-init config files")
+	c.Flags().StringSliceP("cloud-init", "c", []string{}, "Cloud-init config files to include in disk")
+	c.Flags().StringSlice("cloud-init-paths", []string{}, "Cloud-init config files to run during build")
```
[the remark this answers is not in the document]
frelon: Sure, I'm open to suggestions! Maybe `build-hook-paths` or just `build-hooks`..

Makefile
frelon: We should probably add teal back as a flavor then, or use slem4r.. as long as we don't start publishing the artifacts again it should be fine!

davidcassany: What about moving the example to leap 15.5? Leap and SLE share OS essentials and SLE and SLE Micro share them too, we should get the same exact versions of all linux-utils, core-utils, systemd, grub, dracut, etc.

cmd/build-disk.go
```diff
@@ -105,7 +105,8 @@ func NewBuildDisk(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 	c.Flags().Bool("expandable", false, "Creates an expandable image including only the recovery image")
 	c.Flags().Bool("unprivileged", false, "Makes a build runnable within a non-privileged container, avoids mounting filesystems (experimental)")
 	c.Flags().VarP(imgType, "type", "t", "Type of image to create")
-	c.Flags().StringSliceP("cloud-init", "c", []string{}, "Cloud-init config files")
+	c.Flags().StringSliceP("cloud-init", "c", []string{}, "Cloud-init config files to include in disk")
+	c.Flags().StringSlice("build-hooks", []string{}, "Paths to cloud-init config files to run during build")
```
davidcassany, line 109: This rename is not enough, it should be adapted within the `pkg/types/v1/config.go` at `DiskSpec` definition too.

Note for later.  I am also wondering if shouldn't we make this change consistent to all commands and then have `install-hooks`, `upgrade-hooks`, `reset-hooks` flags.

davidcassany (changes requested):
I'd leave the `build-hooks` flag change out of this PR and eventually make a follow up PR to better manage and handle this use case including build, install, upgrade and reset commands or any command that makes use of the hooks concept.

cmd/build-disk.go
frelon: Yep, I brought back the `cloud-init-paths` for now since it is needed for this PR to produce a bootable rpi disk.

I also think we should rework the hooks in a follow up PR :+1:

Makefile
frelon: I would also vote for leap, I remember having trouble accessing the sle-micro repositories when building an early version of teal.

davidcassany (approved):
Awesome :+1:

PR #1866, opened 2023-11-29
Refactor early system mounting

frelon:
An updated mount command (from #1837).

Adds a new feature/dracut-module `elemental-sysroot` responsible for mounting the correct image as a base.
The `elemental-rootfs` feature actually mounts the elemental overlays.

This also reworks the kernel parameters used by elemental.

* elemental.image=active|passive|recovery
* elemental.disable

And also starts to move the active|passive|recovery_mode files to /run/elemental with all the overlay and other mounts.

Fixes https://github.com/rancher/elemental-toolkit/issues/1781, https://github.com/rancher/elemental-toolkit/issues/1830


kkaempf (approved):
This is nearly impossible to review. 😞 

Approving anyways - if it fails, you're the one to fix it 😉

config.yaml.example
```diff
@@ -120,6 +120,29 @@ upgrade:
...
+  write-fstab: true # Write fstab into sysroot/etc/fstab
+  overlay:
+    type: tmpfs # tmpfs|block
+    device: /dev/sda6 # Block device used to store overlay. Used when type is set to block
+    size: 25% # Size of tmpfs as percentag of system memory. Used when type is set to tmpfs
+    paths:
+      - /var
+      - /etc
+      - /srv
+  persistent:
```
davidcassany, line 135: Where is the persistent block device defined? I think this should be configurable and probably default to our persistent partition so it can be omitted from the config file, but still configurable.

config.yaml.example
```diff
@@ -120,6 +120,29 @@ upgrade:
   # grub menu entry, this is the string that will be displayed
   grub-entry-name: Elemental
 
+# configuration used for the 'mount' command
+mount:
+  sysroot: /sysroot # Path to mount system to
+  write-fstab: true # Write fstab into sysroot/etc/fstab
+  overlay:
```
davidcassany, line 127: I am wondering if this is the right name. I mean if we have `persistent` then probably we should have something like `transient` or `ephemeral`. Otherwise one could think this overlay device setting is for the persistent data.

pkg/features/embedded/cloud-config-essentials/system/oem/00_rootfs.yaml
```diff
@@ -19,14 +19,14 @@ stages:
...
-    - if: '[ ! -f "/run/cos/recovery_mode" ]'
+    - if: '[ ! -f "/run/elemental/recovery_mode" ]'
       name: "Persist /etc/machine-id"
       commands:
       - |
         # persist machine-id
-        if [ -s /usr/local/etc/machine-id ]; then
-          cat /usr/local/etc/machine-id > /etc/machine-id
+        if [ -s /run/elemental/persistent/etc/machine-id ]; then
+          cat /run/elemental/persistent/etc/machine-id > /etc/machine-id
```
davidcassany, line 28: I assume this is because the persistent partition is no longer mounted at `/usr/local` by default, right?

pkg/features/embedded/elemental-rootfs/usr/lib/dracut/modules.d/30elemental-rootfs/module-setup.sh
```diff
@@ -0,0 +1,43 @@
...
+    declare systemdsystemunitdir=${systemdsystemunitdir}
+
+    inst_multiple \
+        mount mountpoint sort rmdir findmnt rsync cut realpath basename lsblk
+
+    inst_simple "/etc/elemental/config.yaml"
+
+    # Include utilities required for elemental-setup services,
+    # probably a devoted dracut module makes sense
+    inst_multiple -o \
```
davidcassany, line 33: I think this list needs to be revisited, most of this tools are not needed at all. It should only include the tools required for the elemental-rootfs.service, probably none.

pkg/features/embedded/elemental-rootfs/usr/lib/dracut/modules.d/30elemental-rootfs/module-setup.sh
```diff
@@ -0,0 +1,43 @@
...
+}
+
+# called by dracut
+install() {
+    declare moddir=${moddir}
+    declare systemdutildir=${systemdutildir}
+    declare systemdsystemunitdir=${systemdsystemunitdir}
+
+    inst_multiple \
+        mount mountpoint sort rmdir findmnt rsync cut realpath basename lsblk
```
davidcassany, line 27: same as below, this list feels outdated. Be aware that these might be required to be included to some other module then.

davidcassany (approved):
Looks good to me

Just some minor comments here and there. Little things I am also fine to address them in follow up PRs, so we make further changes more manageable and reviewable.

pkg/features/embedded/cloud-config-essentials/system/oem/00_rootfs.yaml
frelon: Yes, that is correct

PR #1868, opened 2023-11-30
Add other example distros

davidcassany:
This PR adds a Fedora based image (blue) and an Ubuntu based image (orange)

### DIFF
--- a/examples/orange/Dockerfile
+++ b/examples/orange/Dockerfile
@@ -4,7 +4,7 @@ ARG VERSION
 FROM ${TOOLKIT_REPO}:${VERSION} as TOOLKIT
 
 # OS base image of our choice
-FROM ubuntu:22.04 as OS
+FROM ubuntu:23.04 as OS
 ARG REPO
 ARG VERSION
 ENV VERSION=${VERSION}
@@ -25,6 +25,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-ins
   systemd \
   systemd-sysv \
   systemd-timesyncd \
+  systemd-resolved \
   openssh-server \
   openssh-client \
   tzdata \
@@ -47,6 +48,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-ins
   dbus-daemon \
   patch \
   netplan.io \
+  locales \
   xz-utils
 
 # Create non FHS paths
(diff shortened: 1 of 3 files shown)


frelon:
Awesome work 🚀

davidcassany:
Ok this last commit makes the CI to react on `arm64` and `all-distros` labels. If `arm64` is added it will trigger, in addition, tests for `aarch64` builds. If the label `all-distros` is included it will trigger all examples in the CI. If both labels are present only SUSE distributions are tested on aarch64 (no `blue` nor `orange`).

Full suite of tests is only run for `green` (openSUSE Leap) flavor.

.github/workflows/pr.yaml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - main
   pull_request:
+    types: [ unlabeled, labeled, opened, synchronize, reopened ]
```
kkaempf, line 7: Does this run on PRs without any labels now ? 😨

tests/smoke/smoke_test.go
```diff
@@ -100,11 +100,12 @@ var _ = Describe("Elemental Smoke tests", func() {
 			Expect(out).Should(ContainSubstring("UTC"))
 		})
 
-		It("has default localectl configuration from cloud-init", func() {
+		// locale setting doesn't work on Fedora ¯\_(ツ)_/¯
+		/*It("has default localectl configuration from cloud-init", func() {
```
kkaempf, line 104: So a bug in Fedora makes us disable this test for all other distros ? 🤔

.github/workflows/pr.yaml
davidcassany: yes, these are the events that actually trigger the run. So on opening, synchronizing, labeling, unlabeling, etc.

tests/smoke/smoke_test.go
davidcassany: Not sure if a bug in Fedora... but yes that can be improved

davidcassany: Re-enabled

.github/workflows/pr.yaml
davidcassany: Without any label runs the exact same thing that used to run, but this PR adds the functionality that when the `all-distro` label is around in addition it also runs the CI for other distros. Running `arm64` tests by including the `arm64` label was already possible. 

The change here on this line only express under which events the workflow is re-triggered. The default value is `opened`, `synchronize` and  `reopened`. Then I added `labeled` and `unlabeled` so it also reacts on label changes and there is no need to add the desired labels before hand at PR creation.

I think probably a better approach would be moving the `arm64` tests out of the PR scope and run them only on demand by manual workflow and/or based on some schedule, but not on PRs.

PR #1869, opened 2023-12-01
Fix initrd and kernel ISO paths

davidcassany:
This changes the kernel path from `/boot/<arch>/loaderlinux` to `/boot/<arch>/loader/linux`. An equivalent change is applied on initrd path.

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -313,10 +313,10 @@ func ISOLoaderPath(arch string) string {
 
 // ISOKernelPath returns path use to store the kernel
 func ISOKernelPath(arch string) string {
-	return ISOLoaderPath(arch) + "linux"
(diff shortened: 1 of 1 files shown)


pkg/constants/constants.go
```diff
@@ -313,10 +313,10 @@ func ISOLoaderPath(arch string) string {
 
 // ISOKernelPath returns path use to store the kernel
 func ISOKernelPath(arch string) string {
-	return ISOLoaderPath(arch) + "linux"
+	return filepath.Join(ISOLoaderPath(arch), "linux")
```
davidcassany, line 316: @ldevulder note the arch is passed as an argument instead of using runtime architecture to enable cross arch builds. We might be looking to build arm64 images in a x86_64 host, I think we are so close to make that possible.

PR #1870, opened 2023-12-01
Log errors of runner commands only in debug mode

davidcassany:
During any installation we currently have this logs

```
INFO[2023-12-01T12:14:42Z] Partitioning device...                       
DEBU[2023-12-01T12:14:42Z] Running cmd: 'parted --script --machine -- /dev/vda unit s mklabel gpt' 
DEBU[2023-12-01T12:14:42Z] Running cmd: 'partx -u /dev/vda'             
ERRO[2023-12-01T12:14:42Z] Error running command: exit status 1
```
Where the error message is likely to appear in clearly visible red red. This particular error happens because after running any partition related change in disk run `partx` to reload and update partitions. This is only a best effort call, meaning we ignore any error (if running in a container this is likely to always fail, but doesn't necessarily mean there is an installation error).
Given the fact there a few command calls for which we ignore reported errors it doesn't make much sense having the `runner.Run` method to log at error level all command failures, this is caller responsibility to know if this should be logged at error level or not. 

Hence in this PR on `runner.Run` errors we actually log everything (command error and command output) at debug level. Since the method returns the error and the command output, the caller still has the chance to log error as error level if this is the desired behavior.

### DIFF
--- a/pkg/types/v1/runner.go
+++ b/pkg/types/v1/runner.go
@@ -53,7 +53,8 @@ func (r RealRunner) Run(command string, args ...string) ([]byte, error) {
 	cmd := r.InitCmd(command, args...)
 	out, err := r.RunCmd(cmd)
 	if err != nil {
-		r.error(fmt.Sprintf("Error running command: %s", err.Error()))
+		r.debug(fmt.Sprintf("'%s' command reported an error: %s", command, err.Error()))
+		r.debug(fmt.Sprintf("'%s' command output: %s", command, out))
 	}
 	return out, err
 }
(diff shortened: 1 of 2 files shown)


davidcassany:
@frelon this is an approach to what we discussed a while ago to prevent having misleading log messages. Probably we can do some more stuff to improve this particular case, but in any case the change suggested here does not harm in any way (just moving logs from error level to debug level) and actually makes sense from a wider perspective.

Then we can still consider calling `partx` over block devices only in partitioner wrapper. Probably by checking the special bit rather than relaying only on `/dev` path? Not sure, opened to any suggestion and idea.

frelon (approved):
Makes sense, then we can log an error outside of the Runner if appropriate :+1:

PR #1871, opened 2023-12-05
Fix RISCV64 Tumbleweed example build

ldevulder:
This allow to build the Tumbleweed example with `Makefile`.


kkaempf (approved):
We're building against a moving target here.
Not ideal, but also without any alternatives 🤷🏻

ldevulder:
> We're building against a moving target here.
> Not ideal, but also without any alternatives 🤷🏻

Yes, we have to wait until RISCV64 is added to Factory/Tumbleweed.

Makefile
```diff
@@ -8,7 +8,7 @@ FLAVOR?=green
 ARCH?=$(shell uname -m)
 PLATFORM?=linux/$(ARCH)
 IMAGE_SIZE?=20G
-PACKER_TARGET?=qemu.elemental-${ARCH}
+PACKER_TARGET?=qemu.elemental-$(ARCH)
```
frelon, line 11: This variable can be removed instead.

Makefile
```diff
@@ -19,17 +19,22 @@ DOCKER_SOCK?=/var/run/docker.sock
 GIT_COMMIT?=$(shell git rev-parse HEAD)
 GIT_COMMIT_SHORT?=$(shell git rev-parse --short HEAD)
 GIT_TAG?=$(shell git describe --candidates=50 --abbrev=0 --tags 2>/dev/null || echo "v0.0.1" )
-VERSION?=${GIT_TAG}-g${GIT_COMMIT_SHORT}
+VERSION?=$(GIT_TAG)-g$(GIT_COMMIT_SHORT)
```
frelon, line 22: What's the reason for changing `{}` to `()` ? In my eyes `{}` is used for variable references and `()` for running commands and assigning the output.

ldevulder: In bash yes but in Makefile `()` can be used too (see [here](https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_6.html)). As far as I remember in Makefile syntax it is used for variables declared inside the Makefile, which is the case for most of them (more all here). `{}` works too in GNU Makefile (not sure in non-GNU version, but we don't use it I know).

In our case we already had both, the idea here was mainly to be coherent and use the same for all vars, and use the initial Makefile syntax.

davidcassany (approved):
LGTM

Issue #1874, opened 2023-12-11
Refactor to snapshotter interface (including loopdevices implementation)


davidcassany:
This is to include the snapshotter concept in elemental-toolkit in some sort of _backward compatible_ (probably fully backward compatible but almost) fashion.

Part of #1863

PR #1875, opened 2023-12-11
Wrap mount.Interface to our own interface

davidcassany:
This sets the actual mount interface we need in elemental and wraps `k8s.io/mount-utils` use in a single place. This makes our mounter interface way easier to substitute by another implementation if needed at some point. It also makes evident we need a simple mount interface.

Part of #1874


frelon (approved):
LGTM!

PR #1879, opened 2023-12-13
Add backwards-compatibility for liveCD detection

frelon:
This commit adds back the /run/cos/live_mode file if booting in live-mode. /run/elemental/live_mode will exist side-by-side while /run/cos is removed.


davidcassany (approved):
lgtm

PR #1880, opened 2023-12-13
Rename overlay -> ephemeral mounts

frelon:
This change aims to make the distinction between ephemeral, persistent, overlay and bind mounts a bit more clear.

The system mounts ephemeral and persistent directories on boot. A persistent mount can be materialized using a linux overlay-mount or a bind-mount.


davidcassany (approved):
LGTM

PR #1882, opened 2023-12-13
Cleanup caches when PR is closed

frelon:
Add a workflow that will run when a PR is closed that cleans up all caches created by the PR.


davidcassany (approved):
Very nice!

ldevulder (approved):
Very nice!

Also something we can re-use to clean the cache on Elemental CI, I will look at it.

PR #1883, opened 2023-12-13
Refactor some elemental package methods to be used as stand alone functions

davidcassany:
This PR refactors some methods in elemental package. Basically it turns some of the methods in to _static_ functions not tied to any object.

This is to allow reusing them without the need of an Elemental object. In fact I am also wondering of running a follow up PR completely removing the Elemental object from https://github.com/rancher/elemental-toolkit/blob/a7b7f2ea402359b952b4acae2bd89ec41f50d0c7/pkg/elemental/elemental.go#L33-L35
as this is a useless object. In only keeps the configuration which is anyway already available in all places where the elemental package is in use. So in a follow up PR I would even consider dropping the object entirely and simply pass the configuration struct as a parameter.

Part of #1874

### DIFF
--- a/pkg/action/build-disk.go
+++ b/pkg/action/build-disk.go
@@ -153,7 +153,7 @@ func (b *BuildDiskAction) BuildDiskRun() (err error) { //nolint:gocyclo
 	activeRoot := recRoot
 
 	// Create recovery root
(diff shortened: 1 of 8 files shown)


frelon (approved):
LGTM

PR #1885, opened 2023-12-13
Load mount-layout from .env-files

frelon:
This PR adds code to load environment variables OVERLAY, RW_PATHS, PERSISTENT_STATE_PATHS and PERSISTENT_STATE_BINDS during mounting.

The variables are read from /run/cos/cos-layout.env and /run/elemental/mount-layout.env if they exist.


cmd/config/config.go
```diff
@@ -275,6 +275,11 @@ func ReadMountSpec(r *v1.RunConfig, flags *pflag.FlagSet) (*v1.MountSpec, error)
 		return mount, err
 	}
 
+	if err := applyMountEnvVars(r, mount); err != nil {
```
davidcassany, line 278: Not 1005 sure, but shouldn't this call happen before the `applyKernelCmdline`? this way kernel parameters have priority. I think it was done this way before.

davidcassany (approved):
Looks good to me.

Just a small comment

cmd/config/config.go
frelon: Good point! It's hard to tell in the old code but I will try it on a fresh system and see which takes precedence!

frelon: Hmm, it looks like the values in `/run/cos/cos-layout.env` have been picked over kernel-cmdline previously. Right now it's only actually used for the ephemeral overlay settings so I'm inclined to use the kernel-cmdline if it exists.. WDYT?

davidcassany: Yes I agree :+1:

PR #1886, opened 2023-12-14
Provide early boot systemd unit files within the associated dracut module

davidcassany:


pkg/features/embedded/elemental-setup/etc/dracut.conf.d/02-elemental-setup.conf
```diff
@@ -1,4 +1,2 @@
-install_items+=" /usr/lib/systemd/system/elemental-setup-initramfs.service /etc/systemd/system/initrd.target.requires/elemental-setup-initramfs.service "
-install_items+=" /usr/lib/systemd/system/elemental-setup-rootfs.service /etc/systemd/system/initrd-fs.target.requires/elemental-setup-rootfs.service "
```
davidcassany, line 2: I saw dracut error messages trying to execute those, that how all this PR started.

pkg/features/embedded/elemental-rootfs/etc/dracut.conf.d/02-elemental-rootfs.conf
```diff
@@ -1,2 +1 @@
-install_items+=" /usr/lib/systemd/system/elemental-rootfs.service /etc/systemd/system/initrd-fs.target.requires/elemental-rootfs.service "
```
davidcassany, line 1: This is redundant with the lines from https://github.com/rancher/elemental-toolkit/blob/48966fa3e341d0b8b036fde7c27f3ea55d9cc747/pkg/features/embedded/elemental-rootfs/usr/lib/dracut/modules.d/30elemental-rootfs/module-setup.sh#L36-L40

davidcassany (commented):
https://github.com/rancher/elemental-toolkit/blob/48966fa3e341d0b8b036fde7c27f3ea55d9cc747/pkg/features/embedded/elemental-setup/usr/lib/dracut/modules.d/10elemental-setup/module-setup.sh#L27-L31

`${moddir}` relates to the root folder of the dracut module, this is the folder of this current file, hence this path was essentially wrong.

pkg/features/embedded/elemental-rootfs/etc/dracut.conf.d/02-elemental-rootfs.conf
frelon: Yep, I think it was just a copy-paste that stuck around from the `elemental-setup` feature

pkg/features/features.go
```diff
@@ -146,11 +146,7 @@ func Get(names []string) ([]*Feature, error) {
 				return features, fmt.Errorf("Conflicting features: %s and %s", FeatureElementalRootfs, FeatureImmutableRootfs)
 			}
 
-			units := []*systemd.Unit{
-				systemd.NewUnit("elemental-rootfs.service"),
```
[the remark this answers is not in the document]
davidcassany: good point I am not sure if this will still be de case. I mean the unit is there but mostly managed by the initramfs systemd unit.

The thing is that I to make `inst_simple` command in `module-setup.sh` functional having the service file within the module makes it easy. Probably we could use the full path in there as long as the unit file and the dracut module are in the same feature it shouldn't be a problem having a hardcoded full path in there. I can give it a quick test.

frelon (approved):
LGTM!

PR #1887, opened 2023-12-14
Changes to cleanup-cache workflow

frelon:
Tested this in https://github.com/frelon/cache-test/actions/runs/7209527181/job/19640659692 and seems to work when permissions are added for the whole workflow and not only jobs.


davidcassany (approved):
LGTM

PR #1888, opened 2023-12-14
Add disable-boot-entry flag to reset-command

frelon:
Read the disable-boot-entry flag from ELEMENTAL_RESET_DISABLE_BOOT_ENTRY environment variable


davidcassany (approved):
LGTM

Out of curiosity, in which context you needed that? Or you just saw it was missing on the list?

frelon:
> LGTM
> 
> 
> 
> Out of curiosity, in which context you needed that? Or you just saw it was missing on the list?

I found this when trying to build an rpi image in the SeedImage where the elemental-register command passes flags using environment variables 👍

PR #1890, opened 2023-12-18
Adding new snapshotter interface

davidcassany:
This PR defines the snapshotter interface and snapshotter configuration structures. I does not provide any implementation of the interface it just provides all the required pieces to read the snapshotter configuration:

```yaml
snapshotter:
  type: loopdevice
  max-snaps: 4
  config:
    fs: ext2
    size: 0
 ```
 
 Config is defined internally as a `interface{}` and the format and schema of `config` depends on the `type` value. This PR essentially provides the structures definition and the logic to unmarshall `config` depending on `types` value.
 
In a follow up PR I expect to include the loopdevice implementation and its unit tests (but keeping the interface unused still). Finally in third of forth PR I expect to include the full refactor to actually start using the interface in `install`, `reset` and `upgrade`, commands.  All this is an effort to make PRs easier to follow and review.

Part of #1874

### DIFF
--- a/pkg/snapshotter/loopdevice.go
+++ b/pkg/snapshotter/loopdevice.go
@@ -1,0 +1,77 @@
+/*
+Copyright © 2022 - 2023 SUSE LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package snapshotter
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: pkg/snapshotter/loopdevice.go [hunk 1]
```diff
@@ -0,0 +1,77 @@
...
+)
+
+var _ v1.Snapshotter = (*LoopDevice)(nil)
+
+type LoopDevice struct {
+	cfg            v1.Config
+	snapshotterCfg v1.SnapshotterConfig
+	loopDevCfg     v1.LoopDeviceConfig
+	rootDir        string
+	/*currentSnapshotID int
```
frelon, line 43: Should these fields be kept?

frelon (approved):
LGTM

pkg/snapshotter/loopdevice.go
davidcassany: They are commented because the linter complained about being unused 🤷🏽‍♂️ They are back and being used on on the next PR I am preparing with the full implemenentation of this `loopdevice.go`, currently writing unit tests.

PR #1892, opened 2023-12-20
Split arm and x86_64 tests

davidcassany:
This commit splits arm64 and x86_64 tests. With this PR arm64 tests on self hosted runners can only be executed on demand, they are no longer possible in PRs.

Later on we can consider making them executable on scheduled basis if we consider them to be stable enough.


davidcassany:
Once approved I will update the required build checks, as they got renamed due to the matrix change and now expected ones are simply not happening.

frelon (approved):

PR #1893, opened 2023-12-22
Add default cloud-init-paths to build commands

frelon:
This will run build-hooks from `/oem`, `/system/oem` and `/usr/local/cloud-config` by default.


kkaempf (approved):
🤞🏻

PR #1894, opened 2024-01-05
Update github.com/go-git/go-git to v5.11.0

frelon:


anmazzotti (approved):
I wonder what is the reason for the change since go-git is not a direct dependency.

However this looks fine.

frelon:
> I wonder what is the reason for the change since go-git is not a direct dependency.
> 
> However this looks fine.

The reason was this [dependabot alert](https://github.com/rancher/elemental-toolkit/security/dependabot/54).

It's probably not something we would run into, but better safe than sorry :+1:

PR #1898, opened 2024-01-05
Assign elemental-setup*service files to elemental-toolkit

### LINKED ISSUE #1896: None of the elemental-setup-*.service files  are owned by a package
Running
~~~
rpm -qf /usr/lib/systemd/system/elemental-setup-*
~~~
shows `is not owned by any package` for all of them.

kkaempf:
Fixes #1896


anmazzotti (approved):

frelon (approved):
Nice :+1:

davidcassany:
IMHO we do not need it anymore as we already moved the services from `/usr` to `/etc` where user provided data is also expected to appear. To my understanding it could be unexpected having data under `/usr` (with `/usr/local` as the exception) that is not owned by any package.

PR #1900, opened 2024-01-09
Change default rootfs feature

frelon:
Use the new elemental-rootfs and elemental-sysroot instead of immutable-rootfs.


kkaempf (approved):
🤞🏻

Issue #1848, opened 2023-10-23
Move grub configuration out of state partition


frelon:
As discussed in https://github.com/rancher/elemental-toolkit/pull/1794#issuecomment-1757044517 we could move the grub configuration out of state partition.

This would mean we could skip adding the state partition for expandable images built using `build-disk` and also resize the recovery-image, resulting in a huge saving in space for disk images.

We can also consider adding a `--include-bootloader` (or any other understandable name) to reinstall grub during an upgrade procedure. Note this could potentially break the capability to boot a fallback image.

kkaempf:
👍🏻 for smaller images

frelon:
Upgrading the bootloader to be merged in #1906 closing this as done

PR #1903, opened 2024-01-10
Move grub-config to /etc/elemental

frelon:
Moves `grub.cfg` from `/etc/cos` to `/etc/elemental` in the source image.

Part of #1848


davidcassany (approved):
LGTM

PR #1904, opened 2024-01-11
Install grub.cfg into EFI System Partition

frelon:
Change the bootloader install logic to install the grub.cfg into the EFI System Partition (ESP).

This needs some changes to how root is set in the grub.cfg as well as moving grub_oem_env file to ESP when created in install/reset/upgrade command.

This also moves the grub modules to the ESP.

Part of #1848

### DIFF
--- a/pkg/bootloader/grub.go
+++ b/pkg/bootloader/grub.go
@@ -413,8 +413,9 @@ func (g *Grub) Install(rootDir, bootDir, deviceLabel string) (err error) {
 	return g.InstallConfig(rootDir, bootDir)
 }
 
-// InstallConfig installs grub configuraton files to the expected location.  rootDir is the root
-// of the OS image, bootDir is the folder grub read the configuration from, usually state partition mountpoint
+// InstallConfig installs grub configuraton files to the expected location.
+// rootDir is the root of the OS image, bootDir is the folder grub read the
+// configuration from, usually EFI partition mountpoint
 func (g Grub) InstallConfig(rootDir, bootDir string) error {
 	for _, path := range []string{constants.FallbackEFIPath, constants.EntryEFIPath} {
 		grubFile := filepath.Join(rootDir, g.elementalCfg)
--- a/pkg/bootloader/grub.go
+++ b/pkg/bootloader/grub.go
@@ -32,8 +32,11 @@ import (
 )
 
 const (
-	defaultGrubPrefix = "/EFI/BOOT"
-	grubCfgFile       = "grub.cfg"
+	grubCfgFile = "grub.cfg"
+)
+
+var (
+	defaultGrubPrefixes = []string{"/EFI/ELEMENTAL", "/EFI/BOOT"}
 )
 
 func getGModulePatterns(module string) []string {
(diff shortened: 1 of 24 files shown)


pkg/bootloader/grub_test.go
```diff
@@ -111,15 +112,21 @@ var _ = Describe("Booloader", Label("bootloader", "grub"), func() {
...
 		Expect(data).To(Equal(grubCfg))
-		_, err = fs.Stat(fmt.Sprintf("%s/grub2/x86_64-efi/loopback.mod", bootDir))
+
+		data, err = fs.ReadFile(filepath.Join(bootDir, "/EFI/ELEMENTAL/grub.cfg"))
 		Expect(err).To(BeNil())
-		_, err = fs.Stat(fmt.Sprintf("%s/grub2/x86_64-efi/xzio.mod", bootDir))
+		Expect(data).To(Equal(grubCfg))
+
+		// Check everything is copied in boot directory
+		_, err = fs.Stat(filepath.Join(bootDir, "x86_64-efi/loopback.mod"))
```
kkaempf, line 125: not sure about hardcoded architectures 🤔

frelon: Hmm, I think it's fine since it's only for the tests. There was no platform set though, so the tests would fail when ran on another architecture. I added linux/amd64 as the default platform now and in the future we could consider running these tests for all supported platforms. (I see there is already a few tests for arm64 and risc-v as well).

kkaempf (approved):
🤞🏻

pkg/bootloader/grub.go
```diff
@@ -402,8 +393,8 @@ func (g *Grub) SetDefaultEntry(partMountPoint, imgMountPoint, defaultEntry strin
 }
 
 // Install installs grub into the device, copy the config file and add any extra TTY to grub
-func (g *Grub) Install(rootDir, bootDir, stateLabel string) (err error) {
-	err = g.InstallEFI(rootDir, bootDir, constants.EfiDir, stateLabel)
+func (g *Grub) Install(rootDir, bootDir, deviceLabel string) (err error) {
+	err = g.InstallEFI(rootDir, bootDir, constants.EfiDir, deviceLabel)
```
davidcassany, line 397: We can finally get rid of this `constants.EfiDir` parameter and use `bootDir` instead. Keeping them as separate parameters at `InstallEFI` method probably makes sense, even if with the current setup they are the same value.

Probably there are more  `constants.EfiDir` references that now can be parametrized as the bootDIr.

pkg/bootloader/grub.go
```diff
@@ -425,21 +416,25 @@ func (g *Grub) Install(rootDir, bootDir, stateLabel string) (err error) {
 // InstallConfig installs grub configuraton files to the expected location.  rootDir is the root
 // of the OS image, bootDir is the folder grub read the configuration from, usually state partition mountpoint
```
davidcassany, line 417: nit: The description ending is wrong

> usually state partition mountpoint

should be changed to 

> usually efi partition mountpoint

davidcassany (approved):
LGTM, just a couple of small remarks

pkg/bootloader/grub.go
frelon: Yep, I left it in not to make any breaking changes in the interface. We could consider this in a follow up PR!

PR #1905, opened 2024-01-17
Mount readonly EFI partition to /run/elemental/efi

frelon:
Mount EFI partition in running system to /run/elemental/efi.

This enable us to move the boot-assessment grub configuration files into the ESP.


davidcassany (approved):
LGTM :+1:

PR #1906, opened 2024-01-17
Refactor to switch to snapshotter interface

### LINKED ISSUE #1874: Refactor to snapshotter interface (including loopdevices implementation)

### LINKED ISSUE #1921: Rethink some of the options for build-disk (probably we should assume all expandable disks are unpriviliged and all full disks are always privileged builds and only support for squashfs recovery for expandable images)

### LINKED ISSUE #1922: Rethink upgrade command, probably the --recovery flag should imply an active AND recovery image upgrade (also include bootloader upgrade?).

davidcassany:
This commit adopts snapshotter interface in install, reset and upgrade commands. The change implies changes to the respective specs, grub configuration and dracut modules.

This commit also changes the behavior of recovery system upgrades. Now recovery upgrades are an optional step of a system upgrade. Recovery image can't be upgraded without upgrading the active system.

Finally build-disk command is also changed to be better aligned with upgrade and install procedures. Expandable disks are an unprivileged build and non expandable ones require privileges as they relay on snapshotter.

Fixes #1874
Fixes #1921 
Fixes #1922

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -97,14 +97,12 @@ const (
 	OEMDir             = "/run/elemental/oem"
 	PersistentDir      = "/run/elemental/persistent"
 	PersistentStateDir = "/run/elemental/persistent/.state"
-	ActiveDir          = "/run/elemental/active"
 	TransitionDir      = "/run/elemental/transition"
 	EfiDir             = "/run/elemental/efi"
 	ImgSrcDir          = "/run/elemental/imgsrc"
 	WorkingImgDir      = "/run/elemental/workingtree"
 	OverlayDir         = "/run/elemental/overlay"
 	RunningStateDir    = "/run/initramfs/elemental-state" // TODO: converge this constant with StateDir/RecoveryDir when moving to elemental-rootfs as default rootfs feature.
-	LegacyStateDir     = "/run/initramfs/cos-state"
 
 	// Running mode sentinel files
 	ActiveMode   = "/run/elemental/active_mode"
@@ -183,6 +181,7 @@ const (
 	LegacyImagesPath  = "cOS"
 	LegacyPassivePath = LegacyImagesPath + "/passive.img"
 	LegacyActivePath  = LegacyImagesPath + "/active.img"
+	LegacyStateDir    = "/run/initramfs/cos-state"
 )
 
 // GetDefaultSystemEcludes returns a list of transient paths
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -57,15 +57,14 @@ const (
 	HTTPTimeout        = 60
 	GPT                = "gpt"
 	BuildImgName       = "elemental"
-	//UsrLocalPath       = "/usr/local"
-	OEMPath        = "/oem"
-	PersistentPath = PersistentDir
-	ConfigDir      = "/etc/elemental"
-	OverlayMode    = "overlay"
-	BindMode       = "bind"
-	Tmpfs          = "tmpfs"
-	Autofs         = "auto"
-	Block          = "block"
+	OEMPath            = "/oem"
+	PersistentPath     = PersistentDir
+	ConfigDir          = "/etc/elemental"
+	OverlayMode        = "overlay"
+	BindMode           = "bind"
+	Tmpfs              = "tmpfs"
+	Autofs             = "auto"
+	Block              = "block"
 
 	// Maxium number of nested symlinks to resolve
 	MaxLinkDepth = 4
@@ -175,7 +174,7 @@ const (
 	Rsync = "rsync"
 
 	// Snapshotters
-	MaxSnaps                  = 4
+	MaxSnaps                  = 2
 	LoopDeviceSnapshotterType = "loopdevice"
 	ActiveSnapshot            = "active"
 	PassiveSnapshot           = "passive_%d"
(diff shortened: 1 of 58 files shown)


kkaempf:
We should do a code walk-through to get this merged 🤔

pkg/constants/constants.go
```diff
@@ -57,14 +57,15 @@ const (
...
 	BuildImgName       = "elemental"
-	UsrLocalPath       = "/usr/local"
-	OEMPath            = "/oem"
-	ConfigDir          = "/etc/elemental"
-	OverlayMode        = "overlay"
-	BindMode           = "bind"
-	Tmpfs              = "tmpfs"
-	Autofs             = "auto"
-	Block              = "block"
+	//UsrLocalPath       = "/usr/local"
```
frelon, line 60: Good find! Can even consider removing this.

pkg/elemental/elemental.go
```diff
@@ -441,28 +436,31 @@ func CopyFileImg(c v1.Config, img *v1.Image) error {
 
 // DeployImage will deploy the given image into the target. This method
 // creates the filesystem image file and fills it with the correspondant data
-func DeployImage(c v1.Config, img *v1.Image) (interface{}, error) {
+func DeployImage(c v1.Config, img *v1.Image) error {
```
frelon, line 439: Nice not having to return the `interface{}` !

pkg/features/embedded/cloud-config-essentials/system/oem/08_boot_assessment.yaml
```diff
@@ -84,41 +76,36 @@ stages:
...
     # Similarly, we could trigger boot assessment in other cases
     after-upgrade:
-    - <<: *efimount
-    - name: "Set upgrade sentinel"
+    - name: "Set upgrade sentinel on active"
+      if: '[ ! -f "/run/elemental/recovery_mode" ]' 
       commands:
-      - |
-          grub2-editenv /tmp/mnt/EFI/boot_assessment set enable_boot_assessment=yes
+      - grub2-editenv /run/elemental/efi/boot_assessment set enable_boot_assessment=yes
```
frelon, line 94: Isn't `/run/elemental/efi` going to be mounted read-only by default?

frelon (approved):
Big PR, so it's hard to get to everything.. I left some comments but generally looks very good!

We could also reconsider setting `constants.MaxSnaps = 2` as discussed yesterday.

Good work!

fgiudici (approved):
Skimmed through the changes after yesterday code walk-through, looks good... well, actually great job!
Hope to see this merged soon, so we can start testing/using it 🥳 
Nice to see the snapshotter interface getting alive, code looks more readable too.

pkg/features/embedded/cloud-config-essentials/system/oem/08_boot_assessment.yaml
davidcassany: No, during install we mount everything rw, in ugrades we remount it with RW (state, recovery and EFI) and in reset we do a very similar approach to install (we umount everything but recovery and then mount again with RW later on)

davidcassany: To be honest I am not very happy with the boot assessment logic shipped as a custom grub setup, probably we could consider simplifying it by not making it optional and simply including it as part of the default grub configuration.

As soon as we start implementing other snapshotter we will have to revisit grub setup again. Good thing is that nowadays it is way simpler than what it used to be.

davidcassany:
> We could also reconsider setting `constants.MaxSnaps = 2` as discussed yesterday.

Yes I have to apply these little changes yet. As soon as I manage to manually upgrade an elemental cluster from current staging to a Dev including this PR I'll mark this PR ready to be merged and arrange this little details we discussed yesterday.

davidcassany:
Finally got green run on the CI [here](https://github.com/rancher/elemental/actions/runs/7716302915) and managed to run a regular UI based upgrade from staging to this current head.

tests/common/common.go
```diff
@@ -18,7 +18,7 @@ package common
 
 import "flag"
 
-const upImg = "ghcr.io/rancher/elemental-toolkit/elemental-green:v1.3.0"
+const upImg = "ghcr.io/rancher/elemental-toolkit/elemental-green:v0.10.7-g3e4a3c56"
```
frelon, line 21: Why go back to such an old release?

davidcassany: To keep grub compatibility, this is compatible with v1.1.0 but not really with v1.2 or v1.3. In fact I think we should stick with v1.1.0 here, but not sure we ever published it to the registry. This was one of the things I wanted to review after merging this.

The issue is with the `${mode}` and `${img}` variables in grub, in v1.2 and v1.3 we assume the image is referred by the ${mode} and the dracut module computes the path. This is hard make it compatible with v1.1.0, v1.2.0 and the upcoming v2.0.x as this implies having several paths adjustments.

If we assume the ${img} and is the full path and ${mode} is just to discriminate recovery vs active args this turns to be easier in terms of backwards compatibility. We have to keep in mind in the grub domain further changes are around the corner with the btrfs snapshotter and that our backward compatibility constraint should stick to v1.1.0 version, as being the latest released elemental-toolkit version. Because of that I took a very conservative approach, because we are likely to need revising it in any case.

PR #1908, opened 2024-01-18
Include SUT utilities within the repository

davidcassany:
With this commit the ele-testhelpers repository is no longer a dependency. The only logic we required from that repository was the SUT package and this commit brings in this specific logic into the current repository.

This allows having all the logic of the integration tests self contained in the current repositry.

SUT is added as a testing utility.


frelon (approved):
This is great! :+1:

PR #1909, opened 2024-01-19
Bump twpayne/vfs module to latest version

davidcassany:
Specifically bumping from v1.7.2 to v4.3.0. We were pretty outdated.

This commit could not get rid of the old version dependency completely because yip requires a vfs v1 within the plugin API and we implement a yip plugin for partitioning in elemental-toolkit. Because of that both versions are coexisting.


pkg/cloudinit/cloudinit.go
```diff
@@ -33,12 +35,37 @@ import (
 type YipCloudInitRunner struct {
 	exec    executor.Executor
 	fs      vfs.FS
+	v1fs    v1vfs.FS
 	console plugins.Console
 }
 
 // NewYipCloudInitRunner returns a default yip cloud init executor with the Elemental plugin set.
 // It accepts a logger which is used inside the runner.
 func NewYipCloudInitRunner(l v1.Logger, r v1.Runner, fs vfs.FS) *YipCloudInitRunner {
```
davidcassany, line 44: Here is the ugly part of having to deal with different versions of the go-vfs library. Yip uses version v1, while upstream is already in v4.

kkaempf: We should create a respective issue in upstream yip

pkg/types/v1/fs.go
```diff
@@ -22,19 +22,22 @@ import (
...
-	Open(name string) (*os.File, error)
-	Chmod(name string, mode os.FileMode) error
+	Chmod(name string, mode fs.FileMode) error
 	Create(name string) (*os.File, error)
-	Mkdir(name string, perm os.FileMode) error
-	Stat(name string) (os.FileInfo, error)
-	Lstat(name string) (os.FileInfo, error)
-	RemoveAll(path string) error
+	Glob(pattern string) ([]string, error)
+	Link(oldname, newname string) error
```
davidcassany, line 28: This `Link` support for hardlinks the main reason for this change. V1 did not include support for hardlinks.

fgiudici (approved):

pkg/cloudinit/cloudinit.go
davidcassany: done https://github.com/mudler/yip/issues/123

PR #1912, opened 2024-01-24
Bind mount /oem in elemental-setup-initramfs

frelon:
Bind /oem into /sysroot/oem when running elemental-setup-initramfs to enable running initramfs stages from config files in oem partition.


davidcassany (approved):
LGTM

PR #1918, opened 2024-01-25
Include elemental partitions in mount command

davidcassany:
This PR essentially adds an additional step in mount command to mount elemental partitions before hand (OEM, Persistent, EFI, etc.). Elemental partitions should already mounted and prepare at initramfs stage and even before the persistent paths are set (otherwise instead of being defined in a persistent area they are on `/run`'s tmpfs).


frelon (approved):

davidcassany:
Confirmation run done at https://github.com/rancher/elemental/actions/runs/7691982270/job/20958173282

PR #1919, opened 2024-01-29
Test that persistence is working

frelon:
In the smoke-test we add a randomly generated file to /var/log and after reboots we check that the file is still there and contains the same data.


davidcassany (approved):
LGTM

PR #1920, opened 2024-01-29
This commit does removes the assumption of a boot folder (/boot) just for grub setup

davidcassany:
With this commit all grub binaries (EFI images and modules) and configurations are installed always in defined prefixes.

Defined prefixes are always appended to the EFI path, which is usually the EFI partition mountpoint at run time.

This causes a duplication of data if multiple prefixes are defined.

In addition this commit also removes unused variables.

### DIFF
--- a/pkg/action/build-iso.go
+++ b/pkg/action/build-iso.go
@@ -30,7 +30,6 @@ import (
 )
 
 const (
-	grubPrefixDir  = "/EFI/BOOT"
 	isoBootCatalog = "/boot/boot.catalog"
 )
 
@@ -73,7 +72,7 @@ func NewBuildISOAction(cfg *v1.BuildConfig, spec *v1.LiveISO, opts ...BuildISOAc
 	}
 
 	if b.bootloader == nil {
-		b.bootloader = bootloader.NewGrub(&cfg.Config)
+		b.bootloader = bootloader.NewGrub(&cfg.Config, bootloader.WithGrubPrefixes(constants.FallbackEFIPath))
 	}
 
 	return b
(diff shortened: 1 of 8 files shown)


pkg/action/build-iso.go
```diff
@@ -187,27 +186,30 @@ func (b *BuildISOAction) ISORun() error {
 }
 
 func (b *BuildISOAction) PrepareEFI(rootDir, uefiDir string) error {
-	return b.bootloader.InstallEFIFallbackBinaries(rootDir, uefiDir, b.spec.Label)
+	err := b.renderGrubTemplate(uefiDir)
```
davidcassany, line 189: EFI partition was missing the grub.cfg, included in the ISO filesystem but not in the EFI (vfat) filesystem.

pkg/action/build-iso.go
```diff
@@ -187,27 +186,30 @@ func (b *BuildISOAction) ISORun() error {
...
+	if err != nil {
+		return err
+	}
+	return b.bootloader.InstallEFI(rootDir, uefiDir)
 }
 
 func (b *BuildISOAction) PrepareISO(rootDir, imageDir string) error {
-	err := utils.MkdirAll(b.cfg.Fs, filepath.Join(imageDir, grubPrefixDir), constants.DirPerm)
+	// Include EFI contents in iso root too
+	return b.PrepareEFI(rootDir, imageDir)
```
davidcassany, line 198: With this commit PrepareEFI and PrepareISO are just identical, both simply prepare EFI bootloader. One is used while booting as an ISO and the other while booting the ISO as a disk (dd ISO to a raw image and boot it as a disk)

davidcassany (commented):
I think the main issue was a missing grub.cfg in EFI partition while creating the ISO, however I did some more changes in order to simplify the setup as was getting a bit confused by unused variables and duplicated variables (bootDir and efiDir were always the same and not really properly distinguished).

In the future we can probably argue if it is needed a more standard setup in EFI that acually looks more like traditional installation:

```
/EFI/BOOT/<efi binaries and small static grub.cfg>
/EFI/ELEMENTAL/<efi binaries and small static grub.cfg>
/boot/grub2/<the actual grub.cfg and grub modules>
/boot/<kernel and initrd for systemdboot or ISOs>
```

pkg/types/v1/bootloader.go
```diff
@@ -17,12 +17,11 @@ limitations under the License.
 package v1
 
 type Bootloader interface {
-	Install(rootDir, bootDir, deviceLabel string) (err error)
```
davidcassany, line 20: Device label was completely unused

pkg/types/v1/bootloader.go
```diff
@@ -17,12 +17,11 @@ limitations under the License.
...
+	Install(rootDir, bootDir string) (err error)
 	InstallConfig(rootDir, bootDir string) error
 	DoEFIEntries(shimName, efiDir string) error
-	InstallEFI(rootDir, bootDir, efiDir, deviceLabel string) error
```
davidcassany, line 23: `bootDir` and `efiDir` were set always to the same value and the code is not really prepared to properly use them as different concepts.

frelon (approved):

PR #1934, opened 2024-02-06
Remove BindMount for /oem in initramfs stage

frelon:
Since we now mount /oem in the mount command this is superfluous and might cause other oddities.


anmazzotti (approved):
Looks good! :crossed_fingers:

PR #1935, opened 2024-02-06
Add back the BindPath for /oem

frelon:
Without this, the registration breaks (initramfs stages put in /oem are not read and executed).

Edit: also, /oem should be unmounted because of the Conflicts declaration in the oem-generator...


davidcassany (approved):
LGTM

PR #1937, opened 2024-02-06
Add Volumes in Mount spec

davidcassany:
This commit allows to set specific additional mount points for block devices in mount command. They can be set by label, partlabel, uuid and device path.

In addition this commit also introduces some logic to precompute initial fstab lines for sysroot and
other mounts done in previous stages.

### DIFF
--- a/pkg/action/mount.go
+++ b/pkg/action/mount.go
@@ -17,51 +17,111 @@ limitations under the License.
 package action
 
 import (
+	"bufio"
 	"fmt"
 	"path/filepath"
+	"regexp"
 	"strings"
 
+	"github.com/hashicorp/go-multierror"
 	"github.com/rancher/elemental-toolkit/pkg/constants"
-	"github.com/rancher/elemental-toolkit/pkg/elemental"
 	v1 "github.com/rancher/elemental-toolkit/pkg/types/v1"
 	"github.com/rancher/elemental-toolkit/pkg/utils"
 )
 
 const overlaySuffix = ".overlay"
+const labelPref = "LABEL="
+const partLabelPref = "PARTLABEL="
+const uuidPref = "UUID="
+const devPref = "/dev/"
+const diskBy = "/dev/disk/by-"
+const diskByLabel = diskBy + "label"
+const diskByPartLabel = diskBy + "partlabel"
+const diskByUUID = diskBy + "uuid"
+const runPath = "/run"
 
 func RunMount(cfg *v1.RunConfig, spec *v1.MountSpec) error {
+	var fstabData string
+	var err error
+
 	cfg.Logger.Info("Running mount command")
 
-	cfg.Logger.Debug("Mounting elemental partitions")
-	if err := elemental.MountPartitions(cfg.Config, spec.Partitions.PartitionsByMountPoint(false)); err != nil {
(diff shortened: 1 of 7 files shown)


pkg/action/mount.go
```diff
@@ -192,68 +266,109 @@ func MountOverlayPath(cfg *v1.RunConfig, sysroot, overlayDir, path string) error
...
+	}
+
+	scanner := bufio.NewScanner(strings.NewReader(strings.TrimSpace(string(output))))
+	for scanner.Scan() {
+		lineFields := strings.Fields(scanner.Text())
+		if len(lineFields) != 4 {
+			continue
+		}
+		if lineFields[2] == "btrfs" {
+			r := regexp.MustCompile(`(/.+)\[.*\]`)
```
davidcassany, line 357: On btrfs mounted subvolume added in brackets something like `/dev/sda1[/@/.snapshots/1/snapshot]` hence we need to remove it  from the SOURCE field, that's why there this extra parsing logic here.

pkg/config/config.go
```diff
@@ -220,46 +220,26 @@ func NewInitSpec() *v1.InitSpec {
...
-
+func NewMountSpec() *v1.MountSpec {
 	return &v1.MountSpec{
 		Sysroot:    "/sysroot",
 		WriteFstab: true,
-		Partitions: ep,
+		Volumes: []*v1.VolumeMount{
+			{
+				Mountpoint: constants.PersistentDir,
+				Device:     fmt.Sprintf("PARTLABEL=%s", constants.PersistentPartName),
```
davidcassany, line 230: GPT partition labels are not configurable, only filesystem labels, so adding these here is fine, this is not does collision with eventual custom user labels. 

In addition the user has the capability to completely rewrite this list in config.yaml if needed.

pkg/types/v1/config.go
```diff
@@ -251,15 +251,22 @@ type InitSpec struct {
...
+	Volumes    []*VolumeMount   `yaml:"volumes,omitempty" mapstructure:"volumes"`
 	Ephemeral  EphemeralMounts  `yaml:"ephemeral,omitempty" mapstructure:"ephemeral"`
 	Persistent PersistentMounts `yaml:"persistent,omitempty" mapstructure:"persistent"`
 }
 
+type VolumeMount struct {
+	Mountpoint string   `yaml:"mountpoint,omitempty" mapstructure:"mountpoint"`
+	Device     string   `yaml:"device,omitempty" mapstructure:"device"`
+	Options    []string `yaml:"options,omitempty" mapstructure:"options"`
+	Persistent bool     `yaml:"persistent,omitempty" mapstructure:"persistent"`
```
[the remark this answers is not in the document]
frelon: Yep, that looks really good! I'm also missing a type-field if you want to specify the fs type (which could default to `auto`).

pkg/action/mount.go
```diff
@@ -192,68 +279,110 @@ func MountOverlayPath(cfg *v1.RunConfig, sysroot, overlayDir, path string) error
 	return nil
 }
 
-func findmnt(runner v1.Runner, mountpoint string) (string, error) {
```
frelon, line 195: Nice!

frelon (approved):
Looks great!

PR #1938, opened 2024-02-07
Prevent unnecessary copy of image tree to tmp folder

davidcassany:
This commit prevents copying already accessible image trees to a temporary foldery to later on sync it again to the final filesystem image.

Prevents rsyncing twice the root tree for directory and file image sources. This tends to be the case for recovery image deployments, which usually are just a copy of the local active system.


frelon (approved):
Very nice! :+1:

PR #1942, opened 2024-02-07
Use go 1.22

frelon:
Also includes an update to the README for the correct copyright year.

### DIFF
--- a/.obs/specfile/elemental-toolkit.spec
+++ b/.obs/specfile/elemental-toolkit.spec
@@ -40,7 +40,7 @@ Requires:       util-linux
 Requires:       gptfdisk
 
 %if 0%{?suse_version}
-BuildRequires:  golang(API) >= 1.18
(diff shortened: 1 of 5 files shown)


.obs/specfile/elemental-toolkit.spec
```diff
@@ -49,7 +49,7 @@ BuildRequires:  golang-packaging
 %global commit     d1ae3f9a425de2618f9058f3b37583ef3ce52c7d
 %gometa
 %if (0%{?centos_version} == 800) || (0%{?rhel_version} == 800)
-BuildRequires:  go1.20
+BuildRequires:  go1.22
```
kkaempf, line 52: <s>This might not be available in SLE Micro 6 🤔 </s>

Should become available 🤞🏻

kkaempf (approved):
🤞🏻

PR #1949, opened 2024-02-09
Add build-docker-cli for local development

anmazzotti:
Part of rancher/elemental#1202


frelon:
Is a new target really necessary? It's already possible to do `make VERSION=dev build` to get the same result.

anmazzotti:
> Is a new target really necessary? It's already possible to do `make VERSION=dev build` to get the same result.

Good one. I'll close this PR then and update the doc.

PR #1950, opened 2024-02-12
Remove oem partition from grub kernel args

anmazzotti:
Part of rancher/elemental#1210


frelon:
Hmm, not including the oemlabel means /oem is not mounted during boot, which in turn means the image is not expanded (the yip config for expanding the disk is placed in the /oem partition by `build-disk` command), which is the reason for the test system timing out waiting to be expanded.

anmazzotti:
> Hmm, not including the oemlabel means /oem is not mounted during boot, which in turn means the image is not expanded (the yip config for expanding the disk is placed in the /oem partition by `build-disk` command), which is the reason for the test system timing out waiting to be expanded.

Yes, I just noticed all the tests failing. 
But at least we found the root cause, I think this was introduced with the disk images.

PR #1959, opened 2024-02-14
Install podman in example Dockerfiles

frelon:


kkaempf (approved):
what could possibly go wrong ? 😆

PR #1960, opened 2024-02-15
Improved upgrade test

### LINKED ISSUE #1928: Create an upgrade test from an old installation to current PR
This issue is devoted to explore and test different approaches regarding backward compatibility.

We currently test only downgrades (from current PR to and older system). We should also test the other direction. For that I'd suggest the following.

From an older system upgrade/download latest elemental client (the PR build) and use it to upgrade the old system to the latest and greatest OS (the PR build).

In addition we should also make sure that the old system is based on elemental-toolkit v1.1.1, hence probably we should also make a new v1.1.1 release.

frelon:
The upgrade test will now prepare a VM with an older (v1.1.4) version and then run the upgrade in podman to upgrade to the current PR container. 

This will make sure we can upgrade from v1.1.4 to current and also that the correct version of the upgrade command is tested (the new one, not the old installed one).

Fixes #1928

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -3,7 +3,7 @@ on:
   push:
     tags:
       - v**
-  pull_request_target:
+  pull_request:
     types:
       - opened
       - synchronize
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -3,7 +3,7 @@ on:
   push:
     tags:
       - v**
-  pull_request:
+  pull_request_target:
     types:
       - opened
       - synchronize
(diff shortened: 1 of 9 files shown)


frelon:
Got a green upgrade run at https://github.com/rancher/elemental-toolkit/actions/runs/7917180349/job/21613080610?pr=1960 going back to `pull_request_target` event.

frelon:
> I am a bit lost with these changes... Shouldn't it be a new test, so we test both directions:
> 
> 1. from current head downgrade to a names tag/version
> 
> 2. from a named/tag version upgrade to current head
> 
> 
> 
> 1 is what we already had, we are missing 2.
> 
> 
> 
> I guess this also apply to the recovery tests which in fact is just an upgrade test executed from recovery system.

I can add back the old upgrade test and call it downgrade instead, then we get both tests named correctly 👍

davidcassany (approved):
Nice! LGTM :+1:

PR #1963, opened 2024-02-19
Fix docs publishing error

frelon:
Fix error found here: https://github.com/rancher/elemental-toolkit/actions/runs/7959595286/job/21726878077


davidcassany (approved):
LGTM

PR #1965, opened 2024-02-20
Build toolkit on nightly

frelon:
We could also consider building the toolkit when pushing to main, but this will fix the error for now.


fgiudici (approved):
LGTM

PR #1966, opened 2024-02-21
Add upgrade-recovery command to upgrade recovery partition only

anmazzotti:


cmd/config/config.go
```diff
@@ -452,6 +452,29 @@ func ReadUpgradeSpec(r *v1.RunConfig, flags *pflag.FlagSet) (*v1.UpgradeSpec, er
 	return upgrade, err
 }
 
+func ReadUpgradeRecoverySpec(r *v1.RunConfig, flags *pflag.FlagSet) (*v1.UpgradeRecoverySpec, error) {
```
davidcassany, line 455: I am not entirely sure this is actually needed, probably we can still use the  ` ReadUpgradeSpec`. This mostly a matter on how we want to present the configuration yaml. Currently we have:

```yaml
install:
  # installation details
  ...
upgrade:
  # upgrade details, also including recovery setup nowadays
  ...
reset:
  # reset details
  ...
```
Having this method here assumes there is also another `upgrade-recovery` stanza. Probably we can just reuse the  `upgrade` one as it already includes the recovery information, in that case the configuration load would be the same for both cases.

An alternate way of coding it would be keeping the current `upgrade` but distinguishing between three different paths: system-only, system-and-recovery and recovery-only. Currently the only path that is not implemented is the recovery-only. So, probably, in terms coding is easier to start with a single command as it is now by adding in spec struct an additional flag or enum like variable to distinguish the three paths and adapt the current already existing action. Then from that I'd try to strip out in separate methods system and recovery upgrades.

PR #1967, opened 2024-02-21
Use backup legacy grub.cfg location

frelon:
During build-disk, if the new /etc/elemental/grub.cfg config is not found, revert to using /etc/cos/grub.cfg for backwards compatibility.


davidcassany (approved):
LGTM :+1:

PR #1969, opened 2024-02-22
bump elemental-toolkit to v2

### LINKED ISSUE #1927: Bump elemental-tookit to v2
In #1906 we introduced breaking changes so we should mark it as a major release. We should arrange whatever is needed to mark current head as a v2.

davidcassany:
Fixes #1927


cmd/build-disk.go
```diff
@@ -19,15 +19,15 @@ package cmd
 import (
 	"os/exec"
 
-	"github.com/rancher/elemental-toolkit/pkg/constants"
-	eleError "github.com/rancher/elemental-toolkit/pkg/error"
-	v1 "github.com/rancher/elemental-toolkit/pkg/types/v1"
+	"github.com/rancher/elemental-toolkit/v2/pkg/constants"
+	eleError "github.com/rancher/elemental-toolkit/v2/pkg/error"
+	v2 "github.com/rancher/elemental-toolkit/v2/pkg/types/v2"
```
frelon, line 24: Should we rename `pkg/types/v2` to just `pkg/types`? I don't think it makes sense to version those types more than is done for the entire module.

davidcassany: This a good point, yes we could consider having a single package if we do not foresee supporting both at the same time with the same code.

cmd/build-disk.go
```diff
@@ -19,15 +19,15 @@ package cmd
 import (
 	"os/exec"
 
-	"github.com/rancher/elemental-toolkit/pkg/constants"
-	eleError "github.com/rancher/elemental-toolkit/pkg/error"
-	v1 "github.com/rancher/elemental-toolkit/pkg/types/v1"
+	"github.com/rancher/elemental-toolkit/v2/pkg/constants"
+	eleError "github.com/rancher/elemental-toolkit/v2/pkg/error"
+	"github.com/rancher/elemental-toolkit/v2/pkg/types"
```
davidcassany, line 24: @frelon if I understood what you meant you were imagining something like that right? So no v1 or v2, only types which somehow also reflects we are not supporting both on the same code base.

frelon: Yep, otherwise we could consider keeping the `pkg/types/v1` directory and only making changes in `v2`, but it feels like overkill to me :+1:

frelon (approved):

PR #1974, opened 2024-02-23
Add upgrade-recovery subcommand

anmazzotti:
> A different take of https://github.com/rancher/elemental-toolkit/pull/1966
> 
> This PR does not actually introduce a new `upgrade-recovery` subcommand, but instead just adds a `--recovery-only` `upgrade` subcommand argument.

Scratch the above.
This is now something in the middle.

`elemental upgrade-recovery --recovery-system.uri <imgRef>` is the final choice.  

`elemental upgrade --recovery` can still be used as usual to upgrade both recovery and system at the same time.

The `upgrade-recovery` subcommand shares the `UpgradeSpec` with `upgrade` to reuse the same configuration/initialization logic, but filters it at a command level.
Only `--recovery-system` probably makes sense here, most of other flags should be disabled.

### DIFF
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -452,6 +452,29 @@ func ReadUpgradeSpec(r *v1.RunConfig, flags *pflag.FlagSet) (*v1.UpgradeSpec, er
 	return upgrade, err
 }
 
+func ReadUpgradeRecoverySpec(r *v1.RunConfig, flags *pflag.FlagSet) (*v1.UpgradeSpec, error) {
+	upgrade, err := config.NewUpgradeSpec(r.Config)
+	if err != nil {
+		return nil, fmt.Errorf("failed initializing upgrade recovery spec: %v", err)
+	}
+	vp := viper.Sub("upgrade-recovery")
+	if vp == nil {
+		vp = viper.New()
+	}
+	// Bind upgrade-recovery cmd flags
+	bindGivenFlags(vp, flags)
+	// Bind upgrade-recovery env vars
+	viperReadEnv(vp, "UPGRADE_RECOVERY", constants.GetUpgradeRecoveryKeyEnvMap())
+
+	err = vp.Unmarshal(upgrade, setDecoder, decodeHook)
+	if err != nil {
+		r.Logger.Warnf("error unmarshalling UpgradeSpec: %s", err)
+	}
+	err = upgrade.SanitizeForRecoveryOnly()
+	r.Logger.Debugf("Loaded upgrade UpgradeSpec: %s", litter.Sdump(upgrade))
+	return upgrade, err
+}
+
 func ReadBuildISO(b *v1.BuildConfig, flags *pflag.FlagSet) (*v1.LiveISO, error) {
 	iso := config.NewISO()
 	vp := viper.Sub("iso")
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -429,7 +429,7 @@ func ReadResetSpec(r *v1.RunConfig, flags *pflag.FlagSet) (*v1.ResetSpec, error)
 	return reset, err
 }
 
-func ReadUpgradeSpec(r *v1.RunConfig, flags *pflag.FlagSet) (*v1.UpgradeSpec, error) {
+func ReadUpgradeSpec(r *v1.RunConfig, flags *pflag.FlagSet, recoveryOnly bool) (*v1.UpgradeSpec, error) {
 	upgrade, err := config.NewUpgradeSpec(r.Config)
 	if err != nil {
 		return nil, fmt.Errorf("failed initializing upgrade spec: %v", err)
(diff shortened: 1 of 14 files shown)


frelon (commented):
I prefer adding the `upgrade-recovery` command. 

Adding flags that completely changes the implementation of the `upgrade` command feels like a recipe for unintended consequences.

anmazzotti:
Part of rancher/elemental#1218

pkg/action/upgrade.go
```diff
@@ -211,19 +211,11 @@ func (u *UpgradeAction) mountRWPartitions(cleanup *utils.CleanStack) error {
...
-			return elementalError.NewFromError(err, elementalError.MountRecoveryPartition)
-		}
-		cleanup.Push(umount)
-	} else {
-		umount, err = elemental.MountRWPartition(u.cfg.Config, u.spec.Partitions.State)
-		if err != nil {
-			return elementalError.NewFromError(err, elementalError.MountStatePartition)
-		}
-		cleanup.Push(umount)
+	umount, err = elemental.MountRWPartition(u.cfg.Config, u.spec.Partitions.State)
```
davidcassany, line 214: This is likely to produce an undesired effect or even fail, it is likely to attempt mounting the current root as RW. By default `/run/initramfs/elemental-state` is already mounted as RW on a running system with the current setup. So it should not be needed to actually mount it as RW, it already is as RW.

Also, on a running system on btrfs, the active subvolume is mounted by default on this partition and it could also be the reported mount point from ghw library. This is the actual reason this was only called in recovery mode in the former code. Probably we need to move or add the logic at https://github.com/rancher/elemental-toolkit/blob/95389608574d77b25d583c1acbcac714ab5ee8d7/pkg/snapshotter/btrfs.go#L736-L761 into `utils.GetAllPartitions` method so we can have a known and consistent mountpoint defined for state partition.
The problem is that current Partitions struct and the underlaying ghw library assumes a single mountpoint per partition, while there could be more. In fact in the upgrade pod mountpoints under `/host/run` and `/run` are duplicated, we have not criteria from elemental-toolkit to distinguish them or even reporting them.
At mid term I'd probably drop ghw library use in favor of an `lsblk` wrapper (we used to have that at a time) and report at list of mountpoints per partition. But this is relatively ambitious refactor... specially in terms of testing.

cmd/config/config.go
```diff
@@ -452,6 +452,29 @@ func ReadUpgradeSpec(r *v1.RunConfig, flags *pflag.FlagSet) (*v1.UpgradeSpec, er
 	return upgrade, err
 }
 
+func ReadUpgradeRecoverySpec(r *v1.RunConfig, flags *pflag.FlagSet) (*v1.UpgradeSpec, error) {
+	upgrade, err := config.NewUpgradeSpec(r.Config)
+	if err != nil {
+		return nil, fmt.Errorf("failed initializing upgrade recovery spec: %v", err)
+	}
+	vp := viper.Sub("upgrade-recovery")
```
davidcassany, line 460: Are we sure this is correct? shouldn't it be `vp := viper.Sub("upgrade")`? IIRC this states the sub schema item, hence it maps whatever is loaded from the yaml under `upgrade-recovery` stanza. I guess this could be tested and validated by checking this method properly reads `upgrade` section of a yaml. I assume this falls into the scope of `cmd/config/config_test.go`.

I think this should be fixed or at least validated with a unit test in this PR.

davidcassany (changes requested):
Nice work :+1:

I think the only actual request change I have to actually include the yaml loading unit test. I think the current sub viper setup is wrong (speaking from memory I might miss some detail), but in any case a simple unit test in that front will tell us and actually verify it works when fixed, in case it needs to be fixed.

The other issues, probably it is worth addressing them in a separate PR, not so sure. If we don't face obvious breaks and regressions with this I'd vote to merge it and iterate from that point. I think it might be easier this way.

pkg/action/upgrade-recovery.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+		u.spec.State.Partitions[constants.RecoveryPartName] = recoveryPart
+	}
+
+	// Hack to ensure we are not using / or /.snapshots mountpoints. Btrfs based deployments
+	// mount state partition into multiple locations
+	statePath := filepath.Join(u.spec.Partitions.State.MountPoint, constants.InstallStateFile)
+	if u.spec.Partitions.State.MountPoint == "/" || u.spec.Partitions.State.MountPoint == "/.snapshots" {
+		statePath = filepath.Join(constants.RunningStateDir, constants.InstallStateFile)
+	}
```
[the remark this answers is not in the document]
anmazzotti: Good catch. I did inherit the hack and I didn't notice it was removed. 
It's gone now. Thank you

pkg/action/upgrade.go
anmazzotti: Interesting. I actually changed this to always remount it because the CI tests were failing because `/run/initramfs/elemental-state` could not be found. I will revert this change here, most likely this is not the solution if there was an issue with it. 
I also noticed the state partition was already RW mounted, this is why I added the debug log somewhere else to confirm this. Then the test just worked. :shrug:

cmd/config/config.go
anmazzotti: Yes, sorry I got confused (again :D ) with it.
I removed it all and just added a (bit ugly) flag to sanitize the same stanza when used for recovery only, since it's the only logical difference in there.

PR #1977, opened 2024-02-26
Fix regression introduced in 547cb932

### LINKED ISSUE #1973: Cant update bootloader environment variables on upgrade in a pod or container
**elemental-toolkit version:**

Current head e7630b0a1f497e167fa45b4ea316f81038d47e45

tested on x86_64

**Describe the bug**

The upgrade fails if elemental client is running inside a container or inside a POD.

**To Reproduce**

Run an upgrade with current HEAD SLE Micro images from the elemental-operator. In this context the upgrade will happen in a POD and executed by the `suc-upgrade` script. This is a particular environment.

**Expected behavior**

Upgrade succeeds and elemental is capable to write bootloader environment variables.

**Additional context**

The root cause of this issue is that elemental client expects certain partitions to be mounted in certain constant locations, this is the case for EFI partition. EFI partition is expected to be mounted at `/run/elemental/efi` and during upgrade it is remounted as RW based on the detected mountpoint rather than based on a constant path. Hence during upgrades using suc-upgrade script `/host/run/elemental/efi` is remounted RW and at the same time elemental client tries to write to `/run/elemental/efi`.

This is an inconsistentcy in snapshotters implementation that do not include have visibility in `ElementalPartitions` object that tracks the mount point of essential elemental partitions. IMHO we could include that object in snapshotters init method as part of the interface so snapshotter do have the relevant disk knowledge.

davidcassany:
Fixes #1973

### DIFF
--- a/pkg/features/embedded/cloud-config-essentials/system/oem/08_boot_assessment.yaml
+++ b/pkg/features/embedded/cloud-config-essentials/system/oem/08_boot_assessment.yaml
@@ -16,22 +16,14 @@ stages:
     # enable_boot_assessment to 'yes'.
     # This can be then customized easily by having a cloud-config file which always enables boot assessment on 
     # the boot stage.
-    - &remountEFIrw
-      name: "Mounting EFI as RW"
-      if: 'mountpoint -q /run/elemental/efi'
-      commands:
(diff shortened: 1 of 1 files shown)


anmazzotti (approved):
Looks good.

PR #1979, opened 2024-02-26
Relay on the snapshotter setup of the installed system for upgrades

davidcassany:
Fixes rancher/elemental#1243


anmazzotti (approved):
LGTM

PR #1984, opened 2024-02-29
Keep previous behavior for persistent binded paths

### LINKED ISSUE #1982: Root password lost after an OS upgrade
**elemental-toolkit version:** 1.1.1

**elemental-operator version:** 1.5.0 (Dev)

**CPU architecture, OS, and Version:** x86, Elemental OS v2.0.2 (based on SLE Micro 5.5)

**Describe the bug**

After an upgrade from Elemental OS v2.0.2 (Stable) to v2.1.0 (Dev) with `managedOSVersionName` method the root password previously configured during deployment is not working anymore. Erased?

**To Reproduce**

Install Rancher Manager v2.8.2, `elemental-operator` Dev (1.5.0), deploy a simple one node cluster with Elemental OS Stable (2.0.2) and update it to Dev (2.1.0) with `ManagedOSVersionName`. Wait for the node to reboot and try to connect with either ssh on directly on the console (for baremetal of VM).

`ManagedOSImage` used:
```yaml
apiVersion: elemental.cattle.io/v1beta1
kind: ManagedOSImage
metadata:
  name: with-os-version-name
  namespace: fleet-default
spec:
  clusterTargets:
    - clusterName: cluster-k3s
  managedOSVersionName: v2.1.0-unstable
```

**Expected behavior**

OS upgrade done and root connection still working.

**Logs**

[elemental-operator.log](https://github.com/rancher/elemental-toolkit/files/14434018/elemental-operator.log)

**Additional context**

N/A

davidcassany:
With this patch the current (stable release) behavior of binded persistent paths is kept.

This commit adds a new rsync wrapper method (`MirrorData`) which includes the `--delete` flag. Named mirror because it actually causes the target and source to have the same exact data causing the deletion of any file already present in _target_ that is not present in _source_. This is the behavior we want when dumping the image root-tree to a new snapshot.

Current behavior updates at early boot all files in a given path from the image to the persisted path. This includes a couple of non obvious behaviors:

* Reverts changes done in files that were originally part of the image on every boot.
* Files included in path configured as persistent that are present in image A but not present in image B and kept in the persistent storage when upgrading from A to B.

Fixes #1982

### DIFF
--- a/pkg/action/mount.go
+++ b/pkg/action/mount.go
@@ -225,17 +225,14 @@ func MountBindPath(cfg *v1.RunConfig, sysroot, overlayDir, path string) error {
 	pathName := strings.ReplaceAll(trimmed, "/", "-") + ".bind"
 	stateDir := fmt.Sprintf("%s/%s", overlayDir, pathName)
 
-	// Only sync data once, otherwise it could modify persistent data from a previous boot
-	if ok, _ := utils.Exists(cfg.Fs, stateDir); !ok {
-		if err := utils.MkdirAll(cfg.Fs, stateDir, constants.DirPerm); err != nil {
-			cfg.Logger.Errorf("Error creating upperdir %s: %s", stateDir, err.Error())
-			return err
-		}
+	if err := utils.MkdirAll(cfg.Fs, stateDir, constants.DirPerm); err != nil {
+		cfg.Logger.Errorf("Error creating upperdir %s: %s", stateDir, err.Error())
+		return err
+	}
 
-		if err := utils.SyncData(cfg.Logger, cfg.Runner, cfg.Fs, base, stateDir); err != nil {
-			cfg.Logger.Errorf("Error shuffling data: %s", err.Error())
-			return err
-		}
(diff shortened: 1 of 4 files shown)


fgiudici (approved):
Wondering if could be worth to add a unit test in utils_test.go to cover MirrorData... but maybe not after all, as is just the old SyncData (and the same as the actual SyncData but with the --delete 🤷🏼 ).
LGTM!

ldevulder (approved):
Let's go to merge and test!

davidcassany:
> Wondering if could be worth to add a unit test in utils_test.go to cover MirrorData... but maybe not after all, as is just the old SyncData (and the same as the actual SyncData but with the --delete 🤷🏼 ). LGTM!

Yes I also wondered about it, then I realized that all lines of this method are actually covered by unit tests (kudos to codecov report), the only four lines that are not covered, are the error exits for mount command and they are already not covered...

ldevulder:
> Yes I also wondered about it, then I realized that all lines of this method are actually covered by unit tests

And I also think that in case of issue the Elemental CI should trap it quickly.

PR #1988, opened 2024-03-01
Fix Recovery only upgrades

anmazzotti:
Part of rancher/elemental#1218

### DIFF
--- a/pkg/elemental/elemental.go
+++ b/pkg/elemental/elemental.go
@@ -294,7 +294,7 @@ func UnmountFileSystemImage(c v1.Config, img *v1.Image) error {
 // CreateFileSystemImage creates the image file for the given image. An root tree path
 // can be used to determine the image size and the preload flag can be used to create an image
 // including the root tree data.
-func CreateFileSystemImage(c v1.Config, img *v1.Image, rootDir string, preload bool) error {
+func CreateFileSystemImage(c v1.Config, img *v1.Image, rootDir string, preload bool, excludes ...string) error {
 	c.Logger.Infof("Creating image %s from rootDir %s", img.File, rootDir)
 	err := utils.MkdirAll(c.Fs, filepath.Dir(img.File), cnst.DirPerm)
 	if err != nil {
@@ -303,7 +303,7 @@ func CreateFileSystemImage(c v1.Config, img *v1.Image, rootDir string, preload b
 	}
 
 	if img.Size == 0 && rootDir != "" {
-		size, err := utils.DirSizeMB(c.Fs, rootDir)
+		size, err := utils.DirSizeMB(c.Fs, rootDir, excludes...)
 		if err != nil {
 			return err
 		}
@@ -340,7 +340,7 @@ func CreateFileSystemImage(c v1.Config, img *v1.Image, rootDir string, preload b
 // CreateImageFromTree creates the given image including the given root tree. If preload flag is true
 // it attempts to preload the root tree at filesystem format time. This allows creating images with the
 // given root tree without the need of mounting them.
-func CreateImageFromTree(c v1.Config, img *v1.Image, rootDir string, preload bool, cleaners ...func() error) (err error) {
+func CreateImageFromTree(c v1.Config, img *v1.Image, rootDir string, preload bool, excludes []string, cleaners ...func() error) (err error) {
 	defer func() {
 		for _, cleaner := range cleaners {
 			if cleaner == nil {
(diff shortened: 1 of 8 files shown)


pkg/elemental/elemental.go
```diff
@@ -432,6 +432,11 @@ func DeployImage(c v1.Config, img *v1.Image) error {
 	transientTree := strings.TrimSuffix(img.File, filepath.Ext(img.File)) + ".imgTree"
 	if img.Source.IsDir() {
 		transientTree = img.Source.Value()
+		err = DumpSource(c, transientTree, img.Source)
```
davidcassany, line 435: DumpSource is not needed here as the source is already a directory. `DumpSource` just copies the source contents to the given directory (so for a dir source type it only rsyncs from one path to antoher)

anmazzotti:
Back to draft since this seems to be breaking the btrfs install.

pkg/utils/fs.go
```diff
@@ -38,23 +38,30 @@ import (
 )
 
 // DirSize returns the accumulated size of all files in folder. Result in bytes
-func DirSize(fs v1.FS, path string) (int64, error) {
+func DirSize(fs v1.FS, path string, excludes ...string) (int64, error) {
```
davidcassany, line 41: We definitely need a unit test exercising this change. I can be easily coded in `pkg/utils/utils_test.go` there is already a context there for this function, so just adding one or two runs to exercise the extra `return` and eventual corner cases should be enough to be sure this is functional and raise on regressions.

pkg/elemental/elemental.go
```diff
@@ -294,16 +294,16 @@ func UnmountFileSystemImage(c v1.Config, img *v1.Image) error {
 // CreateFileSystemImage creates the image file for the given image. An root tree path
 // can be used to determine the image size and the preload flag can be used to create an image
 // including the root tree data.
-func CreateFileSystemImage(c v1.Config, img *v1.Image, rootDir string, preload bool) error {
-	c.Logger.Infof("Creating image %s", img.File)
+func CreateFileSystemImage(c v1.Config, img *v1.Image, rootDir string, preload bool, excludes ...string) error {
```
davidcassany, line 297: I am wondering if do we have a need to actually bring this excluding paths up that far. Do we have any call of `CreateImageFromTree` where we don't set it to the defaults or were this could have an undesired effect? Probably the same applies for `CreateFileSystemImage`.

I mean probably it is just enough adding the default excludes inside  `CreateFileSystemImage` and on `CreateImageFromTree` and there is no need to map this change that far up. I think having it ported all way up males sense if we make use of it at action level, otherwise probably we can keep it minimal.

pkg/utils/fs.go
davidcassany: I am also wondering if `filepath.SkipDir` is equivalent to `fs.SkipDir` from `io/fs` package. 

The function we are using https://github.com/twpayne/go-vfs/blob/master/walk.go seams to honor the `io/fs` version, however I have not had a closer look to check if there is some sort of compatibility or there is some detail I am missing.

pkg/elemental/elemental.go
anmazzotti: Good point. I didn't want to sneak it in because I was afraid it could have unexpected effects. 
This is also used by the build-disk, however I agree it *should* not have any undesired effect. 

I think we can hardcode this in CreateImageFromTree, since we need to pass the exclusion list to CreateFileSystemImage and later utils.SyncData.

anmazzotti:
Did a full re-test after latest fixes and rebasing:
1. Deploy a cluster with 2 nodes (one btrfs, one loopdevice)
2. Trigger a OS+Recovery upgrade using ManagedOSImage 
3. Trigger a RecoveryOnly upgrade with ManagedOSImage

Seems to all work good now!

pkg/utils/utils_test.go
```diff
@@ -696,7 +696,12 @@ var _ = Describe("Utils", Label("utils"), func() {
 			Expect(err).ShouldNot(HaveOccurred())
 			Expect(size).To(Equal(int64(3072)))
 		})
-		It("Returns the expected size of a test folder", func() {
+		It("Returns the size of a test folder when skipping subdirectories", func() {
```
davidcassany, line 699: Nice, loving this sort of tests. They are super useful in case someday in the future we change some apparently simple utility.

davidcassany (approved):
Nice, looks good :+1:

PR #1991, opened 2024-03-06
Adding debug messages for snapper configuration

davidcassany:


anmazzotti (approved):
Tested successfully, so looks good to me.

davidcassany:
With the current change it can't run the fallback test as the downloaded image does not include snapper. Which is good, it fails early in the process and it simply refuses to upgrade.  Merging as the issue is in test-fallback itself, it should try to upgrade to the self image.

PR #1992, opened 2024-03-06
Temporarily fix upgrade tests

davidcassany:
I think we could consider three different test scenarios for toolkit:

* upgrade from a previous version to current head code [this is the current upgrade test]
* upgrade from current head version to an older version [this would be the downgrade test]
* upgrade from current head to current head [this would what downgrade and recovery tests are doing in this current PR]

As soon as we get a v2 released in elemental-toolkit we could easily implement this three tests scenarios. In fact there is only one missing.

### DIFF
--- a/tests/recovery/recovery_test.go
+++ b/tests/recovery/recovery_test.go
@@ -79,7 +79,7 @@ var _ = Describe("Elemental Recovery upgrade tests", func() {
 		When("using specific images", func() {
 			It("upgrades to a specific image and reset back to the installed version", Label("third-test"), func() {
 				By(fmt.Sprintf("upgrading to %s", comm.UpgradeImage()))
-				cmd := s.ElementalCmd("upgrade", "--recovery", "--recovery-system.uri", comm.UpgradeImage())
+				cmd := s.ElementalCmd("upgrade-recovery", "--recovery-system.uri", comm.UpgradeImage())
 				By(fmt.Sprintf("running %s", cmd))
 				out, err := s.Command(cmd)
 				_, _ = fmt.Fprintln(GinkgoWriter, out)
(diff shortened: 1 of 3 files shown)


tests/recovery/recovery_test.go
```diff
@@ -68,8 +68,9 @@ var _ = Describe("Elemental Recovery upgrade tests", func() {
 			s.Reboot()
 			s.EventuallyBootedFrom(sut.Active)
 
-			upgradedVersion := s.GetOSRelease("TIMESTAMP")
-			Expect(upgradedVersion).ToNot(Equal(currentVersion))
+			// TODO check upgrade matches expectations
```
davidcassany, line 71: With the current change in this PR, the upgrade happens to the same exact image, hence the /etc/os-release will not change after the upgrade. In fact to validate upgrades would be better to test state.yaml file. We could consider adding some helpers to fetch this file in SUT.

anmazzotti:
What I do when manually testing is that I rebuild elemental-toolkit with an arbitrary GIT_COMMIT, then I build a second image using that elemental-toolkit version, and finally I can test upgrade was done by executing `elemental version --long` (that should include the GIT_COMMIT value).

Maybe in the tests we could follow a similar logic.

davidcassany:
> What I do when manually testing is that I rebuild elemental-toolkit with an arbitrary GIT_COMMIT, then I build a second image using that elemental-toolkit version, and finally I can test upgrade was done by executing `elemental version --long` (that should include the GIT_COMMIT value).
> 
> Maybe in the tests we could follow a similar logic.

The intention was to downgrade and upgrade to known and meaningful states for a basic compatibility check. So that upgrading from current stable (v1.1.x) and downgrading to v2 (assuming head is >= 2.0.0). Version 2 is still not tagged, but as soon as it is I'd just stick to that.

Then, what I think is also relevant is upgrading to the same head of the PR, this is likely to unveil some upgrade issues introduced by the PR that are not exercised with the upgrade and downgrade test, such as the one you saw with the snapper config. This would be the closer we can get to emulate an upgrade to an eventual higher version.

Also something we are not doing in this tests and we should is reboot to fallback. The problem is that in github performance is really poor and this will make upgrade tests quite longer.

davidcassany:
Closed in favor of #2003 and #2004

PR #1995, opened 2024-03-07
Make ext2 the default FS for recovery on disks

davidcassany:


davidcassany:
Confirmed in test logs that the disk was build with ext2

```
INFO[2024-03-07T10:40:47Z] Creating image /build/build/recovery/recovery.img from rootDir /build/build/recovery.img.root 
DEBU[2024-03-07T10:40:47Z] Image size 891M                              
DEBU[2024-03-07T10:40:47Z] Running cmd: 'mkfs.ext2 -L COS_SYSTEM -d /build/build/recovery.img.root /build/build/recovery/recovery.img' 
```

frelon (approved):
Nice! Looks like you need to update the tests as well, approving anyways :+1:

davidcassany:
> Nice! Looks like you need to update the tests as well, approving anyways 👍

Yes, gonna force merge. Then make v2 available and then update tests to actually use v1.1.4 for upgrade tests and v2.0.0 for downgrade and recovery tests. So they will be green again and testing actual tags.

frelon:
> > Nice! Looks like you need to update the tests as well, approving anyways 👍
> 
> Yes, gonna force merge. Then make v2 available and then update tests to actually use v1.1.4 for upgrade tests and v2.0.0 for downgrade and recovery tests. So they will be green again and testing actual tags.

Ah, I was thinking of the cli unit-tests which expect "mksquashfs" to be called.

PR #1998, opened 2024-03-07
Align sysroot mount options with SLE Micro, which in turn is applying…

davidcassany:
Align sysroot mount options with SLE Micro, which in turn is applying default mount options


davidcassany:
As of today the current default images for SLE Micro I could found they mount root and snapshots with
```
UUID=cf1b3a39-54c1-4f32-a370-b225e1dcc1c2 / btrfs ro 0 0
UUID=cf1b3a39-54c1-4f32-a370-b225e1dcc1c2 /.snapshots btrfs defaults,subvol=@/.snapshots 0 0
```
from `/etc/fstab`. This is also great as relies on system default rather than a static list of mount options.

frelon (approved):

PR #1999, opened 2024-03-08
Fix unterminated quoted string

frelon:
Fixes error in https://github.com/rancher/elemental-toolkit/actions/runs/8200605115/job/22427690109


davidcassany (approved):
:+1:

PR #2001, opened 2024-03-11
Switches tests to Ubuntu and KVM for tests

davidcassany:
With this change the performance of the integration tests improves a lot.


davidcassany:
Did some PoC runs here https://github.com/davidcassany/elemental-toolkit/pull/14 see the test-installer got executed in only 3min

frelon (approved):
Awesome :tada:

PR #2013, opened 2024-03-15
Change location of bootargs.conf

frelon:
Use the same location as grub.conf, /etc/elemental.

The grub config is backwards compatible and tries to source from /etc/cos/bootargs.conf as well.


pkg/features/embedded/grub-config/etc/elemental/grub.cfg
```diff
@@ -57,6 +57,12 @@ function set_loopdevice {
   loopback ${volume} ${1}
 }
 
+## Sources bootargs from the current volume
+function source_bootargs {
+  source (${volume})/etc/cos/bootargs.cfg
```
davidcassany, line 62: Isn't this failing if the file is not there? 🤔

davidcassany (approved):
LGTM

pkg/features/embedded/grub-config/etc/elemental/grub.cfg
frelon: Not when I tried it.. I was going to check if the file existed first but seems like it's not needed :shrug:

PR #2014, opened 2024-03-15
Use elemental instead of cos

frelon:
* Allow both cos.setup and elemental.setup in kernel cmdline.
* Use elemental.setup and elemental.disable in ISO grub config.


davidcassany (approved):
Nice!

Issue #1930, opened 2024-02-01
Move recovery kernel and initrd out of recovery image


davidcassany:
Consider extracting the kernel, initrd and bootarguments in EFI partition to:

* Get closer to systemd-boot (it might be relevant in the future)
* Get rid of the loopback requirement in grub.cfg
* Make grub.cfg equivalent regardless of btrfs subvolumes or loop devices.

Requires matching kernel, initrd and args with snapshots and probably find a mechanism to actually asses the loaded kernel and initrd matches the appropriate snapshot.

davidcassany:
I have been thinking further about it and I think that probably we could start by considering extracting the kernel and the initrd from the recovery image in a similar way we do the ISOs, this way the ISOs and RAW images would look almost the same. That would also allow us having squashfs compressed images for recovery and keep secure boot enabled.

frelon:
ADR here: https://github.com/rancher/elemental-toolkit/wiki/ADRs#0001-extract-recovery-kernel

PR #2016, opened 2024-03-18
Extract recovery kernel

### LINKED ISSUE #1864: Default to squashfs for recovery partition
Would be nice if do not require running `mkfs.ext4` with the `-d` flag as we have no clue on what happens on extended attributes and probably some other details. `mkfs.ext4` docs are not really specifying any further detail.

Related to https://github.com/rancher/elemental/issues/487

frelon:
This PR refactors the deployment of the recovery-system in our different code-branches according to the [ADR](https://github.com/rancher/elemental-toolkit/wiki/ADRs#0001-extract-recovery-kernel).

 - [x] build-iso
 - [x] build-disk
 - [x] install
 - [x] upgrade-recovery

Fixes #1864 #1930

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,7 @@ build-iso:
 	mkdir -p $(ROOT_DIR)/build
 	$(DOCKER) run --rm -v $(DOCKER_SOCK):$(DOCKER_SOCK) -v $(ROOT_DIR)/build:/build \
 		--entrypoint /usr/bin/elemental $(TOOLKIT_REPO):$(VERSION) --debug build-iso --bootloader-in-rootfs -n elemental-$(FLAVOR).$(ARCH) \
-		--local --platform $(PLATFORM) --squash-no-compression -o /build $(REPO):$(VERSION)
+		--local --platform $(PLATFORM) -o /build $(REPO):$(VERSION)
 
 .PHONY: build-disk
 build-disk:
@@ -99,7 +99,7 @@ build-disk:
 	$(DOCKER) run --rm -v $(DOCKER_SOCK):$(DOCKER_SOCK) -v $(ROOT_DIR)/build:/build \
 		--entrypoint /usr/bin/elemental \
 		$(TOOLKIT_REPO):$(VERSION) --debug build-disk --platform $(PLATFORM) --expandable -n elemental-$(FLAVOR).$(ARCH) --local \
-		--squash-no-compression -o /build --system $(REPO):$(VERSION)
+		-o /build --system $(REPO):$(VERSION)
 	qemu-img convert -O qcow2 $(ROOT_DIR)/build/elemental-$(FLAVOR).$(ARCH).raw $(ROOT_DIR)/build/elemental-$(FLAVOR).$(ARCH).qcow2
 	qemu-img resize $(ROOT_DIR)/build/elemental-$(FLAVOR).$(ARCH).qcow2 $(DISKSIZE) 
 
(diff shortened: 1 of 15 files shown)


pkg/elemental/elemental.go
```diff
@@ -470,6 +471,109 @@ func DeployImage(c types.Config, img *types.Image) error {
...
+	} {
+		if exist, _ := utils.Exists(cfg.Fs, file); !exist {
+			cfg.Logger.Debugf("File '%s' not found, skipping", file)
+			continue
+		}
+
+		target := filepath.Join(bootDir, filepath.Base(file))
+		if exist, _ := utils.Exists(cfg.Fs, target); exist {
+			cfg.Logger.Debugf("Removing old file %s", target)
+			err = cfg.Fs.Remove(target)
```
frelon, line 533: This is to make sure we deploy the newest files.

On error we could potentially leave the system unbootable because of a mix of kernel/initrd/grub-bootargs.

I think this is safe since we are updating from the active system, but we could consider deploying the artifacts into a temporary boot-dir and then moving it to `<recovery>/boot` once everything is deployed.

Makefile
```diff
@@ -90,7 +90,7 @@ build-iso:
 	mkdir -p $(ROOT_DIR)/build
 	$(DOCKER) run --rm -v $(DOCKER_SOCK):$(DOCKER_SOCK) -v $(ROOT_DIR)/build:/build \
 		--entrypoint /usr/bin/elemental $(TOOLKIT_REPO):$(VERSION) --debug build-iso --bootloader-in-rootfs -n elemental-$(FLAVOR).$(ARCH) \
-		--local --platform $(PLATFORM) --squash-no-compression -o /build $(REPO):$(VERSION)
+		--local --platform $(PLATFORM) -o /build $(REPO):$(VERSION)
```
davidcassany, line 93: Will it default to compressed squashfs images?

pkg/action/build-iso.go
```diff
@@ -161,7 +161,20 @@ func (b *BuildISOAction) ISORun() error {
...
+		return err
+	}
+
+	image := &types.Image{
+		Source: types.NewDirSrc(rootDir),
+		File:   filepath.Join(isoDir, constants.ISORootFile),
+		FS:     constants.SquashFs,
+	}
+
+	err = elemental.DeployRecoverySystem(b.cfg.Config, image, bootDir)
```
davidcassany, line 177: I really like this method :+1:

pkg/elemental/elemental.go
davidcassany: Yes touching bootloader binaries is tricky. In fact this is also a valid concern we currently have for upgrades too when using the bootloader install flag. I'd be happy to brainstorm a little around this topic, if someone faces issues in that area, even it is unlikely, would be quite frustrating. I'd be in favor to apply some more elaborated logic, the question here is which would be the desired behavior.

davidcassany (approved):
Nice! Great job :+1:

I am wondering if shouldn't we create an issue to discuss bootloader upgrade. Probably we should define some better approach to be more robust in case of failures.

Makefile
frelon: Only for our example isos/disks since I noticed we saved ~50MB on the final image which could speed up CI sligthly.

pkg/elemental/elemental.go
frelon: Agreed! I will create a ticket to stabilize the upgrade! It should probably include some integration tests as well.

PR #2017, opened 2024-03-19
Test qemu secure boot

frelon:
Add qemu settings for secure boot and add test in smoke suite to actually test that secure boot is enabled.


frelon:
Since the workflow is running from the main branch the updated values of `ELMNTL_FIRMWARE_VARS` is not used yet.

davidcassany (approved):
LGTM

PR #2018, opened 2024-03-20
make configurable boot-assessment checks

davidcassany:
Part of rancher/elemental#1263

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -227,7 +227,7 @@ func NewMountSpec() *types.MountSpec {
 			}, {
 				Mountpoint: constants.EfiDir,
 				Device:     fmt.Sprintf("PARTLABEL=%s", constants.EfiPartName),
-				Options:    []string{"rw", "defaults"},
+				Options:    []string{"ro", "defaults"},
 			},
 		},
 		Ephemeral: types.EphemeralMounts{
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -227,7 +227,7 @@ func NewMountSpec() *types.MountSpec {
 			}, {
 				Mountpoint: constants.EfiDir,
 				Device:     fmt.Sprintf("PARTLABEL=%s", constants.EfiPartName),
-				Options:    []string{"ro", "defaults"},
+				Options:    []string{"rw", "defaults"},
 			},
 		},
 		Ephemeral: types.EphemeralMounts{
--- a/pkg/features/embedded/boot-assessment/etc/systemd/system/elemental-boot-assessment.service
+++ b/pkg/features/embedded/boot-assessment/etc/systemd/system/elemental-boot-assessment.service
@@ -2,7 +2,8 @@
 Description=Elemental system configuration
 After=network-online.target systemd-logind.service
 Wants=network-online.target systemd-logind.service
-ConditionPathExists=/run/elemental/active_mode
+ConditionKernelCommandLine=elemental.health_check
+ConditionPathExists=!/run/elemental/recovery_mode
 
 StartLimitAction=reboot
 StartLimitIntervalSec=300
--- a/pkg/features/embedded/boot-assessment/etc/systemd/system/elemental-boot-assessment.service
+++ b/pkg/features/embedded/boot-assessment/etc/systemd/system/elemental-boot-assessment.service
@@ -1,0 +1,19 @@
+[Unit]
+Description=Elemental system configuration
+After=network-online.target systemd-logind.service
+Wants=network-online.target systemd-logind.service
+ConditionPathExists=/run/elemental/active_mode
+
+StartLimitAction=reboot
+StartLimitIntervalSec=300
+StartLimitBurst=5
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/sbin/elemental-boot-assessment
+Restart=on-failure
+RestartSec=30
+
+[Install]
+WantedBy=multi-user.target
--- a/pkg/features/embedded/boot-assessment/system/oem/08_boot_assessment.yaml
+++ b/pkg/features/embedded/boot-assessment/system/oem/08_boot_assessment.yaml
@@ -7,6 +7,16 @@
 
 name: "Boot assessment"
 stages:
+    rootfs.before:
+    - name: "Create upgrade failure sentinel if necessary"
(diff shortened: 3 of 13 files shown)


pkg/config/config.go
```diff
@@ -228,7 +228,7 @@ func NewMountSpec() *types.MountSpec {
 			}, {
 				Mountpoint: constants.EfiDir,
 				Device:     fmt.Sprintf("PARTLABEL=%s", constants.EfiPartName),
-				Options:    []string{"ro", "defaults"},
+				Options:    []string{"rw", "defaults"},
```
davidcassany, line 231: Having EFI writable is aligned with SL Micro and it also simplifies a lot the `08_boot_assessment.yaml` file. I have mixed feelings about this, I am not this is better.

pkg/features/embedded/boot-assessment/etc/systemd/system/elemental-boot-assessment.service
```diff
@@ -0,0 +1,19 @@
+[Unit]
+Description=Elemental system configuration
+After=network-online.target systemd-logind.service
+Wants=network-online.target systemd-logind.service
+ConditionPathExists=/run/elemental/active_mode
+
+StartLimitAction=reboot
+StartLimitIntervalSec=300
+StartLimitBurst=5
```
davidcassany, line 9: If there are more than 5 failures within a time frame of 5min this will trigger a reboot.

pkg/features/embedded/boot-assessment/usr/sbin/elemental-boot-assessment
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+declare checkResultsPath="/run/elemental/boot-assessment"
+declare checkersPath="/usr/libexec/elemental-checker"
```
davidcassany, line 4: Any script/executable included in this folder will be executed with  `check` argument. If it returns no error marks the script as passed (so it won't be executed again in any eventual retry), if it returns an error raises the error flag and continues with the rest of the checks. In case any of the executed scripts fails this script runs an `exit 1` which causes the service to fail and restart again after 30seconds.

pkg/features/embedded/boot-assessment/usr/sbin/elemental-boot-assessment
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
frelon, line 1: We could probably take this even further and add a `elemental health-check` command. Then we add some test-coverage to this code :+1:

pkg/features/embedded/boot-assessment/usr/sbin/elemental-boot-assessment
```diff
@@ -0,0 +1,38 @@
...
+}
+
+doBootAssessment
+if [ $? -ne 0 ]; then
+  echo 1 > "${upgradeFailureSentinel}"
+  exit 1
+fi
+
+grub2-editenv "${efiPath}/boot_assessment" set enable_boot_assessment=
+grub2-editenv "${efiPath}/boot_assessment" set boot_assessment_tentative=
```
davidcassany, line 38: If all checkers pass it clears the boot_assessment grub variables so on an eventual manual reboot grub won't consider it a failed upgrade.

frelon (approved):
Nice improvement to the boot-assessment!

davidcassany (commented):
If no checkers are provided boot-assessment behaves as it used to do, with the only real difference that the boot-assessment is verified as passed in a slightly later phase than it used to do (`boot`  yip's stage).

pkg/features/embedded/boot-assessment/usr/sbin/elemental-boot-assessment
davidcassany: Yes I was also thinking about it. But first wanted to actually verify the approach. I have also been thinking on including the boot-assessment logic in grub.cfg directly and just use a variable to enable it or not.

pkg/config/config.go
frelon: Yep, was also hesitant about this, but it is fairly common. :shrug:

pkg/features/embedded/boot-assessment/usr/sbin/elemental-boot-assessment
davidcassany: For I think I'll leave it as this simple script. Let's reconsider this once we have the boot assessment in place and functional all the way up to Micro images.

pkg/features/embedded/boot-assessment/system/oem/08_boot_assessment.yaml
```diff
@@ -0,0 +1,74 @@
...
+#   active system.
+
+name: "Boot assessment"
+stages:
+    after-install-chroot:
+    # The boot assessment logic is added here as part of a bootassessment.cfg file. This file is
+    # sourced in /grubcustom/custom.cfg file, which is only meant to source additional files.
+    # So far there is no regexp module in most grub2 efi signed images, hence it is not possible to
+    # to iterate over the configuration files in /grubcustom folder to load them all. Instead the
+    # pattern is to manually append additional sourced files in /grubcustom/custom.cfg.
```
davidcassany, line 19: I could not manage to find a proper way of doing it. I did not want to include this in grub.cfg because this is coupled with the boot-assessment feature and grub is provided by another independent feature. I like the idea of keeping the option of installing without any kind of boot-assessment logic.

My desired setup was defining a `/grubcustom` folder for which the main `grub.cfg` would source all the existing configuration files, this way we could easily include additional custom changes besides the `bootassessment.cfg`. However I could not find a way to iterate over files found in a path, this could be possible by using the `regexp` grub2 module, but this does not come included signed efi images by default.

davidcassany:
The fallback test new logic tests the following: upgrades to a broken system:

1. Upgrades to a broken system, the new OS can't boot which causes a reboot to passive
2. Rebooting to passive succeeds, then from there it upgrades to a system including a checker that always fails. So after actually rebooting to active will reboot again after 5 failed attempts to pass the checker.
3. The reboot sequence will then attempt to boot to the broken system, which will fail again. Then the boot-assessment logic in grub will select the next system, which the snapshot installed during installation.
4. The former installed image will successfully boot and detect it booted into passive.

frelon (approved):
Awesome, great job!

PR #2019, opened 2024-03-20
Update github.com/canonical/go-efilib

frelon:


davidcassany (approved):
Wow, we already missed many releases in between of that library and comes with a bump on few others.

LGTM

PR #2022, opened 2024-03-21
Fix passive snapshots listing in grub.cfg

davidcassany:


pkg/features/embedded/grub-config/etc/elemental/grub.cfg
```diff
@@ -90,10 +90,10 @@ menuentry "${display_name}" --id active {
 }
 
 for passive_snap in ${passive_snaps}; do
-  menuentry "${display_name} (snapshot ${passive_snap})" --id ${passive_snap} {
+  menuentry "${display_name} (snapshot ${passive_snap})" --id passive${passive_snap} ${passive_snap} {
     set mode=passive
     search --no-floppy --label --set=root ${state_label}
-    set_volume ${passive_snap}
```
davidcassany, line 96: This variable inside the menu is not expanded when looping, in fact it is expanded when choosing the actual menu item and it takes the value of the last assignement in the loop. Should have been passed as an argument to the `menuentry` call.

pkg/features/embedded/grub-config/etc/elemental/grub.cfg
```diff
@@ -90,10 +90,10 @@ menuentry "${display_name}" --id active {
 }
 
 for passive_snap in ${passive_snaps}; do
-  menuentry "${display_name} (snapshot ${passive_snap})" --id ${passive_snap} {
+  menuentry "${display_name} (snapshot ${passive_snap})" --id passive${passive_snap} ${passive_snap} {
```
davidcassany, line 93: Ensuring the ID is not set to a number, as then it turns to be confusing with the fallback values, which is just a counter of the menu items, starting from `0` at top. So if a menu is chosen by ID it will not definitely be a number. If a number is used, then menu counter applies.

frelon (approved):
Good catch!

PR #2023, opened 2024-03-21
Use ContainSubstring in secure boot test

frelon:
The output will contain a line-break, check that the string contains "SecureBoot enabled" instead.


davidcassany (approved):
LGTM

PR #2026, opened 2024-03-22
docs: fix build-iso example

fgiudici:
add the "--bootloader-in-rootfs" flag


frelon:
Very nice, we should make sure this is added in https://github.com/rancher/elemental-docs/pull/207 as well, since the long term plan has been to move the toolkit docs to https://elemental.docs.rancher.com

frelon (approved):

PR #2027, opened 2024-03-25
Iterate over passive snapshots on boot-assessment

davidcassany:
This commit prevents falling back to recovery system during the boot assessment process.


frelon (approved):

frelon (approved):
LGTM

PR #2028, opened 2024-03-25
Skip symlink if kernel already exists

frelon:
If kernel or initrd is named the default vmlinuz/initrd we skip the symlinking since the booting will work anyways.

### DIFF
--- a/pkg/elemental/elemental.go
+++ b/pkg/elemental/elemental.go
@@ -549,6 +549,11 @@ func DeployRecoverySystem(cfg types.Config, img *types.Image, bootDir string) er
 		"linux":   kernel,
 		"initrd":  initrd,
 	} {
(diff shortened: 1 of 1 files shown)


pkg/elemental/elemental.go
```diff
@@ -549,6 +549,11 @@ func DeployRecoverySystem(cfg types.Config, img *types.Image, bootDir string) er
 		"linux":   kernel,
 		"initrd":  initrd,
 	} {
+		// if kernel/initrd is actually named "vmlinuz"/"initrd" we just skip the symlink part.
+		if kernel == name || initrd == name {
+			continue
```
ldevulder, line 554: It might be a good idea to add a `Debugf` call just to simplify the debugging in case of issue:
```suggestion
			cfg.Logger.Debugf("No symlink needed, keeping exising %s file", name)
			continue
```

ldevulder (approved):
Just a small suggestion, but good to me!

PR #2030, opened 2024-03-25
Backwards compatible recovery boot configuration

frelon:
The recovery system has been changed to extract the kernel/initrd and bootargs.cfg in order to not rely on loopback devices.

This commit adds backwards compatibility to the grub configuration in case we have upgraded the bootloader but not the recovery system on a host.


davidcassany (approved):
LGTM

PR #2032, opened 2024-03-25
Snapshottable recovery system

### LINKED ISSUE #2021: Snapshot recovery upgrade
The new behavior of deploying the recovery system is prone to errors if an error during the upgrade.

The current method of overwriting the kernel/initrd/bootargs.cfg and symlinks should be switched to a more snapshotty approach deploying the recovery system to a new directory and switching over to use that directory once all files are deployed.

This would also help clean up old files not overwritten during the upgrade.

frelon:
Deploy the entire recovery system to the same folder (kernel, initrd and rootfs).

During upgrade deploy to a transitional folder and then switch it with the current recovery system and then delete the old one.

This makes sure we clean up old recovery systems and don't risk mixing systems during upgrade.

Fixes #2021

### DIFF
--- a/pkg/action/upgrade-recovery.go
+++ b/pkg/action/upgrade-recovery.go
@@ -84,15 +84,15 @@ func NewUpgradeRecoveryAction(config *types.RunConfig, spec *types.UpgradeSpec, 
 	return u, nil
 }
 
-func (u UpgradeRecoveryAction) Info(s string, args ...interface{}) {
+func (u UpgradeRecoveryAction) Infof(s string, args ...interface{}) {
 	u.cfg.Logger.Infof(s, args...)
 }
 
-func (u UpgradeRecoveryAction) Debug(s string, args ...interface{}) {
+func (u UpgradeRecoveryAction) Debugf(s string, args ...interface{}) {
 	u.cfg.Logger.Debugf(s, args...)
 }
 
-func (u UpgradeRecoveryAction) Error(s string, args ...interface{}) {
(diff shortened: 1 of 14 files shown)


pkg/action/upgrade-recovery.go
```diff
@@ -146,48 +146,73 @@ func (u *UpgradeRecoveryAction) Run() (err error) {
...
+
 		return err
 	}
 
+	// Remove old boot-dir when new recovery system is in place
+	if ok, _ := utils.Exists(u.cfg.Fs, oldBootDir); ok {
+		err = u.cfg.Fs.RemoveAll(oldBootDir)
+		if err != nil {
+			u.Errorf("failed removing old recovery image: %s", err.Error())
+			return err
```
davidcassany, line 197: I am wondering if we should actually return on this error. If we return here this ends up in a very weird state, as it reports error and abruptly finishes but the upgrade is indeed already done 🤔 On reboot it will be a there actually.

Probably once the swap is successfully done we should not error out on errors to give chance to continue with any remaining step. 

Anyway this might be a separate issue as this would also apply for regular upgrades. At that point we could eventually collect relevant errors but continue with the execution until the end.

davidcassany (approved):
LGTM

Looks like some unit tests needs to be arranged though.

pkg/action/upgrade-recovery.go
frelon: It's a good point, we should also remove the oldBootDir on newer upgrades if it already exists!

PR #2033, opened 2024-03-26
Fix the fix

frelon:
The previous fix does not compare filenames, only a path to a filename which will always be false.

This commit corrects the comparison and adds a log when skipping.


davidcassany (approved):
LGTM

PR #2036, opened 2024-03-28
Remove empty mount paths

### LINKED ISSUE #2035: Persistent paths are mounted read-only
**elemental-toolkit version:** `v2.0.0`

**Describe the bug**
The persistent paths defined in mount-layout are all mounted read-only:

```yaml
name: "Elemental Rootfs Layout Settings"
stages:
  rootfs:
    - if: '[ ! -f "/run/elemental/recovery_mode" ]'
      name: "Layout configuration"
      environment_file: /run/elemental/mount-layout.env
      environment:
        OVERLAY: "tmpfs:25%"
        RW_PATHS: "/var /etc /srv"
        PERSISTENT_STATE_PATHS: >-
          /etc/systemd
          /etc/kubernetes 
          /etc/rancher
        PERSISTENT_STATE_BIND: "true"
```

**To Reproduce**
The issue with the snippet above is a trailing space on the **second** path entry: `/etc/kubernetes `.
`PERSISTENT_STATE_BIND` also has to be set, this can only be reproduced in bind mode.
It will result in an empty mount: 

```
Mar 27 16:14:24 host-1 elemental[888]:   Persistent: types.PersistentMounts{
Mar 27 16:14:24 host-1 elemental[888]:     Mode: "bind",
Mar 27 16:14:24 host-1 elemental[888]:     Paths: []string{
Mar 27 16:14:24 host-1 elemental[888]:       "",
Mar 27 16:14:24 host-1 elemental[888]:       "/etc/systemd",
Mar 27 16:14:24 host-1 elemental[888]:       "/etc/rancher",
.
.
```

Ultimately this will lead to all paths mounted in read only: 

```
/dev/vda5 on /etc/kubernetes type ext4 (ro,relatime)
/dev/vda5 on /etc/systemd type ext4 (ro,relatime)
/dev/vda5 on /etc/rancher type ext4 (ro,relatime)
```

**Expected behavior**
Trimming the blank entry should be a safe behavior. It would be also best to print a warning so that the user can fix the typo eventually.

frelon:
When running mount command with a layout file that contains spaces, those spaces will be converted to empty paths.

This commit removes those paths in the Sanitize method for the MountSpec.

Fixes #2035

### DIFF
--- a/pkg/action/mount.go
+++ b/pkg/action/mount.go
@@ -25,21 +25,24 @@ import (
 	"strings"
 
 	"github.com/hashicorp/go-multierror"
+
 	"github.com/rancher/elemental-toolkit/v2/pkg/constants"
 	"github.com/rancher/elemental-toolkit/v2/pkg/types"
 	"github.com/rancher/elemental-toolkit/v2/pkg/utils"
 )
 
-const overlaySuffix = ".overlay"
-const labelPref = "LABEL="
-const partLabelPref = "PARTLABEL="
-const uuidPref = "UUID="
-const devPref = "/dev/"
-const diskBy = "/dev/disk/by-"
-const diskByLabel = diskBy + "label"
-const diskByPartLabel = diskBy + "partlabel"
-const diskByUUID = diskBy + "uuid"
-const runPath = "/run"
+const (
+	overlaySuffix   = ".overlay"
+	labelPref       = "LABEL="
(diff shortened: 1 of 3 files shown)


anmazzotti (approved):
Looks good!

I wonder if we should issue a warning, as a user I'd like to know, but this can also be fixed silently as does have no impact.

PR #2040, opened 2024-04-10
Make squashfs defaults more consistent

davidcassany:
This PR makes ensures squashfs compression can be disabled and leaves the default compression to what used to be the case for `squash-no-compression: true`. Default compression then defaults to default mksquashfs compression (gzip) with a block of 1024k.

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -19,7 +19,6 @@ package constants
 import (
 	"os"
 	"path/filepath"
-	"runtime"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


davidcassany:
The result is that where ever `squash-no-compression` option was used it can be removed with the same result. Also the tests that where using the no compression option now should really result in non compressed images.

pkg/constants/constants.go
```diff
@@ -265,20 +264,12 @@ func GetCloudInitPaths() []string {
...
 func GetDefaultSquashfsCompressionOptions() []string {
-	options := []string{"-comp", "xz", "-Xbcj"}
-	// Set the filter based on arch for best compression results
-	if runtime.GOARCH == "arm64" {
-		options = append(options, "arm")
-	} else {
-		options = append(options, "x86")
-	}
-	return options
+	return []string{"-b", "1024k"}
```
davidcassany, line 272: Use default mksquashfs compression with a block of 1024k bytes

fgiudici (approved):
looks good here!

frelon (approved):
LGTM!

PR #2046, opened 2024-04-17
Document cloud-config-defaults for SELinux support

anmazzotti:
This should actualize the documentation. I believe the `cloud-config` feature is deprecated, as it was split between `cloud-config-essentials` and `cloud-config-defaults`, where the defaults includes SELinux support.


kkaempf:
This PR is lacking context

frelon (approved):

PR #2048, opened 2024-04-18
add symlink for missing binary on ubuntu 22.04+

furlongm:
grub2-editenv does not exist on ubuntu 22.04+ - it is just called grub-editenv

Currently it fails like so:

```
ERRO[2024-04-17T22:54:28-04:00] sh: 1: grub2-editenv: not found
 : failed to run grub2-editenv /oem/grubenv set boot_assessment_check=yes: exit status 127
ERRO[2024-04-17T22:54:28-04:00] sh: 1: grub2-editenv: not found
 : failed to run grub2-editenv /oem/grubenv unset last_boot_attempt: exit status 127
ERRO[2024-04-17T22:54:28-04:00] 2 errors occurred:
        * failed to run grub2-editenv /oem/grubenv set boot_assessment_check=yes: exit status 127
        * failed to run grub2-editenv /oem/grubenv unset last_boot_attempt: exit status 127
```

The symlink fixes the issue

### DIFF
--- a/examples/orange/Dockerfile
+++ b/examples/orange/Dockerfile
@@ -52,6 +52,9 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-ins
   podman \
   xz-utils
 
+# Symlink grub2-editenv
+RUN ln -sf /usr/bin/grub-editenv /usr/bin/grub2-editenv
+
 # Create non FHS paths
 RUN mkdir -p /oem /system
 


frelon (approved):
Very nice! I think we could also check which `editenv` command exists and use that like we do in the bootloader code, but I will merge this in the meantime! :+1:

davidcassany:
> Very nice! I think we could also check which `editenv` command exists and use that like we do in the bootloader code, but I will merge this in the meantime! 👍

Interesting I guess the problem here is for the cloud-config of boot assessment, in the golang client we already check if grub2-editenv or grub-editenv should be used. I guess there is no other simple way than adding some logic in the cloud-config file to actually fallback to one or the other when a miss is detected.

PR #2049, opened 2024-04-18
Suggest SELinux enforcing config

anmazzotti:
Turns out that setting `SELINUX=enforcing` in `/etc/selinux/config` is ineffective if `enforcing=0` is passed as kernel arg.
The only way that I managed to successfully enforce SELinux is to do it through `setenforce 1` at boot.
`setsebool secure_mode_policyload on` can be also set to forbid further changes.

Not a great solution, but the only one that works today.


anmazzotti:
Closing in favor of the [better solution](https://github.com/rancher/elemental-toolkit/pull/2052).

PR #2052, opened 2024-04-22
Enable SELinux for tumbleweed and green

### LINKED ISSUE #2054: elemental-toolkit tests suite in enforce mode for TW

davidcassany:
This PR uses SELinux in enforce mode for the active/passive systems and in permissive mode for recovery and ISO systems.

The elemental labelling code changed slightly,  basically the change is that now it is also executed as part of the close transaction step once the new root is already rsynced. In addition it also tries to label from the outer system any eventual mountpoint that was labeled in a chroot env (`/dev`, `/proc`, `/sys`, etc.).

Current unsolved issues:

* In github actions the host has SELinux enabled and relabeling the new root within a container turns to fail (see build disk and build iso logs in GH)... as a consequence the disk and iso images are not properly labeled, hence they are not really usabled in enforce mode. This does not happen in a host without the `container-selinux` policy module. This has impact on our capability to build ISOs and Disks inside a k8s cluster using unprivileged pods.

* Running upgrades from the recovery system (`elemental upgrade` command) is not functional if the target is in btrfs. There seams to be an issue with selinux labels when mounting the `.snapshots` subvolume preventing snapper to properly operate. One option would be to rethink the btrfs snapshotter env in recovery mode, instead of mounting `.snapshots` in `<some/root>/.snapshots` just use `/.snapshots` so any eventual relabel is done with the appropriate path.

Fixes #2054

### DIFF
--- a/pkg/elemental/elemental.go
+++ b/pkg/elemental/elemental.go
@@ -363,6 +363,11 @@ func CreateImageFromTree(c types.Config, img *types.Image, rootDir string, prelo
 			return err
 		}
 
+		err = SelinuxRelabel(c, rootDir)
+		if err != nil {
+			c.Logger.Warnf("failed SELinux labelling at %s: %v", rootDir, err)
+		}
+
 		err = utils.CreateSquashFS(c.Runner, c.Logger, rootDir, img.File, c.SquashFsCompressionConfig)
 		if err != nil {
 			c.Logger.Errorf("failed creating squashfs image for %s: %v", img.File, err)
@@ -398,6 +403,12 @@ func CreateImageFromTree(c types.Config, img *types.Image, rootDir string, prelo
 				c.Logger.Errorf("failed creating dir structure: %v", err)
 				return err
 			}
+
+			err = ApplySELinuxLabels(c, img.MountPoint, nil)
+			if err != nil {
+				c.Logger.Errorf("failed SELinux labelling at %s: %v", img.MountPoint, err)
+				return err
+			}
 		}
 	}
 	return err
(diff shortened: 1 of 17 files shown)


pkg/features/embedded/cloud-config-defaults/system/oem/04_accounting.yaml
```diff
@@ -14,4 +14,4 @@ stages:
          entity: |
             kind: "shadow"
             username: "root"
-            password: "cos"
+            password: "root"
```
frelon, line 17: Nice :+1:

anmazzotti:
@davidcassany FYI on Tumbleweed a `patterns-microos-selinux` is available. I [used](https://github.com/rancher/elemental/blob/main/Dockerfile.dev.os#L63) it for the dev image.

pkg/features/embedded/cloud-config-defaults/system/oem/04_accounting.yaml
davidcassany: I was also considering using `elemental` ;) I'd be fine with anything to be honest, no strong opinion at all

fgiudici (dismissed):

pkg/features/embedded/cloud-config-defaults/system/oem/04_accounting.yaml
davidcassany: let me put the password change in a different and independent PR ;)

frelon (approved):
Very nice, this should be enough to start experimenting with selinux in the cluster provisioning!

pkg/elemental/elemental.go
```diff
@@ -659,19 +670,20 @@ func CopyCloudConfig(c types.Config, path string, cloudInit []string) (err error
...
 
 // SelinuxRelabel will relabel the system if it finds the binary and the context
-func SelinuxRelabel(c types.Config, rootDir string) error {
-	policyFile, err := utils.FindFile(c.Fs, rootDir, filepath.Join(cnst.SELinuxTargetedPolicyPath, "policy.*"))
+func SelinuxRelabel(c types.Config, rootDir string, extraPaths ...string) error {
 	contextFile := filepath.Join(rootDir, cnst.SELinuxTargetedContextFile)
 	contextExists, _ := utils.Exists(c.Fs, contextFile)
 
-	if err == nil && contextExists && c.Runner.CommandExists("setfiles") {
+	if contextExists && c.Runner.CommandExists("setfiles") {
```
frelon, line 677: maybe we should check that rootDir is RW as well? might be enough to get the recovery test green.

davidcassany: It is RW... disable selinux and  the test will pass, I believe the snapper error msg is just misleading, it could not write due to some denial, but not because of being RO.

PR #2053, opened 2024-04-22
Move green to tumbleweed

davidcassany:
This PR is moving the green flavor to be based on TW instead of Leap. Leap 15.x series are getting old and full test suite should be executed in a more up to date system.

This also gives the chance to include SELinux support for the green tests.


anmazzotti (approved):
Looks good!

fgiudici (approved):

PR #2055, opened 2024-04-23
Change default root password

davidcassany:


tests/vm/sut.go
```diff
@@ -137,9 +137,9 @@ func (s *SUT) ChangeBoot(b string) error {
 
 	switch b {
 	case Active:
-		bootEntry = Cos
+		bootEntry = "active"
```
davidcassany, line 140: Just addressing some old leftovers

tests/vm/sut.go
```diff
@@ -137,9 +137,9 @@ func (s *SUT) ChangeBoot(b string) error {
...
 	case Passive:
-		bootEntry = "fallback"
+		bootEntry = "1"
```
davidcassany, line 142: same as above, just addressing some old leftovers

frelon (dismissed):
Nice!

pkg/features/embedded/cloud-config-defaults/system/oem/04_accounting.yaml
```diff
@@ -14,4 +14,4 @@ stages:
          entity: |
             kind: "shadow"
             username: "root"
-            password: "cos"
+            password: "elemental"
```
anmazzotti, line 17: Nice!

davidcassany:
OK, just realized this requires some more love, if the password changes across versions the test suite will also fail. I think the proper solution is once an for all setting the authorized ssh keys as part of the cloud config in tests, this way we could eventually connect without needing a password at all.

make/Makefile.test
```diff
@@ -13,7 +13,7 @@ endif
 
 .PHONY: test-active
 test-active: prepare-test
-	VM_PID=$$(scripts/run_vm.sh vmpid) go run $(GINKGO) $(GINKGO_ARGS) ./tests/wait-active
+	VM_PID=$$(scripts/run_vm.sh vmpid) COS_PASS=cos go run $(GINKGO) $(GINKGO_ARGS) ./tests/wait-active
```
davidcassany, line 16: The upgrade test requires the legacy password because it boots with an older disk, so the password needs to be added for `test-active` and `test-upgrade`. It always tries to login by the authorized key first and if not it fallbacks to the password method.

I could not find an easy way at ssh package level to set multiple passwords to try.

frelon (approved):
Awesome!

PR #2057, opened 2024-04-23
Refactor selinux method

frelon:
The passed in raiseError can be removed and the error can be checked from the callee.


davidcassany (approved):
Thanks! Nice cleanup some ugly leftover of mine 🤭

PR #2058, opened 2024-04-24
Nightly retry pull

frelon:
Nigthly fails, might be timing related. Retry the pull-toolkit step 3 times if not successful with 3 second sleep between pulls.


davidcassany (approved):
LGTM

PR #2061, opened 2024-04-30
Fix stackable sources for build-iso

davidcassany:
This PR fixes the use of `--overlay-*` flags in build-iso command. With the latest changes the were using the `--delete` option in rsync to dump the sources to be stacked on an already existing source. This PR distinguishes a couple of use cases the former `DumpSource` method equivalent is is `MirrorRoot` now, as this includes `--delete` option for rsync and the creation of essential root-tree mountpoint dirs such as `/dev`, `/sys`, `/proc`... I noted this as using the `overlay-iso` flag in an ISO lead to an image including empty `/dev`, `/tmp`, `/proc`, etc.

In addition this PR also removes the `DeployImage` that was not used anywhere in code, this was just a leftover.

### DIFF
--- a/pkg/elemental/elemental.go
+++ b/pkg/elemental/elemental.go
@@ -542,15 +542,19 @@ func DeployRecoverySystem(cfg types.Config, img *types.Image) error {
 	return nil
 }
 
-// DumpSource dumps the imgSrc data to target. Mirror arguments sets the perfectly mirror source to target regardless if
-// some data is already present in target
-func DumpSource(c types.Config, target string, imgSrc *types.ImageSource, mirror bool) error { // nolint:gocyclo
+// DumpSource dumps the imgSrc data to target. SyncFunc argument is the function used to synchronize file or directory
+// sources (unused for contaier images), defaults to utils.SyncData if nil provided.
+func DumpSource(
+	c types.Config, target string, imgSrc *types.ImageSource,
+	syncFunc func(
+		l types.Logger, r types.Runner, f types.FS, src string, dst string, excl ...string,
+	) error,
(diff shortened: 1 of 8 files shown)


pkg/elemental/elemental.go
```diff
@@ -585,11 +542,17 @@ func DeployRecoverySystem(cfg types.Config, img *types.Image) error {
...
 
-// DumpSource sets the image data according to the image source type
-func DumpSource(c types.Config, target string, imgSrc *types.ImageSource) error { // nolint:gocyclo
+// DumpSource dumps the imgSrc data to target. Mirror arguments sets the perfectly mirror source to target regardless if
+// some data is already present in target
+func DumpSource(c types.Config, target string, imgSrc *types.ImageSource, mirror bool) error { // nolint:gocyclo
 	var err error
 	var digest string
 
+	syncFunc := utils.SyncData
```
frelon, line 551: Since DumpSource is only used from build-iso and MirrorData method I would even suggest to take syncFunc as a parameter. Although it feels kinda weird since it is not used if the source is a container image...

davidcassany: In fact this was my first attempt, but since the signature of this method is big I though this was probably too verbose to put it as a parameter. But you got a point, probably should be set as a parameter and just hint in method comment this is most likely for `utils.SyncData` or `utils.MirrorData`.

davidcassany: @frelon moved it as a parameter, makes the code a bit harder to read but more flexible and without adding extra logic.

frelon (approved):

PR #2062, opened 2024-04-30
Update embedded features docs

frelon:
Adds grub-default-bootargs, boot-assessment and autologin to the list of
features.


anmazzotti (approved):
LGTM!

PR #2065, opened 2024-05-06
Add branding for live CD

frelon:
The other modes are covered in 03_branding.yaml, but not the live_mode.

This commit adds a short motd and hostname when booting from live CD.


davidcassany (approved):
LGTM

PR #2068, opened 2024-05-07
Define layout with yaml format

davidcassany:


davidcassany:
Using the yaml format instead of the environment variables makes the configuration more explicit. Even in this case it mostly assumes all defaults. In fact I am wondering if the actual defaults are actually appropriate, probably would be better having them as part of the default yaml so they are explicit somwhere.

frelon (approved):

PR #2070, opened 2024-05-08
SELinux relabel on boot

frelon:
This commit introduces a new command 'elemental relabel' for relabeling
files and directories.

The command is invoked during initramfs stage in the new optional
'selinux' feature.

In essence it runs setfiles in mounted persistent and ephemeral
directories in order to circumvent a bug with overlayfs and
selinux copy-up on xattrs.

During mount we also put a list of persistent+ephemeral directories in
/run/systemd/extra-relabel.d/elemental.layout in order to make systemd
relabel the directories before loading the policy.


davidcassany (commented):
Nice! Looks quite good.

I miss some unit tests for the relabel command and the WriteSelinuxRelabelFile method. Shouldn't be hard to cover those.

Moreover I think the relabel cloud-config `10_selinux.yaml` I don't think is functional after my changes of the mount setup. Now the mount command setup is written in `/etc/elemental/config.d` in initramfs, which is not visible in `initramfs` stage, as it runs chrooted to the new root. 

Wouldn't it be easier to just omit the `relabel` command and run a script defined in `10_selinux.yaml` that reads the relabel file and relabels those paths and once the kernel is fixed we just drop this `10_selinux.yaml`. What you think?

On the other side I think it would also be good to read the mount configuration from `/run/elemental/` instead of `/etc/elemental`, this way we can see the mount setup after the switch root too.

frelon:
> Nice! Looks quite good.
> 
> 
> 
> I miss some unit tests for the relabel command and the WriteSelinuxRelabelFile method. Shouldn't be hard to cover those.
> 
> 
> 
> Moreover the relabel cloud-config `10_selinux.yaml` I don't think is functional after my changes of the mount setup. Now the mount command setup is written in `/etc/elemental/config.d` in initramfs, which is not visible in `initramfs` stage, as it runs chrooted to the new root. 
> 
> 
> 
> Wouldn't it be easier to just omit the `relabel` command and run a script defined in `10_selinux.yaml` that reads the relabel file and relabels those paths and once the kernel is fixed we just drop this `10_selinux.yaml`. What you think?
> 
> 
> 
> On the other side I think it would also be good to read the mount configuration from `/run/elemental/` instead of `/etc/elemental`, this way we can see the mount setup after the switch root too. Should be hard to set, I can do it in a separate PR.

Yep, good idea to use the relabel file to run the setfiles.. I will change that and then we can remove the relabel command!

PR #2071, opened 2024-05-08
Bootloader-in-rootfs=true

frelon:
Change the default value of bootloader-in-rootfs flag for build-iso
command to true.

### DIFF
--- a/cmd/build-iso.go
+++ b/cmd/build-iso.go
@@ -141,7 +141,7 @@ func NewBuildISO(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 	c.Flags().String("overlay-iso", "", "Path of the overlayed iso data")
 	c.Flags().String("label", "", "Label of the ISO volume")
 	c.Flags().String("extra-cmdline", "", fmt.Sprintf("Extra kernel cmdline (defaults to '%s')", constants.ISODefaultExtraCmdline))
-	c.Flags().Bool("bootloader-in-rootfs", true, "Fetch ISO bootloader binaries from the rootfs")
+	c.Flags().Bool("bootloader-in-rootfs", false, "Fetch ISO bootloader binaries from the rootfs")
 	c.Flags().Var(firmType, "firmware", "Firmware to install, only 'efi' is currently supported")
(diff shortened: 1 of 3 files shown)


cmd/build-iso.go
```diff
@@ -139,7 +139,7 @@ func NewBuildISO(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 	c.Flags().String("overlay-uefi", "", "Path of the overlayed uefi data")
 	c.Flags().String("overlay-iso", "", "Path of the overlayed iso data")
 	c.Flags().String("label", "", "Label of the ISO volume")
-	c.Flags().Bool("bootloader-in-rootfs", false, "Fetch ISO bootloader binaries from the rootfs")
+	c.Flags().Bool("bootloader-in-rootfs", true, "Fetch ISO bootloader binaries from the rootfs")
```
davidcassany, line 142: I am not sure whats the meaning of this flag then... I mean if one uses the flag `--bootloader-in-rootfs` what happens then? is it still set to `true`? If so then I'd say we can mark this flag as deprecated as the only way to set it to `false` would be explicitly stating so from the config yaml.

Also setting a non empty (empty is false for a boolean) default here is tricky, because it means it will always overwrite whatever is in the yaml configuration if I am not wrong. The sequence is, we create an object with our defaults, then we apply the yaml on top (only non empty values) and finally we apply the command line flags (only defined ones). These are all guesses, I guess we should define a test using the `bootloader-in-rootfs` field in the yaml manifest and check if it can actually overwrite the default command value.

Alternatively, if using `--bootloader-in-rootfs` flag turns this value to `false` then it is a bit misleading, isn't it?

I do like the idea of making this the default behavior though. Just need to make sure the behavior is then sane.

frelon: True, after playing around with this a bit more the behavior I wanted is that when --bootloader-in-rootfs is not specified it is actually defaulting to true. The flag can be set to false using the config-file or `--bootloader-in-rootfs=false`

I pushed a new commit that sets the default value in `pkg/config/config.go` instead.

davidcassany (approved):

PR #2072, opened 2024-05-08
Add extra-cmdline flag to build-iso command

frelon:
The extra-cmdline can be used to customize the kernel commandline used
for the ISO.

Some arbitrary flags were moved to the extra-cmdline in order to support
overriding security and consoles.

### DIFF
--- a/cmd/build-iso.go
+++ b/cmd/build-iso.go
@@ -140,7 +140,7 @@ func NewBuildISO(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 	c.Flags().String("overlay-uefi", "", "Path of the overlayed uefi data")
 	c.Flags().String("overlay-iso", "", "Path of the overlayed iso data")
 	c.Flags().String("label", "", "Label of the ISO volume")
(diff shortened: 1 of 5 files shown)


cmd/build-iso.go
```diff
@@ -139,6 +140,7 @@ func NewBuildISO(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 	c.Flags().String("overlay-uefi", "", "Path of the overlayed uefi data")
 	c.Flags().String("overlay-iso", "", "Path of the overlayed iso data")
 	c.Flags().String("label", "", "Label of the ISO volume")
+	c.Flags().String("extra-cmdline", constants.ISODefaultExtraCmdline, "Extra kernel cmdline")
```
davidcassany, line 143: I'd keep this default empty and include the default in `pkg/config/config.go` like other flags. This way the default is applied either you provide a command line flag or you define it in the yaml manifest file.

davidcassany (approved):
Looks good to me, just a small details regarding where to set the default value.

cmd/build-iso.go
frelon: Good point,  I will update :+1:

PR #2073, opened 2024-05-08
Move mount config file to /run/elemental

davidcassany:
This makes possible to see the mount config file in  `initramfs` stage or even after switching root. Makes no functional difference but it can be handy specially for debugging.


frelon (approved):
Really good idea!

PR #2074, opened 2024-05-10
SELinux relabel on boot

frelon:
In this commit we add selinux support to the mount command.

During mount we put a list of persistent+ephemeral directories in
/run/systemd/extra-relabel.d/elemental.layout in order to make systemd
relabel the directories before loading the policy.

We also try to chroot into the new sysroot and run setfiles using a find
wrapper to set on deepest files first.


davidcassany (approved):
Nice, looks good to me :+1:

PR #2078, opened 2024-05-13
Refine relabel step

davidcassany:


cmd/config/config.go
```diff
@@ -286,6 +286,15 @@ func ReadMountSpec(r *types.RunConfig, flags *pflag.FlagSet) (*types.MountSpec,
 }
 
 func applyKernelCmdline(r *types.RunConfig, mount *types.MountSpec) error {
+	lsm, err := r.Config.Fs.ReadFile("/sys/kernel/security/lsm")
```
davidcassany, line 289: SELinux could set by default without needing any kernel param, hence I think is is just a safer test to check if SELinux is enabled within this kernel execution or not.

pkg/action/mount.go
```diff
@@ -420,6 +421,13 @@ func SelinuxRelabel(cfg *types.RunConfig, spec *types.MountSpec) error {
 
 	paths := []string{}
 	paths = append(paths, spec.Ephemeral.Paths...)
+	for _, vol := range append(spec.Volumes, &spec.Persistent.Volume) {
```
davidcassany, line 424: This is to also relabel `/oem` and other eventual volumes mounted as RW.

pkg/action/mount.go
```diff
@@ -434,12 +442,7 @@ func SelinuxRelabel(cfg *types.RunConfig, spec *types.MountSpec) error {
 		return nil
 	}
 
-	extraMounts := map[string]string{
-		"/run":  "/run",
-		"/proc": "/proc",
-	}
-
-	return utils.ChrootedCallback(&cfg.Config, spec.Sysroot, extraMounts, func() error {
+	return utils.ChrootedCallback(&cfg.Config, spec.Sysroot, nil, func() error {
```
davidcassany, line 445: There was no need to bind mount `/proc` as this is already bind mounted as part of any chroot (we bind mount, at least, `/sys`, `/dev` and `/proc` for any chrooted call). There is no need to relabel `/run` as this happens anyway when SELinux policy is loaded.

frelon: Yep, `/run` was mounted because in an earlier iteration the relabel file was written in the chroot, and after moving the writing out I forgot to also remove the extra mounts, good catch! :+1:

pkg/action/mount.go
frelon: Nice, I was thinking of this, but would it not impact the persistent volume as well?

cmd/config/config.go
frelon: Looks good! I would move this out of the `applyKernelCmdline` method tho..

frelon (approved):
Some small comments, LGTM!

pkg/action/mount.go
davidcassany: Well, the idea is to also relabel the persistent volume. The only case in which is not relabeled is when this is mounted under `/run`. I have not fully tested it yet, but I'd like to be capable to set the persistent volume mountpoint to `/var`, in that case we should also relabel `/var`.

To be honest to it looks like we might need to revisit this part as long as we gain experience and encounter new corner cases. It is hard to generalize.

PR #2080, opened 2024-05-14
Relabel presistent paths only at first boot

davidcassany:
With this commit any RW paths gets a marker file added `.relabelled` to acknowledge this path got relabelled. At relabelling time if this marker file is found the path is omitted at relabelling stage. The idea is to only relabel persistent paths at first boot. This prevents relabelling persistent paths on each boot and gives the opportunity to select which paths to relabel by removing the markers.

Fixes https://github.com/rancher/elemental/issues/1417

### DIFF
--- a/pkg/action/mount.go
+++ b/pkg/action/mount.go
@@ -419,16 +419,7 @@ func SelinuxRelabel(cfg *types.RunConfig, spec *types.MountSpec) error {
 		return err
 	}
 
-	paths := []string{}
-	paths = append(paths, spec.Ephemeral.Paths...)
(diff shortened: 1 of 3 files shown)


davidcassany:
This is slightly inspired by https://github.com/openSUSE/microos-tools/blob/master/selinux/98selinux-microos/selinux-microos-relabel.sh

I am even wondering if we should expand `mount.selinux-relabel` config option and instead of making it a boolean, define it as a string and except values like `no|once|always` or something similar. @frelon what you think?

frelon (approved):
Looks good to me! I was also thinking about putting this in the install-state yaml, but this will probably be more robust!

PR #2081, opened 2024-05-15
Remove the work around for bsc#1210690 and simplify code

davidcassany:
The expectation is that this also improves the situation of https://github.com/rancher/elemental/issues/1417 

Without this workaround the relabeling step is way lighter and it shouldn't take that long.


frelon (approved):
LGTM

Issue #2083, opened 2024-05-16
library/rpi 


bobmorane83:
Hi all,

**elemental-toolkit version:**
❯ docker run -it --rm ghcr.io/rancher/elemental-toolkit/elemental-cli:latest version
v1.3.0+g56b2723

**CPU architecture, OS, and Version:**
On my RPi : Linux pi 6.6.28+rpt-rpi-v8 #1 SMP PREEMPT Debian 1:6.6.28-1+rpt1 (2024-04-22) aarch64 GNU/Linux
Same on my Mac M1

**Describe the bug**
❯ docker run -it --rm ghcr.io/rancher/elemental-toolkit/elemental-cli:latest build-iso --config-dir=./ rpi
INFO[2024-05-16T13:49:52Z] Starting elemental version v1.3.0            
INFO[2024-05-16T13:49:52Z] Reading configuration from './'              
INFO[2024-05-16T13:49:52Z] Preparing squashfs root (1 source)...        
INFO[2024-05-16T13:49:52Z] Copying rpi:latest source...                 
ERRO[2024-05-16T13:50:05Z] Failed installing OS packages: GET https://index.docker.io/v2/library/rpi/manifests/latest: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:library/rpi Type:repository]] 

**To Reproduce**
❯ docker run -it --rm ghcr.io/rancher/elemental-toolkit/elemental-cli:latest build-iso --config-dir=./ rpi

I'am trying to generate a Raspberry Pi 4 image to boot on, but seems that the repositoty is not available anymore.

Did I missed something ?
Any clue ?
Thanks,
BR,

frelon:
Hi @bobmorane83!

There is a separate make target for building an RPI disk.

Try running 'make ARCH=arm64 build build-os build-rpi-disk' and it should build the toolkit, green OS container image and finally a raw image that can be written to an sd-card and booted!

Let us know how it goes! 👍

bobmorane83:
Thanks @frelon,
Now I got : 

Building arm64 disk
mkdir -p /home/pi/Developement/elemental/elemental-toolkit/build
docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v /home/pi/Developement/elemental/elemental-toolkit/examples:/examples -v /home/pi/Developement/elemental/elemental-toolkit/build:/build \
        --entrypoint /usr/bin/elemental \
[...]

PR #2084, opened 2024-05-17
Allow to compile for RPi

bobmorane83:
Refer #2083


Makefile
```diff
@@ -4,7 +4,7 @@ export ROOT_DIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 DISK?=$(shell ls $(ROOT_DIR)/build/*.qcow2 2> /dev/null)
 DISKSIZE?=20G
 ISO?=$(shell ls $(ROOT_DIR)/build/*.iso 2> /dev/null)
-FLAVOR?=green
+FLAVOR?=green-rpi
```
frelon, line 7: I would keep the default flavor as green, then it can be overridden when building for rpi.

```suggestion
FLAVOR?=green
```

PR #2086, opened 2024-05-21
Fix recovery test

frelon:
During recovery test when writing the upgraded install state yaml file snapper fails trying to list snapshots because of trying to relabel a .snapshots dir that is read-only.

This error does not impact functionality and should be discarded.


davidcassany (approved):
LGTM, even I am not sure we are still having some issue other hidden issue in there. IIRC running in recovery mode and mounting snapshots volume somewhere else than  `/.snapshots` causes SELinux to label erratically. To me it feels that everything related to snapper is not well prepared from the imaging point of view.

In any case, nice fix!

PR #2089, opened 2024-05-22
efivar bug

frelon:
- **Do not return error for efi.ReadLoadOption**
- **Refactor and test efi manager**

### DIFF
--- a/pkg/bootloader/grub.go
+++ b/pkg/bootloader/grub.go
@@ -339,7 +339,7 @@ func (g *Grub) clearEntry() error {
 // createBootEntry will create an entry in the efi vars for our shim and set it to boot first in the bootorder
 func (g *Grub) CreateEntry(shimName string, relativeTo string, efiVariables eleefi.Variables) error {
 	g.logger.Debugf("Creating boot entry for elemental pointing to shim %s/%s", constants.EntryEFIPath, shimName)
-	bm, err := eleefi.NewBootManagerForVariables(efiVariables)
+	bm, err := eleefi.NewBootManagerForVariables(g.logger, efiVariables)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


pkg/bootloader/grub.go
```diff
@@ -339,7 +339,7 @@ func (g *Grub) clearEntry() error {
 // createBootEntry will create an entry in the efi vars for our shim and set it to boot first in the bootorder
 func (g *Grub) CreateEntry(shimName string, relativeTo string, efiVariables eleefi.Variables) error {
 	g.logger.Debugf("Creating boot entry for elemental pointing to shim %s/%s", constants.EntryEFIPath, shimName)
-	bm, err := eleefi.NewBootManagerForVariables(efiVariables)
+	bm, err := eleefi.NewBootManagerForVariables(g.logger, efiVariables)
```
davidcassany, line 342: Nice, always good having more accurate logs

pkg/efi/manager.go
```diff
@@ -57,9 +59,10 @@ func NewBootManagerForVariables(efivars Variables) (BootManager, error) {
 		if err != nil {
 			return BootManager{}, fmt.Errorf("cannot read %s: %v", name, err)
 		}
-		entry.LoadOption, err = efi.ReadLoadOption(bytes.NewReader(entry.Data))
+		entry.LoadOption, err = bm.efivars.ReadLoadOption(bytes.NewReader(entry.Data))
 		if err != nil {
-			return bm, err
+			logger.Warnf("Error reading efi load option for '%s': %s", name, err.Error())
+			continue
```
davidcassany, line 65: I guess this is the actual fix. Where do we see this issue? Which is the impact of this bug. I am wondering if this is something to be backported somewhere.

davidcassany (approved):
Looks good to me,
Just wondering about which use case is this actually fixing.

pkg/efi/manager.go
frelon: Correct, this is from the linked issue, which is coming from [this harvester issue](https://github.com/harvester/harvester/issues/4528)

PR #2093, opened 2024-05-27
[Backport v1.1.x] Do not return error for efi.ReadLoadOption

frelon:
When encountering a boot-entry for a device-path that go-efilib is
unable to parse we should just skip that entry instead of erroring out.

Backporting this manually and minimally since cherry-picking 2e56a099b
had conflicts.


davidcassany (approved):
LGTM

PR #2098, opened 2024-06-10
Ensure subvolume snapshots is mounted to list snapshots

davidcassany:
This is fixes an issue about how we were updating the `state.yaml` file after upgrades. For the corner case of running upgrades from the recovery system the  `<state_partition>/@/.snapshots` is unmounted before upgrading the `state.yaml` and this mount is required to properly list available snapshots and reflect that in `state.yaml`.

### DIFF
--- a/pkg/snapshotter/btrfs.go
+++ b/pkg/snapshotter/btrfs.go
@@ -68,6 +68,7 @@ type Btrfs struct {
 	installing        bool
 	snapperArgs       []string
 	snapshotsUmount   func() error
+	snapshotsMount    func() error
 }
 
 type btrfsSubvol struct {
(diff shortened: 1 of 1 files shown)


kkaempf (commented):
I see you provide a `snapshotsMount` implementation but where's the corresponding `snapshotsUmount` one ?

davidcassany:
Yes, in fact there used to be only a snapshotsUmount one, but no snapshotsMount. It was this way because there was the assumption it was only mounted once and at the very same stage for all cases. This turns not to be accurate enough when upgrading the snapshot from the recovery system (likely to be a squashfs image on a different partition).

pkg/snapshotter/btrfs.go
```diff
@@ -68,6 +68,7 @@ type Btrfs struct {
 	installing        bool
 	snapperArgs       []string
 	snapshotsUmount   func() error
```
davidcassany, line 70: Here is the umount function, which was already coded and used before this PR.

fgiudici (approved):
LGTM

PR #2099, opened 2024-06-11
Bump mount utils

davidcassany:


kkaempf:
Huh ? This seems to bump much more than just the mount utils ?!

pkg/types/mounter.go
```diff
@@ -28,7 +28,7 @@ type Mounter interface {
 }
 
 func NewMounter(binary string) Mounter {
-	return mount.New(binary)
+	return mount.NewWithoutSystemd(binary)
```
davidcassany, line 31: This is a new method provided with the updated interface which provides a mount wrapper that does not pretend to detect systemd. For instance this has impact on `make test-cli` target being executed in a desktop env. In that case every time a new mounter got initiated a systemd dialog appeared to provide the root password interactively, even it we are not mounting anything in our unit tests and they all can run as a non root user.


```
func mount.NewWithoutSystemd(mounterPath string) mount.Interface

NewWithoutSystemd returns a Linux specific mount.Interface for the current system. 
It provides options to override the default mounter behavior. mounterPath allows
using an alternative to `/bin/mount` for mounting. Any detection for systemd functionality
is disabled with this Mounter.

```

A part from that we were already a bit outdated in that library.

go.mod
```diff
@@ -25,7 +27,7 @@ require (
 	github.com/spf13/viper v1.15.0
 	golang.org/x/crypto v0.21.0
 	gopkg.in/yaml.v3 v3.0.1
-	k8s.io/mount-utils v0.23.0
+	k8s.io/mount-utils v0.30.1
```
davidcassany, line 30: This is the actual bump, all other changes in go.mod and in vendor as caused the indirect dependencies and the mount-utils code itself.

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/rancher/elemental-toolkit/v2
 
-go 1.22
+go 1.22.0
```
fgiudici, line 3: do we need the .0 part? If it is for the CodeCov thing, that is solved now 🎉 
... otherwise just ignore this 😅 
```suggestion
go 1.22
```

go.mod
fgiudici: golang vendoring... 😅

fgiudici (approved):
LGTM

go.mod
davidcassany: I did not add manually this change (must a consequence of `go mod venor` or `go mod tidy`). I any case makes sense keeping the version without the patch level, thanks.

PR #2100, opened 2024-06-11
Use fake mounter in tests

davidcassany:
This PR is just to use the fakemounter interface in some units tests where it was not initiated.

This prevent many dialogs asking for root password to appear when running tests in a desktop env do to the systemd integration of the default mounter.

### DIFF
--- a/pkg/action/init_test.go
+++ b/pkg/action/init_test.go
@@ -46,8 +46,10 @@ var _ = Describe("Init Action", func() {
 	var spec *types.InitSpec
 	var enabledUnits []string
 	var errCmd, initrdFile string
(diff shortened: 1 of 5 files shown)


pkg/action/init_test.go
```diff
@@ -56,6 +58,7 @@ var _ = Describe("Init Action", func() {
 		cfg = config.NewRunConfig(
 			config.WithFs(fs),
 			config.WithRunner(runner),
+			config.WithMounter(mounter),
```
davidcassany, line 61: If not defined any specific mounter by default uses the the real mounter

anmazzotti (approved):
LGTM

PR #2102, opened 2024-06-12
[CI] Increase time between pull retries

davidcassany:
There seams to be an issue with the nightly builds, looks like a sync issue between the build-toolkit and build-os jobs. Just seen few times in [nightly workflow](https://github.com/rancher/elemental-toolkit/actions/runs/9458768312/job/26054804761) which uses the same exact logic as the PR workflow.

Increasing the time between retries will not harm and hopefully give it a chance to ghcr.io to sync. With the change retries will span up to 2min before giving up.


davidcassany:
Closing the PR as this sync issue has not happened anymore since few days, it doesn't look like this is a problem anymore.

PR #2105, opened 2024-06-17
Make EFI partition size configurable at install time

### LINKED ISSUE #2104: COS_GRUB too small on ARM / RPi
Creating the boot partition seems to be based on the assumption "boot is always EFI".

So we see [EfiSize = MinPartSize](https://github.com/rancher/elemental-toolkit/blob/28c4e4735e5954f6c1a91b828ce3bc1c01e040f0/pkg/constants/constants.go#L47C2-L47C34) and `COS_GRUB` is created [with this minimal size](https://github.com/rancher/elemental-toolkit/blob/28c4e4735e5954f6c1a91b828ce3bc1c01e040f0/pkg/types/config.go#L525C4-L525C38).

However, this is wrong for ARM, esp. Raspberry Pi where everything (bootloader, firmware, kernel, initrd) must be on a `vfat` partition. The minimal size should be 256M for this `COS_GRUB` partitions.

The simplest change is probably to set `MinSize` to 256.

A better change would not use `EfiSize` for ARM since not all ARM systems boot via EFI. We probably need `BootSize` here 🤔

davidcassany:
Fixes #2104

### DIFF
--- a/cmd/config/fixtures/config/config.yaml
+++ b/cmd/config/fixtures/config/config.yaml
@@ -4,6 +4,9 @@ cloud-init-paths:
 - "some/alternate/path"
 
 install:
+  partitions:
+    efi:
+      size: 512
   target: "someDisk"
 
   no-format: true
--- a/cmd/config/fixtures/config/config.yaml
+++ b/cmd/config/fixtures/config/config.yaml
@@ -5,7 +5,7 @@ cloud-init-paths:
 
 install:
   partitions:
-    efi:
+    bootloader:
       size: 512
   target: "someDisk"
 
--- a/config.yaml.example
+++ b/config.yaml.example
@@ -19,7 +19,7 @@ install:
   #     size: 300
 
   # default partitions
-  # only 'efi', 'oem', 'recovery', 'state' and 'persistent' objects allowed
+  # only 'bootloader', 'oem', 'recovery', 'state' and 'persistent' objects allowed
   # size in MiB
   partitions:
     oem:
--- a/config.yaml.example
+++ b/config.yaml.example
@@ -19,7 +19,7 @@ install:
   #     size: 300
 
   # default partitions
-  # only 'oem', 'recovery', 'state' and 'persistent' objects allowed
+  # only 'efi', 'oem', 'recovery', 'state' and 'persistent' objects allowed
   # size in MiB
   partitions:
     oem:
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -23,8 +23,8 @@ import (
 
 const (
 	BiosPartName       = "bios"
-	EfiLabel           = "COS_GRUB"
-	EfiPartName        = "efi"
+	BootLabel          = "COS_GRUB"
+	BootPartName       = "efi"
 	SystemLabel        = "COS_SYSTEM"
 	RecoveryLabel      = "COS_RECOVERY"
 	RecoveryPartName   = "recovery"
@@ -40,11 +40,11 @@ const (
 	LinuxFs            = "ext4"
 	LinuxImgFs         = "ext2"
 	SquashFs           = "squashfs"
-	EfiFs              = "vfat"
+	BootFs             = "vfat"
 	Btrfs              = "btrfs"
 	BiosFs             = ""
 	MinPartSize        = uint(64)
-	EfiSize            = MinPartSize
+	BootSize           = MinPartSize
 	OEMSize            = MinPartSize
 	StateSize          = uint(8192)
 	RecoverySize       = uint(4096)
@@ -98,7 +98,7 @@ const (
 	OEMDir             = "/run/elemental/oem"
 	PersistentDir      = "/run/elemental/persistent"
 	TransitionDir      = "/run/elemental/transition"
-	EfiDir             = "/run/elemental/efi"
+	BootDir            = "/run/elemental/efi"
 	ImgSrcDir          = "/run/elemental/imgsrc"
 	WorkingImgDir      = "/run/elemental/workingtree"
 	OverlayDir         = "/run/elemental/overlay"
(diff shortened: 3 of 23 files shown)


davidcassany:
In addition to that we could also consider having way higher default value (what about 1G or even 2G?) So if we ever support systemd-boot or unified kernel image (which require having kernel and initrd within the EFI partition) we could eventually consider upgrades without having to reinstall from scratch.

config.yaml.example
```diff
@@ -19,7 +19,7 @@ install:
   #     size: 300
 
   # default partitions
-  # only 'oem', 'recovery', 'state' and 'persistent' objects allowed
+  # only 'efi', 'oem', 'recovery', 'state' and 'persistent' objects allowed
```
anmazzotti, line 22: Just nitpicking, but it would be best to also include an efi partition example, just to highlight the fact that the filesystem to use should be `vfat`. Or can it be omitted even? I don't truly know if we use the defaults or not.

anmazzotti (approved):
Looks good, only one minor remark to provide an example.

config.yaml.example
davidcassany: vfat filesystem is set by default and I was wondering to force it or not (error out if not set to vfat), but then I thought about better giving some flexibility as this is pretty hidden in the setup and if one is actually touching this is probably for a good reason. For instance I don't know if for RPi it is mandatory to boot from vfat or it can be another filesystem.
What we probably could consider in the future is renaming this EFI partition to some firmware neutral name (boot partition?), so there is no explicit assumption we are using EFI firmware. But I did not want to touch that at this time because it feels like a risky change, there might be some corner cases hard to foresee.

anmazzotti: I would not force it as well, but it could be an idea. What I missed is an example for the end-users to follow, let's say I do want to use this feature, then I would miss (without looking at the source code) how can I extend the partition size, and what do I need to configure in order to do that.
I was wondering is that `fat`, `fat32`, etc. before looking at the source, and if it's optional is even better, so I could just code:

```yaml
efi:
  size: 1000
```

davidcassany: Indeed this is the case you can provide only size and this will only modify the default size, but keep all other parameters. About partition sizes we definitely need some docs on that topic and we already have an issue for that https://github.com/rancher/elemental-docs/issues/345

kkaempf:
As said in #2104, calling it `EFI` partition is misleading on Raspberry (and might as well other ARM platforms).
Raspberry doesn't boot from EFI but also needs a `vfat` formatted boot partition.

It might not matter implementation wise, but I'm somewhat concerned about mixing topics.

davidcassany:
> As said in #2104, calling it `EFI` partition is misleading on Raspberry (and might as well other ARM platforms). Raspberry doesn't boot from EFI but also needs a `vfat` formatted boot partition.
> 
> It might not matter implementation wise, but I'm somewhat concerned about mixing topics.

All right, this is a fair point and probably we can start by exposing it with a different name but keep all variables and constants as they are now, which is probably already misleading. Lets start having a look.

davidcassany:
I also dislike the `boot` name as this can easily be confusing with `/boot` 🤔 Shall we call it firmware partition? Something like:

```yaml
install:
  partitions:
    firmware:
      size: 512
    oem:
      ...
```

We could easily name EFI to `firmware` in UX and probably change the `EFI` variable name to `FW` or something like that, however there still many constant literals that refer to `efi` (e.g. the partition is mounted at `/run/elemental/efi`).

Note that in toolkit we are only effectively supporting EFI firmware. So to some extend it makes sense keeping the default constant literals referring to `efi` (e.g. `/run/elemental/efi`).

Shall we try to do such a rename? How far should we attempt to go.

kkaempf:
`firmware` isn't clear either, as there's "boot firmware" (like in RPi) and "kernel firmware" (all the *-firmware rpms).

Yeah, naming is hard 😆

davidcassany:
Probably `bootloader` partition is more accurate? Could be meaningful for `BIOS`, `EFI` and UBOOT-like with EFI Grub

kkaempf:
> Probably `bootloader` partition is more accurate? Could be meaningful for `BIOS`, `EFI` and UBOOT-like with EFI Grub

Let's go for `bootloader` !

cmd/config/fixtures/config/config.yaml
```diff
@@ -4,6 +4,9 @@ cloud-init-paths:
 - "some/alternate/path"
 
 install:
+  partitions:
+    efi:
+      size: 512
```
kkaempf, line 9: How/where do we document the size unit ? (I'd assume it's in megabytes ;-) )

config.yaml.example
```diff
@@ -19,7 +19,7 @@ install:
   #     size: 300
 
   # default partitions
-  # only 'oem', 'recovery', 'state' and 'persistent' objects allowed
+  # only 'bootloader', 'oem', 'recovery', 'state' and 'persistent' objects allowed
   # size in MiB
```
davidcassany, line 23: Here is were we document all sizes are in megabytes. It is also documented in elemental-toolkit docs.

cmd/config/fixtures/config/config.yaml
davidcassany: Yes everything is in megabytes, it is already added as a comment in the example config file and also in elemental-toolkit docs.

pkg/constants/constants.go
```diff
@@ -23,8 +23,8 @@ import (
 
 const (
 	BiosPartName       = "bios"
-	EfiLabel           = "COS_GRUB"
-	EfiPartName        = "efi"
+	BootLabel          = "COS_GRUB"
+	BootPartName       = "efi"
```
davidcassany, line 27: Note this PR changes the variable and constant names, but none of the values. This to ensure no regressions are introduced. If we consider changes values to I'd do it in a separate and follow up PR.

pkg/types/config.go
```diff
@@ -502,7 +502,7 @@ func (pl PartitionList) GetByNameOrLabel(name, label string) *Partition {
 
 type ElementalPartitions struct {
 	BIOS       *Partition
-	EFI        *Partition
+	Boot       *Partition `yaml:"bootloader,omitempty" mapstructure:"bootloader"`
```
davidcassany, line 505: `bootloader` partition is named as `Boot` in the struct, just for the sake of short names.

kkaempf (approved):
Ok to merge 🚀

kkaempf:
I *cannot* add
~~~
partitions:
  bootloader:
    size: 128
~~~
via the Rancher UI 😕

kkaempf:
When trying to edit via `kubectl` I get
~~~
> kubectl edit -n fleet-default MachineRegistration rpi-cluster-nodes
error: machineregistrations.elemental.cattle.io "rpi-cluster-nodes" is invalid
A copy of your changes has been stored to "/tmp/kubectl-edit-rw5pf.yaml"
error: Edit cancelled, no valid changes were saved.
~~~

kkaempf:
Confirmed working, thanks !

PR #2117, opened 2024-06-26
Use nightly tag for ghcr.io nightly builds

anmazzotti:
> This should always replace the `latest` elemental-cli image when building from main.
> 
> This change is desirable so users can reliably pull the `latest` tag and not have to follow the commit tags that on the other hand seem excessive.
> 
> Instead of building an image per commit, we can think about re-instating GitHub releases and build images based on git tags.

Scratch the above, since these commit tags are needed for the PR tests on OBS.
Next best thing is to add a `nightly` tag.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -26,7 +26,7 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Build toolkit
         run: |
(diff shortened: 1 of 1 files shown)


davidcassany (approved):
Should be fine just for the nightly.

I am wondering though if this has benefits though, as I guess this will always push a new image to the registry anyways. Probably we could just retag the already existing version, which should have been build previously on main push.

Issue #2125, opened 2024-07-01
Toolkit built images don't consume cloud-init files named "meta-data" from the cd-rom provider


robertwbl:
**Is your feature request related to a problem? Please describe.**
Harvester puts both user-data & meta-data files containing cloud-init config into the cloud-init drive that gets attached to VMIs ([line 587-588 of cloud-init.go](https://github.com/harvester/harvester/blob/226f72aa58532071232ad548e27970c1cec55733/vendor/kubevirt.io/kubevirt/pkg/cloud-init/cloud-init.go#L587)). user-data contains the cloud config that users specify when creating VMs, and meta-data contains instance data generated by Harvester (this is invisible to the user) such as the instance ID and machine name, which is exactly what I’m looking for in this use case. In my testing this meta-data cloud-init “data source” gets consumed correctly by other OSes such as the Tumbleweed cloud image, setting it’s hostname to whatever the VM name is in Harvester, so I thought I could simply point an Elemental derived OS at this config and be fine, except looking at the [cd-rom provider in the elemental-toolkit](https://github.com/rancher/elemental-toolkit/blob/d5eab7e7f8e088dda82eb225612f1e45018fd939/vendor/github.com/rancher-sandbox/linuxkit/providers/provider_cdrom.go#L31) this is hard coded to only look at files called “user-data” or “config”. I can therefore point to the cloud-init drive attached to the VM, but not the specific file within.

**Describe the solution you'd like**
Would there be appetite for adding “meta-data” to this list of strings given this seems to be a standard file name for cloud-init systems, or a more generic solution such as providing a mechanism for also specifying a path with the cd-rom provider?

**Describe alternatives you've considered**
Manually setting the hostname of Elemental VMs

**Additional context**
Slack thread: https://rancher-users.slack.com/archives/C028DVCAYLD/p1719215874515729

PR #2126, opened 2024-07-01
Add "meta-data" to userDataFiles string list Refers #2125

robertwbl:
As suggested in the issue I've raised, this allows cloud-init on toolkit built images to pick up cloud config in files named "meta-data" such as the one Harvester provides instance data in.

### DIFF
--- a/vendor/github.com/rancher-sandbox/linuxkit/providers/provider_cdrom.go
+++ b/vendor/github.com/rancher-sandbox/linuxkit/providers/provider_cdrom.go
@@ -28,7 +28,7 @@ import (
 // ListCDROMs lists all the cdroms in the system
 func ListCDROMs() []Provider {
 	// UserdataFiles is where to find the user data
-	var userdataFiles = []string{"user-data", "config"}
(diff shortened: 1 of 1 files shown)


robertwbl:
It has occurred to me that perhaps this change is better made upstream here: https://github.com/rancher-sandbox/linuxkit/blob/main/providers/provider_cdrom.go ?

davidcassany (changes requested):
> It has occurred to me that perhaps this change is better made upstream here: https://github.com/rancher-sandbox/linuxkit/blob/main/providers/provider_cdrom.go ?

Indeed, this vendored folder it autogenerated. Hence no changes there should be done there. We'd appreciate if you make the equivalent PR in https://github.com/rancher-sandbox/linuxkit repository and then we will take it from there, make a new release of the library, update the yip dependency and finally update the elemental-toolkit itself.

Does it sound good?

robertwbl:
Thanks @davidcassany that sounds good. I've opened https://github.com/rancher-sandbox/linuxkit/pull/5 and am closing this PR.

PR #2127, opened 2024-07-01
Switch to Leap Micro instead of tumbleweed

davidcassany:


davidcassany:
Leap Micro 6.0 as is it is not usable for building custom images. The default repository only includes a really limited set of packages, missing tools such as NetworkManager and many others.

PR #2129, opened 2024-07-04
Set token to explicit content read permissions

davidcassany:


.github/workflows/build.yaml
```diff
@@ -21,6 +21,8 @@ on:
 
 jobs:
   detect:
+    permissions:
+      content: read
```
pjbgf, line 25: nit: It is also possible to set this at the highest level, and then just override the permissions on jobs that require than `content: read`.

pjbgf (approved):

pjbgf (approved):

PR #2131, opened 2024-07-09
Allow insecure registries

davidcassany:
This PR allows to pull images from insecure registries with the flag `--tls-verify`. 

Part of #2132

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -81,7 +81,7 @@ func addLocalImageFlag(cmd *cobra.Command) {
 
 // addVerifyRegistryFlag add local image flag shared between install, pull-image, upgrade
 func addTLSVerifyFlag(cmd *cobra.Command) {
-	cmd.Flags().Bool("tls-verify", false, "Require HTTPS and verify certificates of registries (default: true)")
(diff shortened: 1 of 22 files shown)


cmd/flags.go
```diff
@@ -78,6 +79,11 @@ func addLocalImageFlag(cmd *cobra.Command) {
 	cmd.Flags().Bool("local", false, "Use an image from local cache")
 }
 
+// addVerifyRegistryFlag add local image flag shared between install, pull-image, upgrade
+func addTLSVerifyFlag(cmd *cobra.Command) {
+	cmd.Flags().Bool("tls-verify", false, "Require HTTPS and verify certificates of registries (default: true)")
```
anmazzotti, line 84: Flag is false by default, not sure if the description is wrong or the actual default, I like it to be on by default or at least that's what I would expect as user.

anmazzotti (changes requested):
Actual default should be true I believe. 
Also I don't fully get how this is supported now, I recall we didn't do verification before, is this a new feature of `github.com/google/go-containerregistry`?

cmd/flags.go
davidcassany: Done

anmazzotti (approved):
Awesome feature!
Looks good.

PR #2134, opened 2024-07-09
Fix workflows, permissions at top level

davidcassany:
This is fixing https://github.com/rancher/elemental-toolkit/pull/2129 which introduced a syntax error, permissions should be at the very top level. See [error](https://github.com/rancher/elemental-toolkit/actions/runs/9843930952)

In addition it is also adding permissions to read PRs, to my understanding this is required to allow GHA runs on PRs from forks.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -19,6 +19,10 @@ on:
       - go.sum
       - examples/**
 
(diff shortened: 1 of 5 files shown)


anmazzotti (approved):
LGTM

PR #2137, opened 2024-07-09
Refactor ci

### LINKED ISSUE #2132: Refactor Toolkit CI to use a private local registry
We should avoid pushing packages on to ghcr.io on PRs, this basically implies the upgrade test and passing container images across jobs. Using the cache seams to be a reasonable approach as it performs well and it is not over complex. 

Providing a private registry to test the upgrade is the actual challenge.

davidcassany:
Fixes #2132

### DIFF
--- a/.github/workflows/build_and_test_x86.yaml
+++ b/.github/workflows/build_and_test_x86.yaml
@@ -39,19 +39,14 @@ jobs:
           path: /tmp/os.tar
           key: ${{ env.cache-name }}-${{ inputs.version }}
       - if: ${{ steps.cache-os.outputs.cache-hit != 'true' }}
-        name: Fetch toolkit image
-        id: cache-toolkit
-        uses: actions/cache/restore@v4
-        env:
-          cache-name: pr-toolkit-build-x86_64
+        name: Pull images
+        uses: ./.github/actions/fetchimages
         with:
-          path: /tmp/toolkit.tar
-          key: ${{ env.cache-name }}-${{ inputs.version }}
-          fail-on-cache-miss: true
+          os: 'false'
+          version: ${{ inputs.version }}
       - if: ${{ steps.cache-os.outputs.cache-hit != 'true' }}
         name: Build OS
         run: |
-          docker load -i /tmp/toolkit.tar
           make ARCH=${{ env.ARCH }} build-os-save
           mv build/elemental-${{ inputs.flavor }}*.tar /tmp/os.tar 
       - if: ${{ steps.cache-os.outputs.cache-hit != 'true' }}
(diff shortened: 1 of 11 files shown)


scripts/run_registry.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
```
davidcassany, line 1: This is a script that handles the local registry instance with a variant of `docker run registry:2`. Challenges here were:

* Need to compute a valid IP of the host that is visible from the qemu VM in "user mode" network setup.
* Need to push to an insecure registry (I could not figure out how work around that, a self signed certificate would still be insecure unless the host is configured to trust it)

scripts/run_registry.sh
```diff
@@ -0,0 +1,78 @@
...
+    echo "Pushing '${img}' to '${reg_img}'"
+
+    # Ugly hack around podman to circumvent the need of adding insecure registries
+    # at /etc/docker/daemon.json when using the docker client 
+    docker save "${img}" | podman load
+    tag=${img##*:}
+    imgID=$(podman images -n | grep "${tag}" | awk '{print $3}' | head -n1)
+    podman tag "${imgID}" "${reg_img}"
+    podman push --tls-verify=false "${reg_img}"
+    podman rmi -f "${imgID}"
```
davidcassany, line 48: The ugliness of all this is two fold:

* We are using `podman push` because docker does not have the equivalent of `--tls-verify=false` and the daemon needs to be configured for that. I want `make test-upgrade` to be executable locally without having to mess with docker daemon configuration.
* GH Runners use podman v3.4.4 (current podman is at v5.1) and this version does not properly keep image reference on `podman load`, so that there is this `podman images` to find the actual image ID, because we can't relay on the image being loaded with the former reference 😞

scripts/run_upgradetest.sh
```diff
@@ -0,0 +1,49 @@
...
+
+  export VM_PID=$(${SCRIPTS_PATH}/run_vm.sh vmpid)
+  export COS_PASS=${ELMNTL_OLDPASS}
+
+  reg_url=$(${SCRIPTS_PATH}/run_registry.sh url)
+
+  pushd "${ROOT_PATH}" > /dev/null
+    go run ${ginkgo} ${ginkgo_args} ./tests/wait-active
+    go run ${ginkgo} ${ginkgo_args} ./tests/upgrade -- \
+      --toolkit-image=docker://${reg_url}/${toolkit_img} --upgrade-image=docker://${reg_url}/${upgrade_img}
```
davidcassany, line 30: This scripts is essentially existing because it was so cumbersome to compose this arguments from variables in the Makefile directly

davidcassany (commented):
This PR allows to run the upgrade test locally too. In fact, that's the thing, if it is runnable locally without having to configure infraestructure, it should also be possible to run it in GHA without needing external services like a registry.

In my env the following sequences passes:

```bash
make build && make build-os
make test-upgrade
```

.github/workflows/build_and_test_x86.yaml
```diff
@@ -14,31 +17,47 @@ concurrency:
 jobs:
   build-os:
     permissions:
-      packages: write
```
davidcassany, line 17: No more `packages: write` in this workflow. Only `contents: read`.

.github/workflows/cache-cleanup.yaml
```diff
@@ -5,6 +5,7 @@ name: Cleanup caches by a branch
 on:
   pull_request_target:
     types:
+      - synchronize
```
davidcassany, line 8: To remove unused cache on pushing commits to an opened PR

scripts/run_registry.sh
anmazzotti: I think GH Actions still uses `ubuntu-22.04` by default, but it should be safe to upgrade to `ubuntu-24.04`instead, even if it's beta. That should contain podman 4.9.3

scripts/run_registry.sh
```diff
@@ -0,0 +1,78 @@
...
+
+set -e
+
+SCRIPT=$(realpath -s "${0}")
+SCRIPTS_PATH=$(dirname "${SCRIPT}")
+TESTS_PATH=$(realpath -s "${SCRIPTS_PATH}/../tests")
+
+: "${ELMNTL_PREFIX:=}"
+: "${ELMNTL_REGCONF:=${TESTS_PATH}/${ELMNTL_PREFIX}test-registry.yaml}"
+: "${ELMNTL_IPADDR=$(ip -4 addr show eth0 | grep -oP '(?<=inet\s)\d+(\.\d+){3}')}"
```
anmazzotti, line 11: In a similar occasion I used `$(ip addr show $(ip route | awk '/default/ { print $5 }') | grep "inet\b" | awk '{print $2}' | cut -d/ -f1)` so that you don't have to guess the interface name. Not sure if it's more reliable, but that still works for me.

anmazzotti (approved):
Great work. Left a couple of nitpicking comments, but nothing essential.
LGTM

scripts/run_registry.sh
davidcassany: Gonna update it with your suggestion ;)

scripts/run_registry.sh
davidcassany: I had to roolback to 22.04 as KVM support fails in 24.04, I think it would be better to wait until manage to run KVM on 24.04, out of the box did not work.

PR #2144, opened 2024-07-15
Fix cache cleaning for PRs and main pushes

davidcassany:
Follow up of #2137, this PR makes sure the cache is cleaned on PR close/synchronize and after pushing to `main` branch.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -5,6 +5,10 @@ on:
     branches:
       - main
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }}
+  cancel-in-progress: true
+
 jobs:
   detect:
     permissions:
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yaml
```diff
@@ -43,17 +47,12 @@ jobs:
           git fetch --prune --unshallow
       - name: Define version
         id: version
-        env:
-          hash: ${{ hashFiles('Dockerfile', '**/go.sum', '**/pkg/**', '**/examples/**', '**/cmd/**', '**/vendor/**', '**/Makefile', '**/main.go') }}
-        run: |
-          version="${{ env.hash }}"
-          version=${version::16}
-          echo "version=${version}" >> $GITHUB_OUTPUT
+        uses: ./.github/actions/version
```
davidcassany, line 50: Moved to an action file as this is used twice and the paths used for the need to be exactly the same. So to avoid misalignment and duplication this logic has been moved to `.github/actions/version`

.github/workflows/build.yaml
```diff
@@ -43,17 +47,12 @@ jobs:
...
       - name: Check cache for Toolkit image
         id: cache-toolkit
         uses: actions/cache/restore@v4
         env:
-          cache-name: pr-toolkit-build-x86_64
+          cache-name: toolkit-build-x86_64-${{ github.event_name }}
```
davidcassany, line 55: This is used for PRs and for pushed to `main`, so to distinguish them we use the `github.event_name`, this way we ensure that once rebuilding in `main` after merging it does not reuse the cache from the PR that caused the push.

.github/workflows/cache-cleanup.yaml
```diff
@@ -8,47 +8,60 @@ on:
...
+    steps:
       - name: Cleanup
         run: |
           gh extension install actions/gh-actions-cache
           
           echo "Fetching list of cache key"
-          cacheKeysForPR=$(gh actions-cache list -R $REPO -B $BRANCH -L 100 | cut -f 1 )
-
+          cacheKeysForPR=$(gh actions-cache list -R "${REPO}" -B "${BRANCH}" -L 100 | cut -f 1 )
+
```
fgiudici, line 50: nit: extra tab?
```suggestion

```

fgiudici (approved):
looks good here!

PR #2146, opened 2024-07-18
[backport] CI refactor to v2.1.x branch

davidcassany:
This allows to run PRs to v2.1.x branch following the same upstream workflow setup, using a local registry and a github cache

This backport required several additional related changes that are not strictly exclusive to the CI workflow refactor. I added all them together in this PR as this way I can verify tests are passing green. Here is an explanatory list of whats included:

* #2061 (This is required to enable adding authorized SSH keys in ISOs, bugfix)
* 8876278f9f4b8396b4e6ecb644701dd920f56f95 (gets rid of the old `cos` password)
* bf6a01ad129f3765d84e80092ae33bdde809ae92 (allows password less connections in tests using an SSH key pair)
* 76159d9f34bf907d26fddf2de9533f2709d17184 (keeps using legacy `cos` passwords for upgrade test which uses and old image to upgrade from)
* #2131 (required to allow CI tests to pull images from an local insecure registry without TLS)

All other four commits are the backport of the CI refactor itself.


Dockerfile
```diff
@@ -11,13 +11,17 @@ WORKDIR /src/
 ADD go.mod .
 ADD go.sum .
 ADD vendor vendor
-RUN go mod download
+
+# Do not download neither verify vendored folder, yip should be bumped to a recent version as
+# some of its dependencies are no longer available upstream. Relay entirely on vendored folder
+# RUN go mod download
+
```
davidcassany, line 18: This is to relay entirely to the vendored folder, which I think is fine for a maintenance branch. Other wise we will see failure in CI as it happened in this [run](https://github.com/rancher/elemental-toolkit/actions/runs/9989529687?pr=2145)

davidcassany:
This PR will show some failures, this is because the old worlflow was configured to run the workflow of the target branch while the new one runs the workflow defined in the branch itself. As a result this PR runs two workflows:

* one based on the [target branch](https://github.com/rancher/elemental-toolkit/actions/runs/9994166714) which will fail because it uses the old workflow (trying to make use of the ghcr.io registry to share images across jobs and VMs) against the current Makefile from this PR
* one based entirely on the [current PR](https://github.com/rancher/elemental-toolkit/actions/runs/9994167344) which must pass as this uses the GHA cache and a local registry to share containers with launched VMs.

cmd/cloud-init.go
```diff
@@ -62,6 +62,10 @@ func NewCloudInitCmd(root *cobra.Command) *cobra.Command {
 			}
 
 			err = cfg.CloudInitRunner.Run(stage, args...)
+			if err != nil {
+				cfg.Logger.Errorf("cloud-init command failed: %v")
```
fgiudici, line 66: ups, `err` forgotten!
```suggestion
				cfg.Logger.Errorf("cloud-init command failed: %v", err)
```

fgiudici (approved):
Looks good, just found a missing `err` var!

PR #2159, opened 2024-08-07
Make RAW disks recovery partition expandable

davidcassany:
This PR makes the recovery partition expandable, so the RAW disk image created by the "build-disk" command only includes the EFI partition, OEM partition and the Recovery partition sized as the recovery image it includes. On first boot the recovery partition is resized to the configured size for recovery (4G by default) and state and persistent partitions created. Finally the deploy command after the first full boot resets from recovery to install an active system to state partition.

In order to do that this PR adds a new Yip stage at boot. Why? Because the earliest stage we had (`rootfs`) is called once the root partition (recovery in this case) is already mounted, hence too late to run any expand operation. The new stage introduced in this PR is called after OEM partition is mounted, but before root device is mounted. This allows running an expand partition operation for recovery on first boot.

As a result RAW images will no longer be around 4G but around 1G, almost like ISO's.


kkaempf (approved):
worth a try 😉

PR #2163, opened 2024-08-19
Expose ELEMENTAL_CLOUD_INIT_PATHS option

anmazzotti:
This adds the `--cloud-init-paths` argument for all `elemental` cli subcommands. This was previously added only for `build-disk`, but I find it useful in all other cases.

The `ELEMENTAL_CLOUD_INIT_PATHS` env variable is also added to achieve the same effect.

The reason to introduce this feature is to address corner cases when using `after-install` and `after-reset`, especially the latter since in the default Elemental images, `/system/oem` and `/usr/local/cloud-config` are read only, and `/oem` is going to be formatted before being evaluated.

Having a dynamic path to be evaluated at run time allows for more freedom.

This is also an enabler for https://github.com/rancher/elemental-operator/issues/806


ldevulder (approved):
Not an expert but I don't see anything that could break the toolkit, so LGTM.

PR #2171, opened 2024-09-04
Try to remove old workflows from github

frelon:
Github finds security issues in old workflows that are no longer used
and do not exist in main branch.

This commit adds back the build-* workflows to try and overwrite the
workflow runs and remove the security issues.


frelon:
The security alerts are gone, closing

PR #2173, opened 2024-09-05
Extend upgrade test to upgrade recovery too

davidcassany:
This PR adds some additional checks for the upgrade tests. It performs a wider set of actions, basically after upgrading the active system from a `v1.x` based toolkit it reboots to recovery to verify it is still on an old version and then reboots to active again to run a recovery upgrade. Finally it checks both active and recovery are functional.

With this change we essentially test that upgrading recovery from `v1.x` based images to `v2.x` is still functional.


frelon (approved):
LGTM

PR #2182, opened 2024-09-10
Implement state subcommand

### LINKED ISSUE #2172: Apply user defined labels to elemental managed snapshots (on install, upgrade, reset)

It would be nice to be able to identify successful upgrades through a user defined string. 
The simplest use case could be to match the upgrade image URI to the correlation id, so that an elemental-toolkit consumer knows an upgrade as already been applied correctly when trying to reconcile the desired OS image version.

A `correlation-id` argument can be added to `elemental upgrade`.
For example:
```
elemental upgrade --system oci://my-os-image:v1.2.3 --correlation-id my-os-image:v1.2.3
```

This correlation ID should be attached to the snapshot created by the upgrade process.
This also implies the snapshot information **must be retrieved back** by consumers.

This could be done by extending and exposing the toolkit's `InstallState`. 
First of all, it should contain the list of snapshots  and some convenient information about them:

```yaml
# The order of the snapshots is important, it should reflect the order in the bootloader. 
# Is this always equivalent to sorting by creationTimestamp?
snapshots: 
 - entryName: Elemental
   system: "oci://my-os-image:v1.2.3"
   creationTimestamp: "2024-08-30T12:51:43Z"
   correlationID: "96948aad3fcae80c08a35c9b5958cd89"
 - entryName: Elemental (Snapshot 1)
   system: "oci://my-os-image:v1.0.0"
   creationTimestamp: "2024-08-25T11:02:12Z"
```

Then a new subcommand could be implemented to display the state.

```bash
> elemental installstate get
date: "2024-08-30T12:52:03Z"
partitions:
  state: ...
  recovery: ...
  .
  .
snapshots:
 - entryName: Elemental
   system: "oci://my-os-image:v1.2.3"
   creationTimestamp: "2024-08-30T12:51:43Z"
   correlationID: "96948aad3fcae80c08a35c9b5958cd89"
 - entryName: Elemental (Snapshot 1)
   system: "oci://my-os-image:v1.0.0"
   creationTimestamp: "2024-08-25T11:02:12Z"
```

Could go as fancy as implementing full reflection, but probably not worth it:

```
> elemental installstate get snapshots 0 correlationID
96948aad3fcae80c08a35c9b5958cd89
```

Also output can be YAML only, but if needed can be extended later with a `--format` argument.

Now, as a machine agent I can: 
- Receive the instruction to upgrade to image `my-os-image:v1.2.3` **and** use it to upgrade recovery as well.
- I can hash this instruction to: `my-os-image:v1.2.3+recovery`
- Invoke `elemental installstate get` and parse the output to understand whether the **current** [0] snapshot's correlationID matches the hash. If yes, I have nothing to do. Go back idling.
- If not, an upgrade has to happen, so I invoke `elemental upgrade --system oci://my-os-image:v1.2.3 --recovery --correlation-id my-os-image:v1.2.3+recovery`
   
The current approach is to run `elemental upgrade` **from** the upgrade image itself, through a container running on the system. Then the upgrade container can check (before upgrade) if some sentinel file matches the host's equivalent, for example `diff /etc/os-release /host/etc/os-release` to determine upgrade needs to happen or not. 
This however needs to spawn a container in the first place, just to determine whether an upgrade has been applied or not.

**Additional context**
This is an enabler for the Elemental CAPI agent. This agent is offering the possibility to run `elemental upgrade` directly for upgrades (before implementing a through-container oriented alternative). Early prototyping (exploting grub vars instead of InstallState) is commented here: https://github.com/rancher-sandbox/cluster-api-provider-elemental/pull/86#issuecomment-2321190754

anmazzotti:
Closes #2172 

This simply dumps the installation state on stdout.

```
$ elemental state
date: "2024-09-09T12:33:25Z"
snapshotter:
    type: btrfs
    max-snaps: 4
    config: {}
efi:
    label: COS_GRUB
oem:
    label: COS_OEM
persistent:
    label: COS_PERSISTENT
recovery:
    label: COS_RECOVERY
    recovery:
        source: dir:///run/rootfsbase
        fs: squashfs
state:
    label: COS_STATE
    snapshots:
        1:
            source: dir:///run/rootfsbase
        2:
            source: oci://192.168.122.10:30000/elemental-os:dev-next
            digest: sha256:11cb5c6f7b6b9e4daff67ec3ec7fa4a028c24445f2e4834a78e93c75d73eb5c3
            active: true
```


docs/elemental_upgrade-recovery.md
```diff
@@ -13,7 +13,7 @@ elemental upgrade-recovery [flags]
       --poweroff                         Shutdown the system after install
       --reboot                           Reboot the system after install
       --recovery-system.uri string       Sets the recovery image source and its type (e.g. 'docker:registry.org/image:tag')
-      --snapshot-labels stringToString   Add labels to the to the system (ex. --labels my-label=foo,my-other-label=bar) (default [])
+      --snapshot-labels stringToString   Add labels to the to the system (ex. --snapshot-labels my-label=foo,my-other-label=bar) (default [])
```
fgiudici, line 16: good catch!

fgiudici (approved):
Looks good here!

PR #2183, opened 2024-09-13
Fix docker build warnings when using buildx

frelon:
The following warnings are fixed in the example flavors + cli
Dockerfile:

```
- StageNameCasing: Stage name 'TOOLKIT' should be lowercase (line 7)
- StageNameCasing: Stage name 'OS' should be lowercase (line 10)
- InvalidDefaultArgInFrom: Default value for ARG
  ${TOOLKIT_REPO}:${VERSION} results in empty or invalid base image name
  (line 7)
- JSONArgsRecommended: JSON arguments recommended for CMD to prevent
  unintended behavior related to OS signals (line 97)
- FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 4)
```


davidcassany (approved):
LGTM

Issue #2188, opened 2024-09-17
Cleanup of unused images in GHCR


frelon:
We should cleanup all unused images from GHCR and only keep the one we are actually using, which are the images matching an Elemental Toolkit release.

Status:
 - [x] cli
 - [x] green
 - [x] orange
 - [x] blue

frelon:
Cleaned up old images in all the toolkit repos

PR #2189, opened 2024-09-17
Add push-images workflow

frelon:
A starting point for a workflow that will build and push the elemental-toolkit images to ghcr.io.

Part of #2188


davidcassany (approved):
LGTM

PR #2191, opened 2024-09-18
Take ref as input in push-images workflow

frelon:
Part of #2188


davidcassany (approved):
Looks good to me, 

Just added some comments about how I think it would be possible to make the dispatch workflow and automated one consistent (I had this sort of issues in latest toolkit worklow changes).

I any case I am fine having it only manual for now and only consider automated ones after having it tested and validated for while.

PR #2192, opened 2024-09-23
 Add cloud-init paths of the new root in 'after-*' hooks

davidcassany:
This commit enables to run the non chrooted 'after-*' hooks included in the newly deployed image root. This specially applies to the install, reset, upgrade and build-disk commands.

Moreover, 'after-disk' command now includes static reference paths to the new root and working directory, so that those can be used within the hooks regardless of the chosen output directory.

Part of rancher/elemental#1516

### DIFF
--- a/pkg/features/features.go
+++ b/pkg/features/features.go
@@ -54,6 +54,7 @@ const (
 	FeatureCloudConfigEssentials = "cloud-config-essentials"
 	FeatureBootAssessment        = "boot-assessment"
 	FeatureAutologin             = "autologin"
+	FeatureArmFirmware           = "arm-firmware"
 )
 
 var (
@@ -67,6 +68,7 @@ var (
 		FeatureCloudConfigDefaults,
 		FeatureCloudConfigEssentials,
 		FeatureBootAssessment,
+		FeatureArmFirmware,
 	}
 
 	Default = []string{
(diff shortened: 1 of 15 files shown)


pkg/features/features.go
```diff
@@ -79,6 +81,7 @@ var (
 		FeatureCloudConfigDefaults,
 		FeatureCloudConfigEssentials,
 		FeatureBootAssessment,
+		FeatureArmFirmware,
```
davidcassany, line 84: I wonder if we should actually include it here or not. It does not have much impact on Elemental as we are not using the default value on our Micro images, however it has on the example images. If not included here for the RPi example we will have to list all features in the command line 🤔 

Probably would make sense having some sort of `--all` flag to force the installation of all features, assuming that default ones is a different set. Any thoughts?

frelon: I would also not include this as the default, unless we check that the target platform actually is `arm`..

A good trade-of would be as you mentioned to add the `all` value to include all features [here](https://github.com/rancher/elemental-toolkit/blob/b4765b9fc1b4048c643e1475893514e796240ed3/cmd/init.go#L55)

davidcassany: For now I did not add the `--all` flag andremoved the `arm-firmware` from the defaults list. However I thought about adding features as arguments instead of a single comma separated value. With this we can easily list them in a Dockerfile as a multi-line arguments list. This way is also possible to use the  `cobra.OnlyValidArgs` method to ensure the provided features exist. Moreover, it is easy to keep backward compatibility, hence we can actually  change this `elemental init` command UX without actually breaking any existing Dockerfile.

frelon: Yep, that looks good! Nice job!

PR #2193, opened 2024-09-24
Expose max snapshots environment variable

anmazzotti:
Little change to expose the MAX_SNAPS as environment variable. 
This will be exploited by the `elemental-register` during install.


davidcassany (approved):
LGTM

PR #2194, opened 2024-09-25
green: install xterm-resize

frelon:
Useful tool when running the tests in serial mode (ELMNTL_DEBUG=yes).


davidcassany (approved):
LGTM

PR #2195, opened 2024-09-30
Bump yip

frelon:
Yip v1.9.5 improves the handling of errors in yaml configs.

This means we can start passing `--strict` flag to `elemental-setup-*` services to report any errors as service failures.


frelon:
Needs rancher/yip#131 in order to not show an error for `elemental-setup-rootfs` service

frelon:
Reopening with only the version-bump

PR #2199, opened 2024-10-02
Add systemd-timesync user

frelon:
This user is needed for systemd-timesyncd service to start without
errors.

### DIFF
--- a/examples/green/Dockerfile
+++ b/examples/green/Dockerfile
@@ -75,6 +75,8 @@ COPY --from=toolkit /usr/bin/elemental /usr/bin/elemental
 RUN systemctl enable NetworkManager.service && \
     systemctl enable sshd.service
 
+RUN useradd --system systemd-timesync
+
 # This is for automatic testing purposes, do not do this in production.
 RUN echo "PermitRootLogin yes" > /etc/ssh/sshd_config.d/rootlogin.conf
 


davidcassany (approved):
LGTM

examples/green/Dockerfile
```diff
@@ -75,6 +75,8 @@ COPY --from=toolkit /usr/bin/elemental /usr/bin/elemental
 RUN systemctl enable NetworkManager.service && \
     systemctl enable sshd.service
 
+RUN useradd --system systemd-timesync
```
davidcassany, line 78: I wonder if there should be a `system-user-*` package for that, I tried to find one and I could not. `system-user-ntp` does not provide it either.

kkaempf:
What is the error _exactly_ and where are you seeing it ?

davidcassany:
> What is the error _exactly_ and where are you seeing it ?

We enable `systemd-timesyncd` by default in our `/system/oem` cloud config setup. Then on our install tests we actually check some services are active. Now `systemd-timesyncd` runs as the `systemd-timesync` user and fails if not present in the system.

davidcassany:
> I just checked on Tumbleweed (latest as of this morning). It has a respective user. So we need to find out where it's coming from. I'm not sure a `useradd` is the correct solution here 🤔
> 
> Btw, `systemd-timesyncd` is coming from `udev` but `udev` doesn't seem to add this user 😕

I am trying to figure out how is this added in our Micro images, it is actually there in them 🤷🏽‍♂️

frelon:
Merged in #2197

PR #2201, opened 2024-10-02
strict elemental setup

frelon:
- **Use --strict flag in elemental-setup services**
- **Set SilenceUsage for run-stage command**


davidcassany (commented):
Would be nice having some integration tests for that. What about setting some cloud-config in installation tests we know it will fail, so that we can check it still boots but on a degraded state, as some `elemental-setup-*` service failed. Even more, ideally we should be capable to see the failure in journalctl.

Alternatively, we also use such a process for the fallback test. Imagine we set a health check test that reports error if systemd is on degraded state.

Essentially, we should define the behavior we expect on a cloud-config failure and then add an integration tests to exercise it.

frelon:
@davidcassany This is true, I will look into some tests in this PR before merging!

davidcassany:
> Alternatively, we also use such a process for the fallback test. Imagine we set a health check test that reports error if systemd is on degraded state.

I am trying to implement a test for that, I think it should be relatively straight forward and I think this would cover the most important from integration test perspective.

davidcassany:
Checked manually that failures show up in journal, I don't think we need explicit test for that.

PR #2211, opened 2024-10-21
Update docs workflow

frelon:
Changes in the rancher github organization have disallowed the actions
we use for publishing the docs to github pages.

This change uses allowed actions to deploy the pages.

### DIFF
--- a/.github/workflows/docs-publish.yaml
+++ b/.github/workflows/docs-publish.yaml
@@ -6,41 +6,55 @@ on:
   push:
     branches:
       - main
-  schedule:
-   - cron: 0 20 * * *
-
-permissions:
-  contents: write
 
 jobs:
-  build-deploy:
+  build:
     runs-on: ubuntu-latest
+    env:
+      HUGO_VERSION: 0.111.0
     steps:
-      - name: Checkout Repo
+      - name: Install Hugo CLI
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-publish.yaml
```diff
@@ -9,38 +9,54 @@ on:
   schedule:
```
davidcassany, line 9: I wonder if shouldn't we remove this schedule.

.github/workflows/docs-publish.yaml
```diff
@@ -9,38 +9,54 @@ on:
...
+      pages: write
+      id-token: write
+    environment:
+      name: gh-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - name: Deploy to GitHub Pages
+        if: github.ref == 'refs/heads/main' && github.event_name == 'push'
```
davidcassany, line 60: I guess this is required to filter out the pull-request and schedule events. Is that correct?

davidcassany (approved):
LGTM

.github/workflows/docs-publish.yaml
frelon: Good call, I will remove it!

.github/workflows/docs-publish.yaml
frelon: > I guess this is required to filter out the pull-request and schedule events. Is that correct?

You got it!

Issue #1782, opened 2023-06-02
Disk encryption


frelon:
**Is your feature request related to a problem? Please describe.**
It should be possible to deploy a fully encrypted system using elemental.

**Describe the solution you'd like**
I have a POC from hackweek for the install-command here: https://github.com/frelon/elemental-cli/tree/cryptsetup which adds configuration options to run cryptsetup on `elemental install` to encrypt partitions.

After that there needs to be changes to the grub-config and immutable-rootfs dracut module in order to actually unlock the partitions. This will probably depend on #1781 being merged.

kkaempf:
Disk encryption is useful to prevent against unauthorized access to data in case a machine gets physically compromised.

Not sure how this is supposed to work in an edge/cluster scenario without anyone to enter the password.

'Supporting' this in Elemental might send the wrong message 🤔

frelon:
> Disk encryption is useful to prevent against unauthorized access to data in case a machine gets physically compromised.

Yep, edge seems the perfect candidate for encryption, since those machines are often more physically available.

> Not sure how this is supposed to work in an edge/cluster scenario without anyone to enter the password.

It's up to the user, but there are lots of ways of unlocking volumes without a password (TPM comes to mind).

Not having disk encryption at the edge feels like an oversight, but I'm happy to have a discussion about this! :+1:

kkaempf:
@agracey 👆🏻

Sporesirius:
NBDE (Network-Bound Disk Encryption) comes to mind. I think it is mainly used in [openshift](https://docs.openshift.com/container-platform/4.13/security/network_bound_disk_encryption/nbde-about-disk-encryption-technology.html), but the server and client are open source, [clevis](https://github.com/latchset/clevis) and [tang](https://github.com/latchset/tang).

kkaempf:
https://lists.opensuse.org/archives/list/factory@lists.opensuse.org/thread/WGWLTHVL4EFKL2DUJIG4RMM46RS3SWKD/ !

jmpolom:
elemental should be able to configure LUKS volumes for any partitions configured (state, persistent, oem, user defined, etc.) but currently appears not to be able to do so. These LUKS volumes should be configurable to unlock via binding to a host security device (like a TPM 2.0 or FIDO2 token) as well as a manual text input method (recovery key) using systemd-cryptenroll.

Any status on when such a feature might be supported? The proof of concept looks like a good start.

kkaempf:
[...]

PR #2219, opened 2024-10-29
Add persistent disk encryption

frelon:
Add flags to encrypt persistent partition on install:

* encrypt-persistent: flag to enable luks encryption on persistent partition.
* enroll-passphrase: string to enroll as passphrase to unlock partition.
* enroll-key-file: key-file to enroll as key to unlock partition.

During install this will invoke cryptsetup to create the LUKS partition and during mount we use systemd-cryptsetup to attach the partition before mounting the contained filesystem.

This also introduces some changes in the grub configuration, the encrypted_volumes variable can be set in grub_oem_env during install to configure which volumes are actually encrypted.

This change is tested in the new `test-encryption` target that runs an `elemental install` with a random passphrase and tests that the persistent partition and any extra-partitions can be unlocked when booted from the recovery system.

In the future the hope is to extend this to include state and oem partition, but that will need more changes to the bootloader configuration (most notably using the `cryptomount` command)

Part of #1782

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,7 +57,8 @@ RUN ARCH=$(uname -m); \
         gptfdisk \
         patterns-microos-selinux \
         btrfsprogs \
-	snapper \
+	    snapper \
+        cryptsetup \
         lvm2 && \
     zypper cc -a
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,7 +57,7 @@ RUN ARCH=$(uname -m); \
         gptfdisk \
         patterns-microos-selinux \
         btrfsprogs \
-	    snapper \
+        snapper \
         cryptsetup \
         lvm2 && \
     zypper cc -a
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -123,6 +123,7 @@ func NewInstallCmd(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 	addSharedInstallUpgradeFlags(c)
 	addLocalImageFlag(c)
 	addPlatformFlags(c)
+	addEncryptionFlags(c)
 	return c
 }
 
--- a/pkg/features/embedded/grub-default-bootargs/etc/elemental/bootargs.cfg
+++ b/pkg/features/embedded/grub-default-bootargs/etc/elemental/bootargs.cfg
@@ -1,9 +1,12 @@
 # bootargs.cfg inherits from grub.cfg several context variables:
 #   'img' => defines the image path to boot from. Active img is statically defined, does not require a value
+#   'mode' => active/passive/recovery, mode to boot
 #   'state_label' => label of the state partition filesystem
 #   'oem_label' => label of the oem partition filesystem
 #   'recovery_label' => label of the recovery partition filesystem
 #   'snapshotter' => snapshotter type, assumes loopdevice type if undefined
+#   'snap_arg' => kernel args for snapshotter
+#   'encrypted_volumes' => comma-separated list of encrypted volume names
 #
 # In addition bootargs.cfg is responsible of setting the following variables:
 #   'kernelcmd' => essential kernel command line parameters (all elemental specific and non elemental specific)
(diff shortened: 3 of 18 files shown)


jmpolom:
Are you considering supporting binding the LUKS volume to the TPM (or FIDO2 device) via systemd-cryptenroll with this? I would recommend either of these options in addition to password and key files.

frelon:
> Are you considering supporting binding the LUKS volume to the TPM (or FIDO2 device) via systemd-cryptenroll with this? I would recommend either of these options in addition to password and key files.

Yep, this will be low hanging fruit for a follow up pr! ;)

frelon:
I have been testing this by booting an installer with `make prepare-installer-test` and then running `elemental install /dev/vda --debug --encrypt-persistent --enroll-passphrase <passphrase>` and then rebooting.

It should also work to add a install yaml like the following:

```
install:
  target: /dev/vda
  encrypt-persistent: true
  enroll-passphrase:
    - "passphrase"
```

This config expands to:

```
install:
  target: /dev/vda
  partitions:
    persistent:
      encryption:
        name: /dev/mapper/cr_persistent
        key_slots:
        -  slot: 0
           passphrase: "passphrase"
```

which should also work, but is a bit cumbersome.

Dockerfile
```diff
@@ -57,7 +57,8 @@ RUN ARCH=$(uname -m); \
         gptfdisk \
         patterns-microos-selinux \
         btrfsprogs \
-	snapper \
+	    snapper \
```
davidcassany, line 60: nit: there is something odd with spaces, I guess a mixture of tabs and spaces.

cmd/install.go
```diff
@@ -123,6 +123,7 @@ func NewInstallCmd(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 	addSharedInstallUpgradeFlags(c)
 	addLocalImageFlag(c)
 	addPlatformFlags(c)
+	addEncryptionFlags(c)
```
davidcassany, line 126: I think those would also make sense on reset command.

pkg/features/embedded/grub-default-bootargs/etc/elemental/bootargs.cfg
```diff
@@ -1,9 +1,12 @@
 # bootargs.cfg inherits from grub.cfg several context variables:
 #   'img' => defines the image path to boot from. Active img is statically defined, does not require a value
+#   'mode' => active/passive/recovery, mode to boot
 #   'state_label' => label of the state partition filesystem
 #   'oem_label' => label of the oem partition filesystem
 #   'recovery_label' => label of the recovery partition filesystem
 #   'snapshotter' => snapshotter type, assumes loopdevice type if undefined
+#   'snap_arg' => kernel args for snapshotter
```
davidcassany, line 8: nit: this is not inherited, this is not part of an _API_ this is just a local variable. I'd not list it here.

pkg/features/embedded/grub-default-bootargs/etc/elemental/bootargs.cfg
```diff
@@ -1,9 +1,12 @@
 # bootargs.cfg inherits from grub.cfg several context variables:
 #   'img' => defines the image path to boot from. Active img is statically defined, does not require a value
+#   'mode' => active/passive/recovery, mode to boot
```
davidcassany, line 3: Good catch!

pkg/types/config.go
```diff
@@ -261,6 +264,36 @@ func (i *InstallSpec) Sanitize() error {
 		}
 	}
 
+	// Setup disk encryption
+	if i.EncryptPersistent {
```
davidcassany, line 268: Shouldn´t error out if EncryptPersistent is set to true and no secret is provided?

davidcassany (commented):
Looks good to me. This is cool!

I miss unit tests and including the encryption configuration in state file. So we can reset and get the persistent encrypted again as it was configured at a time or eventually re-encrypt with different parameters.

frelon:
> Looks good to me. This is cool!
> 
> I miss unit tests and including the encryption configuration in state file. So we can reset and get the persistent encrypted again as it was configured at a time or eventually re-encrypt with different parameters.

Good point, I will have an extra look at the reset functionality with this in mind!

pkg/types/config.go
frelon: > Shouldn´t error out if EncryptPersistent is set to true and no secret is provided?

I actually didn't test this edge-case.. I would expect an empty keyslot to allow anything to unlock it, but I will give it a try!

pkg/features/embedded/grub-default-bootargs/etc/elemental/bootargs.cfg
frelon: Yep, removed in latest commit!

Dockerfile
frelon: Yep, something weird happened when resolving a merge conflict.. fixed!

jmpolom:
Is LUKS encrypted persistent storage volumes now out of scope/off the development plan for elemental toolkit?

PR #2220, opened 2024-10-31
Btrfs snapshotter without snapper

davidcassany:
This pull requests provides a new `subvolumebackend` interface to handle btrfs subvolumes and snapshots. The motivation is to not make snapper strict requirement of the btrfs snapshotter to better suite distributions where snapper is not that much adopted.

In any case (even if snapper is not in use) the snapshotter is based and built around common snapper conventions, this is easier from a maintenance perspective and allows to eventually consider switching to snapper at a later point in time.

This PR adds a new config parameter `snapper` for the btrfs snapshotter:
```yaml
snapshotter:
  type: btrfs
  max-snaps: 4
  config:
    snapper: true
```

If snapper set to `true` (this is the default value) the snapshotter will attempt to use snapper, otherwise it will relay entirely on `btrfs` utility calls.

Snapper can't perform alone all the uses cases that the Elemental Snapshotter interface requires (specially corner cases related to first snapshot), for those cases the `snapperBackend` implementation makes use of the `btrfsBacked` implementation, hopefully this can be improved in up comming snapper releases (openSUSE/snapper#944).

This PR includes a couple minor behavior changes (on the snapperBackend) that are actually reflected in unit tests:
* The snapper configuration is applied directly over the snapshot (before it was applied over the working directory)
* Snapper is used to define the read only and default snapshot for all snapshots except for the first one

### DIFF
--- a/examples/green/snapshotter.yaml
+++ b/examples/green/snapshotter.yaml
@@ -2,4 +2,4 @@ snapshotter:
   type: btrfs
   max-snaps: 4
   config:
-    snapper: false
+    snapper: true
--- a/examples/green/snapshotter.yaml
+++ b/examples/green/snapshotter.yaml
@@ -1,3 +1,5 @@
 snapshotter:
   type: btrfs
   max-snaps: 4
+  config:
+    snapper: false
--- a/pkg/snapshotter/btrfs-backend.go
+++ b/pkg/snapshotter/btrfs-backend.go
@@ -87,13 +87,14 @@ func (d *Date) UnmarshalXML(dec *xml.Decoder, start xml.StartElement) error {
 }
 
 type btrfsBackend struct {
-	cfg      *types.Config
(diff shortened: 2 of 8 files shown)


davidcassany:
I consider this PR ready to be reviewed despite being in draft. Probably minor changes could be applied on top of the current logic to make public subvolumeBackend interface and its implementations in order to include specific unit tests.

anmazzotti (approved):
This looks good to me. I found it confusing that the snapper backend is using the (plain) btrfs one, but I agree that this should be addressed on the snapper side instead.
Nice improvement!

davidcassany:
> I found it confusing that the snapper backend is using the (plain) btrfs one

Yes I agree, so far this is the simplest idea that came to my mind to not duplicate code. Opened to other ideas too 😉

PR #2224, opened 2024-11-08
Expand downgrade test by upgrading back to origin after downgrading

davidcassany:
This PR expands the downgrade tests to include upgrading back again to the current head. So with this test we exercise three use cases:
1. Downgrade from current head to an older image
2. Downgrade/Upgrade active image from passive (being at current head code)
3. Upgrade from downgraded image back to current head (v2.1.1 -> HEAD)

The difference with the upgrade tests is that upgrade test is testing backward compatibility by upgrading v1.1.x -> HEAD, so that upgrade test is not performing any upgrade based brtrfs snapshotter, so this is what is now covered in use case 3 of the above list.


tests/assets/hacks_for_tests.yaml
```diff
@@ -0,0 +1,14 @@
+name: "Hacks for testing environment"
```
davidcassany, line 1: Downgrade tests downgrades TW to Leap 15.5 and this causes certain frictions, the gap is huge. Dbus services have been refactored and `dbus.service` in `/etc` is a symlink to `dbus-broker.service` in TW, however `dbus-broker.service` does not exist in Leap. Since `/etc/systemd` is persisted by default this is causing a glitch.

IMHO  `/etc/systemd` should have not been set as persistent by default #2226

frelon: Changing entirely which distro you are running feels like a risky thing to try to support.. I would say if someone wants to change distros between upgrades the correct way would be to upgrade/downgrade the recovery to the wanted version and then do a reset which includes resetting persistent data..

That said I think this is a nice improvement and having more immutability in `/etc` is a good idea!

frelon (approved):
LGTM!

PR #2235, opened 2024-11-21
RISC-V improvements + fixes

ldevulder:
This PR contains some RISC-V support improvement and fixes for small issues I found:
- Fix Makefile for riscv64
- Fix Dockerfile for riscv64
- Add missing Autologin feature
- Fix architecture platform value

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,6 @@ REPO?=local/elemental-$(FLAVOR)
 TOOLKIT_REPO?=local/elemental-toolkit
 DOCKER?=docker
 BASE_OS_IMAGE?=registry.opensuse.org/opensuse/tumbleweed
-BASE_OS_VERSION?=latest
 DOCKER_SOCK?=/var/run/docker.sock
 
 GIT_COMMIT?=$(shell git rev-parse HEAD)
@@ -31,17 +30,6 @@ PKG:=./cmd ./pkg/...
 LDFLAGS:=-w -s
 LDFLAGS+=-X "github.com/rancher/elemental-toolkit/v2/internal/version.version=$(GIT_TAG)"
 LDFLAGS+=-X "github.com/rancher/elemental-toolkit/v2/internal/version.gitCommit=$(GIT_COMMIT)"
-
-# For RISC-V 64bit support
-ifeq ($(PLATFORM),linux/riscv64)
-BASE_OS_IMAGE=registry.opensuse.org/opensuse/factory/riscv/images/opensuse/tumbleweed
-BASE_OS_VERSION=latest
-ifeq ($(FLAVOR),tumbleweed)
-OS_IMAGE=registry.opensuse.org/opensuse/factory/riscv/images/opensuse/tumbleweed
-ADD_REPO=https://download.opensuse.org/repositories/devel:/RISCV:/Factory:/Contrib:/StarFive:/VisionFive2/standard
-BUILD_OPTS=--build-arg OS_IMAGE=$(OS_IMAGE) --build-arg ADD_REPO=$(ADD_REPO)
-endif
-endif
 
 # default target
 .PHONY: all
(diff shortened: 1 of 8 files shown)


examples/green/Dockerfile
```diff
@@ -87,15 +80,35 @@ RUN echo "PermitRootLogin yes" > /etc/ssh/sshd_config.d/rootlogin.conf
...
+    [[ "${ARCH}" == "riscv64" ]] && mv -f /tmp/riscv_bootargs.cfg /etc/elemental/bootargs.cfg || true
+
 # Generate initrd with required elemental services
-RUN elemental --debug init --force
+RUN ARCH=$(uname -m); \
+    if [[ "${ARCH}" != "riscv64" ]]; then \
+      ADD_FEATURES+=" grub-default-bootargs"; \
+      [[ "${ARCH}" == "aarch64" ]] && ADD_FEATURES+=" arm-firmware"; \
+    fi; \
+    elemental --debug init --force \
```
frelon, line 94: I think we should keep the default features here since these are what's tested in the CI.

examples/green/Dockerfile
```diff
@@ -13,20 +13,13 @@ ARG VERSION
 ENV REPO=${REPO}
 ENV VERSION=${VERSION}
 
-# Workaround for RISC-V, specific kernel might be needed for some boards
-ARG ADD_REPO
-ENV ADD_REPO=${ADD_REPO}
-
 # Install kernel, systemd, dracut, grub2 and other required tools
 RUN ARCH=$(uname -m); \
-    zypper --non-interactive removerepo repo-update || true; \
```
frelon, line 22: Very nice!

Makefile
```diff
@@ -5,7 +5,11 @@ DISK?=$(shell ls $(ROOT_DIR)/build/elemental*.qcow2 2> /dev/null)
 DISKSIZE?=20G
 ISO?=$(shell ls $(ROOT_DIR)/build/elemental*.iso 2> /dev/null)
 FLAVOR?=green
+ifdef PLATFORM
+ARCH=$(subst linux/,,$(PLATFORM))
```
frelon, line 9: I think ARCH is only used to name the resulting disk images, so I would be in favor of removing it completely in the future.

frelon (changes requested):
Please change to use the default features in the green Dockerfile. Apart from that it looks good!

examples/green/Dockerfile
ldevulder: What do you mean? The code will keep the defaults for x86 (yes explicitly written) and only change for arm64 (adding `arm-firmware`) and riscv64 because this one needs specific Grub options.

Makefile
ldevulder: Yes, it seems to be only to name the resulting files. I had my riscv64 images named `elemental-green.x86_64.raw` :smile:

examples/green/Dockerfile
frelon: Yes, but we want to test that the default features added through the implicit case works (also `autologin` is not one of the default features)

ldevulder: In that case I will have to either use a dedicated Dockerfile or defined the list of features only for riscv64 (and arm64 maybe?) and keep the default for x86.

WDYT is the better way for you? I first tried to avoid a pretty much full copy/paste of the Dockerfile only for riscv64, but if you think that an easier way to do it
 that's fine.

ldevulder: In that case I will have to either use a dedicated Dockerfile or defined the list of features only for riscv64 (and arm64 maybe?) and keep the default for x86.

WDYT is the better way for you? I first tried to avoid a pretty much full copy/paste of the Dockerfile only for riscv64, but if you think that an easier way to do it
 that's fine.

frelon: Sure, I think using the defined features for riscv64/arm64 is fine, as long as we stick with the implicit configuration for x86! :+1:

ldevulder: Should be OK, let me know if something else needs to be modified ;-)

frelon (approved):

examples/green/Dockerfile
frelon: Looks good to me! Nice job! :+1:

PR #2239, opened 2024-12-01
Make tar creation deterministic

bmwiedemann:
for this, we sort entries, override owner+group+mtime and omit ctime+atime

based on https://reproducible-builds.org/docs/archives/#full-example

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,9 @@ FROM --platform=$BUILDPLATFORM golang:${GO_VERSION}-alpine AS elemental-bin
 ENV CGO_ENABLED=0
 WORKDIR /src/
 
+# install GNU tar instead of busybox one to support --sort
+RUN apk add --no-cache tar
+
 # Add specific dirs to the image so cache is not invalidated when modifying non go files
 ADD go.mod .
 ADD go.sum .
--- a/pkg/features/generate-tarballs.go
+++ b/pkg/features/generate-tarballs.go
@@ -62,7 +62,7 @@ func main() {
 
 		fmt.Printf("Generate %s from %s\n", output, input)
 
-		cmd := exec.Command("tar", "-C", inputDir, "-czvf", output, input)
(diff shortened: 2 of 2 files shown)


frelon:
Looks like busybox tar implementation does not implement `--sort` flag.. let me have a look how to get around it!

frelon:
Installing GNU tar in the Dockerfile before running go generate ./... seems to fix the issue:

```
...
FROM --platform=$BUILDPLATFORM golang:${GO_VERSION}-alpine AS elemental-bin

ENV CGO_ENABLED=0
WORKDIR /src/

+# install GNU tar instead of busybox one to support --sort
+RUN apk add --no-cache tar
+
# Add specific dirs to the image so cache is not invalidated when modifying non go files
ADD go.mod .
...
```

bmwiedemann:
I added your change.
Do you know if the mtime of 1970-01-01 would break or confuse something/someone? Or is the tar purely for internal use?

frelon:
> I added your change. Do you know if the mtime of 1970-01-01 would break or confuse something/someone? Or is the tar purely for internal use?

It's purely for internal use. The tarballs will be extracted into an oci image and the image will then be written to an iso or disk image.

From what I can tell the mtime is overwritten somewhere in that chain: 

```
elemental:~ # stat  /system/oem/00_layout.yaml
  File: /system/oem/00_layout.yaml
  Size: 673             Blocks: 8          IO Block: 4096   regular file
Device: 0,43    Inode: 991         Links: 1
Access: (0644/-rw-r--r--)  Uid: (    0/    root)   Gid: (    0/    root)
Context: system_u:object_r:default_t:s0
Access: 2024-12-02 15:35:34.019999984 +0000
Modify: 2024-12-02 15:33:04.000000000 +0000
Change: 2024-12-02 15:35:35.973333319 +0000
 Birth: 2024-12-02 15:35:17.843333329 +0000
```

frelon (approved):
LGTM!

PR #2246, opened 2024-12-16
Use latest snapper installation helper

davidcassany:
With the installation helper the btrfs backend instance is no longer required in snapper backend implementation.


pkg/snapshotter/snapper-backend.go
```diff
@@ -49,60 +51,123 @@ type snapperBackend struct {
 func newSnapperBackend(cfg *types.Config, maxSnapshots int) *snapperBackend {
 	// snapper backend embeds an instance of a btrfs backend to fill the gap for the
 	// operatons that snapper can't entirely handle.
-	return &snapperBackend{cfg: cfg, maxSnapshots: maxSnapshots, btrfs: newBtrfsBackend(cfg, maxSnapshots)}
+
+	// TODO detect wheather the current snapper supports the required installation helper
+	return &snapperBackend{cfg: cfg, maxSnapshots: maxSnapshots, btrfs: nil}
```
davidcassany, line 56: We should initiate btrfs backend if snapper does not support the latest installation helper and use it instead

davidcassany: The PR is blocked until a new snapper version is released in TW. Then addressing this TODO will be needed.

PR #2252, opened 2025-01-07
Update copyright year (2025)

frelon:
Happy new year! (Fix lints)


kkaempf:
Better fix your linter 😉 

Updating the copyright year should only happen when you actually touch the code.

PR #2260, opened 2025-02-04
Fix elemental-toolkit build for Dev env

davidcassany:
Align go.mod requirements with specfile requirements


frelon (approved):
Nice catch!

PR #2300, opened 2025-07-22
Minor change to patch upgradeSpec with in-use devices in case they differ from discovered ones

ibrokethecloud:
When multipathing is enabled, the OS boots off the multipath /dev/mapper device

During upgrade elemental uses `jaypipes/ghw` to find block devices from `/sys/block` tree.

This misses the multipath partitions, which can cause issue https://github.com/rancher/elemental-toolkit/issues/2299

The PR introduces a minor change where once the UpgradeSpec is generated from parsing of `/sys/block` tree we lookup the result of `blkid -L $LABEL` to identify partition with label.

This will result in the multipath partition being returned

For example from a sample multipath'd host

```
node-1-dhcp:~ # blkid -L COS_STATE
/dev/mapper/0QEMU_QEMU_HARDDISK_ABCDE-part4

node-1-dhcp:~ # blkid -t LABEL=COS_STATE
/dev/sdb4: LABEL="COS_STATE" UUID="ffacd836-c2bf-4c97-b816-ae54ef9ea56c" BLOCK_SIZE="4096" TYPE="ext4" PARTLABEL="state" PARTUUID="88fdf8e1-fba0-4b88-b537-49f59c5e881c"
/dev/mapper/0QEMU_QEMU_HARDDISK_ABCDE-part4: LABEL="COS_STATE" UUID="ffacd836-c2bf-4c97-b816-ae54ef9ea56c" BLOCK_SIZE="4096" TYPE="ext4" PARTLABEL="state" PARTUUID="88fdf8e1-fba0-4b88-b537-49f59c5e881c"
/dev/sda4: LABEL="COS_STATE" UUID="ffacd836-c2bf-4c97-b816-ae54ef9ea56c" BLOCK_SIZE="4096" TYPE="ext4" PARTLABEL="state" PARTUUID="88fdf8e1-fba0-4b88-b537-49f59c5e881c"
```

If the partition does not match the one generated from parsing `/sys/block` then we update the upgrade spec to use the newly found partition.

### DIFF
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/mount-utils"
 
 	"github.com/rancher/elemental-toolkit/internal/version"
+
 	"github.com/rancher/elemental-toolkit/pkg/config"
 	"github.com/rancher/elemental-toolkit/pkg/constants"
 	v1 "github.com/rancher/elemental-toolkit/pkg/types/v1"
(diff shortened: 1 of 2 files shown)


ibrokethecloud:
@davidcassany i have added an extra commit which fixes the header check.

the `cli` ci is failing because the `make test-deps` target is trying to fetch package,, which no longer exists. 
```
go: github.com/tredoe/osutil/v2@v2.0.0-rc.16: reading github.com/tredoe/osutil/go.mod at revision v2.0.0-rc.16: 
```
There should be no actual issue building this since the module is already vendored.

PR #2327, opened 2025-09-25
Use chroot to label previously binded mountpoints

davidcassany:
This fixes upgrades from Micro 6.1 to 6.2


davidcassany:
Tests relaying on the expanding disks are currently failing in due to an unrelated issue. Apparently, since some days the example green OS can't mount a btrfs filesystem with filesystem auto detect. `mount -t auto /dev/sda4` fails if /dev/sda4 is formatted with btrfs... not sure why is this causing issues for `elemental reset` but not in `elemental install`.... this does not seam to be a problem in orange and blue, there the test-active succeeds without issues.
So for this PR the expectation is that test-installer and test-upgrade should succeed as they do not relay on the disk build. Note this error is also happening in _main_ branch too, [see it here](https://github.com/rancher/elemental-toolkit/actions/runs/17910618063/job/51321104482).

PR #2330, opened 2025-09-26
Use chroot to label previously binded mountpoints

davidcassany:
This prevents errors with non chrooted setfiles calls in which fail due to policy divergences between host and target.

For that purpose a new option to chroot utility has been included to allow chroot without setting the default bind mounts for /dev/, /sys and /proc.

(cherry picked from commit a8982b16ce8be9c2b51054180cee2163bd14486c)


frelon (approved):

davidcassany:
> FWIW I think we should start to merge these fixes to main first and then backport, otherwise we are running the risk of missing one of these forwardports.

I agree, I did it this way because our main branch tests suddenly went red and I did not want to merge it main in that status and moving on with v2.3.x was time pressing.

PR #2334, opened 2025-09-29
[do not merge] [FYI] To Reproduce 2297 harvester 8759

w13915984028:
The constructed test code to reproduce issue 

https://github.com/rancher/elemental-toolkit/issues/2297 and https://github.com/harvester/harvester/issues/8759

enter path `/go/src/github.com/rancher/elemental-toolkit/elemental-toolkit/pkg/cloudinit` then run `go test`


w13915984028:
Issue has been reproduced.

on v1.1.x via https://github.com/rancher/elemental-toolkit/pull/2334
and on main via  https://github.com/rancher/elemental-toolkit/pull/2335

PR #2336, opened 2025-09-29
rename cos -> elemental in tests

frelon:
/run/cos has been renamed to /run/elemental in a previous version.

### DIFF
--- a/tests/wait-active/wait_test.go
+++ b/tests/wait-active/wait_test.go
@@ -38,7 +38,7 @@ var _ = Describe("Elemental booting an expandable disk image", func() {
 	Context("Wait until system is expanded", func() {
 		It("eventually is active", func() {
 			Eventually(func() string {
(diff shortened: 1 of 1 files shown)


davidcassany (approved):
LGTM

tests/wait-active/wait_test.go
```diff
@@ -38,7 +38,7 @@ var _ = Describe("Elemental booting an expandable disk image", func() {
 	Context("Wait until system is expanded", func() {
 		It("eventually is active", func() {
 			Eventually(func() string {
-				out, _ := s.Command("cat /run/cos/active_mode")
+				out, _ := s.Command("cat /run/elemental/active_mode")
```
davidcassany, line 41: I see upgrade test failing, I guess the image upgrading from is not having this path yet... probably we can add here a fallback to check both paths? or update image we are upgrading from.

PR #2337, opened 2025-09-30
Add snapshotter.type flag to build-disk command

frelon:
When building a disk image we do not specify the snapshotter, which
configures the expansion mechanism to set COS_STATE partition filesystem
to ext4.

The green flavor uses the btrfs snapshotter, which means when it boots
and tries to mount the COS_STATE partition it expects a btrfs
filesystem, but finds an ext4 one, which errors.

In this commit we add the --snapshotter.type flag to the build-disk
command and set the snapshotter in the Makefile accordingly.


davidcassany:
Good catch!

What I don't understand is what changed to actually make this issue prominent 🤔 How is that it just appeared since a couple of weeks?

PR #2386, opened 2026-05-08
Update libraries

davidcassany:
Update several libraries, each commit are the changes related to the bump of a specific library.


frelon (approved):
LGTM 😎

PR #2423, opened 2026-06-22
Remove arm64 build support

ldevulder:
Arm64 is not an officially supported platform, so due to constraint resources we have to remove arm64 GH build support.

Code remains available but in best effort mode.


frelon (approved):

davidcassany (approved):
LGTM

PR #2434, opened 2026-09-13
Only consider global Boot#### variables when clearing stale entries

### LINKED ISSUE #2433: install: boot-entry cleanup matches any EFI variable containing "Boot####" under any GUID, and aborts on Lenovo's undeletable lBoot#### mirrors
**Environment.** Harvester v1.8.2 (bundled elemental `1.1.8+gaf99ad9`), Lenovo ThinkStation P620 (BIOS S07KT24A), UEFI, unattended install over PXE. The machine had a previous elemental-based install whose standard boot entry had already been removed.

**Symptom.** `elemental install` fails after partitioning:

```
1 error occurred:
* remove /sys/firmware/efi/efivars/lBoot0006-146b234d-4052-4e07-b326-11220f8e1fe8: invalid argument
```

`lBoot0006` is not a boot entry. It is a Lenovo vendor variable under GUID `146b234d-4052-4e07-b326-11220f8e1fe8`: this firmware keeps an `lBoot####` twin of every `Boot####`, with attributes `0x6` (BootService + Runtime, **not** NonVolatile) — volatile mirrors it regenerates each boot, which is why it refuses to delete them (`chattr -i` + `rm` → `EINVAL` as well). Their contents parse as load options, so the twin of a former `elemental-shim` entry carries that description.

**Cause.** The stale-entry cleanup matches every variable whose *name contains* `Boot[0-9a-fA-F]{4}`, under any GUID:

- v1.1.8: `pkg/utils/grub.go` `ClearBootEntry` — `regexp.MustCompile(`Boot[0-9a-fA-F]{4}`).MatchString(v.Name)` over `efilib.ListVariables()`, no check of `v.GUID`
- main (f04a36c): `pkg/bootloader/grub.go` `clearEntry`, same pattern

The vendor twin matches, its description equals the entry name, the delete fails, and the whole install aborts.

**Expected.** Only names matching `^Boot[0-9a-fA-F]{4}$` under `efi.GlobalVariable` should be candidates. Arguably a failed delete of a *stale* entry should also not abort the install, since the entry that matters is the one created afterwards.

**Reproduce.** On a Lenovo UEFI machine that has had an elemental install before: `ls /sys/firmware/efi/efivars | grep -i 'boot[0-9a-f]\{4\}'` shows `lBoot####-146b234d-…` entries; run `elemental install`.

**Workaround that worked.** Before the install, from the live system: `mount --bind /dev/null /sys/firmware/efi/efivars/lBoot0006-146b234d-4052-4e07-b326-11220f8e1fe8`. The variable then fails to parse as a load option, the cleanup skips it, and the real entry is created normally. Remounting efivars read-only does not help on 1.1.8 (no `WithGrubAutoDisableBootEntry` there): the error just becomes `read-only file system`.

leptitchriss:
The stale boot entry cleanup matched any EFI variable whose name contains "Boot####", under any GUID. Some firmware keeps vendor variables of that shape whose contents also parse as load options (Lenovo mirrors every Boot#### as lBoot#### under its own GUID) and refuses to delete them, so a previous install's entry made every later install fail with "invalid argument" on the vendor variable.

This restricts the candidates to names matching `^Boot[0-9a-fA-F]{4}$` under the EFI global variable GUID, which is what a boot entry is. The behaviour is covered with the mock variables through an `export_test.go` shim, so the method stays unexported. The new spec plants a vendor-GUID twin of a stale entry and an unrelated global entry beside it and checks that only the stale entry is removed; it fails on the previous logic.

Fixes #2433


davidcassany (approved):
Looks good to me, thanks for the contribution and explanation on the specific issue

